PR #21, opened 2025-06-18
fix: use SessionConfig for cloud execution

german-typedef:
When the session is executed on the server, it needs the base SessionConfig to be serialized and pass to the Engine.


src/fenic/_backends/cloud/engine_config.py
```diff
@@ -25,9 +25,11 @@ class CloudSessionConfig:
     Upon initialization, will read the API keys required for the SemanticConfig from the environment variables
     """
 
-    def __init__(self, session_config: ResolvedSessionConfig):
+    def __init__(self, session_config: SessionConfig):
         self.model_api_keys = {}
-        self.session_config = copy.deepcopy(session_config)
+        self.unresolved_session_config = copy.deepcopy(session_config)
+        self.session_config = copy.deepcopy(session_config._to_resolved_config())
```
rohitrastogi, line 31: Unrelated to the scope of this pr, but I'm not sure we need to mutate and thus copy these objects.

german-typedef: We probably could use a better mechanism to determine if we need to create a local vs. a cloud session in the Engine.

PR #32, opened 2025-06-20
feat: Consume query metrics

german-typedef:
1. For cloud execution, the Query Metrics are available in the catalog, they can be retrieved using the execution_id
2. Reading the arrow results for an execution using collect doesn't have to be async (pyarrow has its own threading model)

### DIFF
--- a/src/fenic/_backends/cloud/execution.py
+++ b/src/fenic/_backends/cloud/execution.py
@@ -90,12 +90,7 @@ class CloudExecution(BaseExecution):
             self.session_state.asyncio_loop,
         )
         execution_id = future.result()
-
-        result_future = asyncio.run_coroutine_threadsafe(
-            self._get_execution_result_from_arrow(execution_id),
-            self.session_state.asyncio_loop,
-        )
-        df = result_future.result()
+        df = self._get_execution_result_from_arrow(execution_id)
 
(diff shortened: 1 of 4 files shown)


src/fenic/_backends/cloud/metrics.py
```diff
@@ -0,0 +1,49 @@
...
+            elif metric.metric_name == "lm_num_cached_input_tokens":
+                query_metrics.total_lm_metrics.num_cached_input_tokens = metric.metric_value
+            elif metric.metric_name == "lm_num_output_tokens":
+                query_metrics.total_lm_metrics.num_output_tokens = metric.metric_value
+            elif metric.metric_name == "rm_num_input_tokens":
+                query_metrics.total_rm_metrics.num_input_tokens = metric.metric_value
+            elif metric.metric_name == "rm_num_requests":
+                query_metrics.total_rm_metrics.num_requests = metric.metric_value
+            elif metric.metric_name == "rm_cost":
+                query_metrics.total_rm_metrics.cost = metric.metric_value
```
rohitrastogi, line 45: Maybe we can log that we encountered a metric we weren't expecting.

german-typedef: Agreed, I've added a log entry for that case.

PR #37, opened 2025-06-23
feat: summarization function

sanjibansg:
This PR adds a column function for Summarization to Fenic.

Summarization only works on string and involves either summing up texts to a Paragraph with restrictions on the maximum words to use, or Key Points with restrictions on the maximum points to use.


src/fenic/_backends/local/semantic_operators/summarize.py
```diff
@@ -0,0 +1,77 @@
...
+        self.format = format
+
+        super().__init__(
+            input,
+            CompletionOnlyRequestSender(
+                operator_name="semantic.summarize",
+                inference_config=InferenceConfiguration(
+                    max_output_tokens=self.get_max_tokens(),
+                    temperature=temperature,
+                    response_format=SimpleStringOutputModelResponse,
```
bcallender: we don't need a response format since we are just returning a string. since you're using the default postprocess function, this is probably what's causing the output to be empty.

src/fenic/_backends/local/semantic_operators/summarize.py
```diff
@@ -0,0 +1,77 @@
...
+                ),
+                model=model,
+            ),
+            None,
+        )
+
+    def postprocess(self, responses):
+        return super().postprocess(responses)
```
bcallender: there is no behavior defined in `super().postprocess(...)`. this should just be `return responses`, like map

src/fenic/_backends/local/semantic_operators/summarize.py
sanjibansg: thanks for helping with this! made the change.

src/fenic/_backends/local/semantic_operators/summarize.py
sanjibansg: got it, thanks, made the change.

src/fenic/_backends/local/semantic_operators/summarize.py
```diff
@@ -0,0 +1,70 @@
...
+                inference_config=InferenceConfiguration(
+                    max_output_tokens=self.get_max_tokens(),
+                    temperature=temperature,
+                ),
+                model=model,
+            ),
+            None,
+        )
+
+    def postprocess(self, responses):
```
bcallender: we want to be explicit with the types here, `responses` is `List[Optional[str]]` and it should also return `List[Optional[str]]`. This is why we just need to return the input, it's already in the proper type.

src/fenic/core/types/summarize_structures.py
```diff
@@ -0,0 +1,35 @@
...
+
+class SummaryFormat:
+    """Base class for summary formats."""
+    pass
+
+@dataclass
+class KeyPoints(SummaryFormat):
+    """Summary as a concise bulleted list."""
+
+    max_points: int = 5
```
bcallender: once this gets converted to pydantic, we can add a `Field` validation to this to ensure it is greater than 0

src/fenic/core/types/summarize_structures.py
```diff
@@ -0,0 +1,35 @@
...
+        return "The summary should be presented as a concise bulleted list, with each bullet capturing a distinct and essential idea, and the total number of points not exceeding " + str(self.max_points) + " points."
+
+    def max_tokens(self) -> int:
+        """Calculate the maximum number of tokens for the summary based on the number of key points."""
+        return self.max_points * 75
+
+@dataclass
+class Paragraph(SummaryFormat):
+    """Summary as a cohesive narrative."""
+    max_words: int = 120
```
bcallender: similarly here, we can enforce that the `max_words` is greater than, say, 10 or 20 words.

src/fenic/api/functions/semantic.py
```diff
@@ -393,3 +396,35 @@ def embed(
     return Column._from_logical_expr(
         EmbeddingsExpr(Column._from_col_or_name(column)._logical_expr, model_alias=model_alias)
     )
+
+@validate_call(config=ConfigDict(strict=True, arbitrary_types_allowed=True))
+def summarize(
+    column: ColumnOrName,
+    format: Union[KeyPoints, Paragraph] = None,
```
bcallender: we should provide a default here, probably `Paragraph`

src/fenic/api/functions/semantic.py
```diff
@@ -393,3 +396,35 @@ def embed(
...
+    temperature: float = 0,
+    max_output_tokens: int = 1024,
```
bcallender: i think it's confusing to expose this in addition to the format -- i'm not sure which one has priority. if i set this, will it override the value that is generated by the format? We should probably just allow the user to customize the format, and we will deduce the `max_output_tokens` based on that. As long as we clearly document that the specifications you pass into `format` will decide what the length of the output is.

src/fenic/core/types/summarize_structures.py
```diff
@@ -0,0 +1,35 @@
...
+
+@dataclass
+class KeyPoints(SummaryFormat):
+    """Summary as a concise bulleted list."""
+
+    max_points: int = 5
+
+    def __str__(self):
+        """Return a description of the summary format for KeyPoints."""
+        return "The summary should be presented as a concise bulleted list, with each bullet capturing a distinct and essential idea, and the total number of points not exceeding " + str(self.max_points) + " points."
```
bcallender: not required, but `f-strings` just make things like this easier to read imo. also, if you use them, then you don't have to call `str`, since it will automatically be called to convert the object into a string,

bcallender: TIL: technically, `f-string` will call `__format__` on the object, not `__str__`.  If you haven't defined a `__format__` it will fall back to `__str__`, which is fine for us

src/fenic/core/types/summarize_structures.py
```diff
@@ -0,0 +1,35 @@
...
+
+from dataclasses import dataclass
+
+
+class SummaryFormat:
+    """Base class for summary formats."""
+    pass
+
+@dataclass
+class KeyPoints(SummaryFormat):
```
bcallender: since these are user-facing objects, we can use pydantic BaseModel as the base class here.

src/fenic/_backends/local/semantic_operators/summarize.py
```diff
@@ -0,0 +1,70 @@
...
+
+            Format: {format_description}
+
+            Your summary should enable readers to understand the essential content without needing to read the original text, while clearly indicating this is a condensed version. '''
+    )
+
+    def __init__(
+        self,
+        input: pl.Series,
+        format: Union[KeyPoints, Paragraph],
```
bcallender, line 43: If you're only ever using a `Union` to represent that it can be either format, then you don't need a common base class. If you have a common base class, it makes sense to pass that, and not a `Union`. It doesn't really make sense to have a base class, and then never use it. Either of the first two options are fine.

src/fenic/_backends/local/semantic_operators/summarize.py
sanjibansg: made the change, thanks!

src/fenic/core/types/summarize_structures.py
sanjibansg: made the change, does this look good now?

src/fenic/core/types/summarize_structures.py
sanjibansg: made the change.

src/fenic/api/functions/semantic.py
sanjibansg: when I try adding a default case with `Paragraph()` there, I see a lint issue saying, `Do not perform function call `Paragraph` in argument defaults; instead, perform the call within the function, or read the default from a module-level singleton variable    `

src/fenic/api/functions/semantic.py
sanjibansg: made the change, thanks

src/fenic/core/types/summarize_structures.py
sanjibansg: made the change, thanks

src/fenic/core/types/summarize_structures.py
sanjibansg: made the change, thanks

src/fenic/_backends/local/semantic_operators/summarize.py
sanjibansg: made the change, thanks.

src/fenic/api/functions/semantic.py
bcallender: ah, fair enough. in that case then, this type should be `Optional[SummaryFormat]`, (or if you really feel strongly about the union, `Union[KeyPoints, Paragraph, None]`  which better indicates that it is optional to provide. 

You also still need to update the docstring to remove `max_output_tokens` and inform the user that if a format is not provided, `Paragraph` will be selected with an approximate maximum of 120 words.

src/fenic/core/types/summarize_structures.py
```diff
@@ -1,34 +1,35 @@
...
-    """Base class for summary formats."""
-    pass
-
-@dataclass
-class KeyPoints(SummaryFormat):
+class KeyPoints(BaseModel):
     """Summary as a concise bulleted list."""
 
-    max_points: int = 5
+    max_points: int = Field(default=5, gt=0)
```
bcallender: since these are public facing, their `Attributes` should be described in the docstring. That way, the explanations show up in the generated docs, and when the user is using an IDE, it can show the description of the fields when they mouse over the type.

src/fenic/core/types/summarize_structures.py
bcallender, line 1: we know they are data structures because they are in the `types` module. we can just call this file `summarization.py` or `summarize.py`

bcallender (changes requested):
just a few small nits

src/fenic/api/functions/semantic.py
sanjibansg: made the change, thanks!

src/fenic/core/types/summarize_structures.py
sanjibansg: made the change thanks

src/fenic/core/types/summarize_structures.py
sanjibansg: made the change, thanks!

tests/_backends/local/functions/test_semantic_summarize.py
```diff
@@ -0,0 +1,79 @@
...
+            ],
+        }
+    )
+
+    df = source.select(
+        semantic.summarize(col("text")).alias("summarized_text")
+    )
+    result = df.to_polars()
+    assert result.schema["summarized_text"] == pl.String
+    assert len(result['summarized_text'])<120
```
bcallender, line 31: can you also assert that the result is > 0? just so we have some smoke test that the summarization is working.

tests/_backends/local/functions/test_semantic_summarize.py
```diff
@@ -0,0 +1,79 @@
...
+            ],
+        }
+    )
+    df = source.select(
+        semantic.summarize(col("text"), format=KeyPoints(max_points = 10)).alias("summarized_text")
+    )
+    result = df.to_polars()
+    assert result.schema["summarized_text"] == pl.String
+    # should not exceed 10 key points
+    assert sum(1 for line in result['summarized_text'][0].splitlines() if line.strip().startswith("- ")) <= 10
```
bcallender: same here, there should be at least one point.

tests/_backends/local/functions/test_semantic_summarize.py
sanjibansg: made the change, thanks!

tests/_backends/local/functions/test_semantic_summarize.py
sanjibansg: made the change, thanks!

bcallender (approved):

PR #41, opened 2025-06-23
feat: Persistent views

sanjibansg:
This PR adds the support for persisting views.


src/fenic/api/catalog.py
```diff
@@ -495,15 +495,138 @@ def create_table(
         """
         return self.catalog.create_table(table_name, schema, ignore_if_exists)
 
+    @validate_call(config=ConfigDict(strict=True))
+    def create_view(
```
rohitrastogi: Why is this function exposed as part of the public catalog api? Users should be using save_as_view as the only API to create views.

src/fenic/_backends/local/catalog.py
```diff
@@ -440,6 +536,15 @@ def read_df_from_table(self, table_name: str) -> pl.DataFrame:
...
             ) from e
 
+    def validate_view_schema(self, plan_node):
+        for child in plan_node.children():
+            if child.children():
+                self.validate_view_schema(child)
+                continue
+
+            if child.schema() != child._build_schema():
+                    raise PlanError("Failed to validate schema against existing sources during logical plan construction—check the referenced tables, views or other sources.")
```
rohitrastogi: Can you make this error message a bit more user friendly? What interventions can a user take to fix this? Keep in mind that a user may not be familiar with concepts like LogicalPlans.

src/fenic/_backends/local/catalog.py
```diff
@@ -440,6 +536,15 @@ def read_df_from_table(self, table_name: str) -> pl.DataFrame:
                 f"Failed to read dataframe from table: `{table_identifier.db}.{table_identifier.table}`"
             ) from e
 
+    def validate_view_schema(self, plan_node):
+        for child in plan_node.children():
+            if child.children():
+                self.validate_view_schema(child)
+                continue
+
+            if child.schema() != child._build_schema():
```
rohitrastogi: _build_schema is a private method, which means it should not be used outside of logical plan construction (it's an implementation detail).

Also, I don't think it's required to compare the schemas for ALL nodes (just the source node). If the source nodes have changed, the parent schemas will be different. If the source nodes have not changed, the parents cannot have changed.

Perhaps we can simply collect the source nodes and call each source node's public constructor (instead of using _build_schema) and then compare the schemas.

src/fenic/_backends/local/system_table_client.py
```diff
@@ -0,0 +1,393 @@
...
+            else:
+                logger.debug(f"No schema metadata found for database {database_name}")
+
+            return rows_deleted
+        except Exception as e:
+            raise CatalogError(
+                f"Failed to delete schema metadata for database {database_name}: {e}"
+            ) from e
+
+    def save_view(
```
rohitrastogi, line 220: I think the abstraction would be cleaner if this function just accepted the logical plan (and handled all of the serialization). This way clients don't have to worry about how serialization happens.

src/fenic/_backends/local/system_table_client.py
```diff
@@ -0,0 +1,393 @@
...
+
+            logger.debug(f"Saved View for {database_name}.{view_name}")
+        except Exception as e:
+            raise CatalogError(
+                f"Failed to save view for {database_name}.{view_name}"
+            ) from e
+
+    def get_view(
+        self, database_name: str, view_name: str
+    ) -> Optional[Tuple[object, object]]:
```
rohitrastogi: Likewise, instead of returning opaque Tuple[object, object], we can do the deserialization here in this method so callers don't need to worry about it.

tests/_backends/local/test_metrics.py
```diff
@@ -147,3 +147,225 @@ def test_semantic_metrics(local_session):
             assert operator_metrics.lm_metrics.cost > 0
             assert operator_metrics.rm_metrics.num_input_tokens == 0
             assert operator_metrics.rm_metrics.cost == 0
+
+
+def test_metrics_from_view(local_session):
+    sales_data = {
```
rohitrastogi: Can we pull all of the common test setup into a test fixture? There's no point duplicating all of this for test_metrics_from_view and test_simple_metrics.

tests/_backends/local/io/test_reader.py
```diff
@@ -973,3 +973,39 @@ def test_ingest_array_type(local_session, temp_dir):
...
+John,25,New York
+David,33,Seattle"""
+
+    test_file_path = f"{temp_dir.path}/test_csv_single_file.csv"
+    write_test_file(test_file_path, test_data, local_session, "csv")
+
+    df_csv = local_session.read.csv(test_file_path)
+    df_csv.write.save_as_view("view")
+    df_csv_view = local_session.view("view")
+    df_csv_view.collect()
```
rohitrastogi: we don't need this collect here.

tests/_backends/local/io/test_reader.py
```diff
@@ -973,3 +973,39 @@ def test_ingest_array_type(local_session, temp_dir):
...
+    df = local_session.create_dataframe([{"name": "Alice", "age": 30, "city": "SF"},
+                                          {"name": "Carol", "age": 34, "city": "Boston"}])
+    df.write.save_as_table(table_name, mode="overwrite")
+    df_table = local_session.table(table_name)
+    df_table.write.save_as_view("df_view")
+
+    df_view = local_session.view("df_view")
+
+    df_union = df_view_nested.union(df_view)
+    df_union.collect()
```
rohitrastogi: we also dont need this collect here

tests/_backends/local/io/test_reader.py
```diff
@@ -973,3 +973,39 @@ def test_ingest_array_type(local_session, temp_dir):
             "array_column": pl.List(pl.Float32),
         }
     )
+
+
```
rohitrastogi, line 977: it will be useful to explain what this test is doing in a short comment, because it is not obvious.

I'd also recommend adding some intermediate transform nodes between the nested view nodes to make the test a bit less contrived.

tests/_backends/local/io/test_writer.py
```diff
@@ -323,3 +324,19 @@ def test_write_with_invalid_file_path(local_session, temp_dir):
     output_path = f"{temp_dir.path}/bad_test_file"
     with pytest.raises(FenicValidationError, match="Parquet writer requires a '.parquet' file extension."):
         df1.write.parquet(output_path, mode="overwrite")
+
+def test_view_writer_roundtrip(local_session):
```
rohitrastogi, line 332: I think it may make sense to create a new test file for views since your tests are testing both the read and write path.

Also, it'll be important to include a test for the view invalidation case where a file/table's schema changes in between view registration and view usage.

src/fenic/_backends/local/system_table_client.py
german-typedef: Moved to use the LogicalPlan definition for the view.
Serialization happens using our Serde mechanisms.

tests/_backends/local/io/test_writer.py
german-typedef: I've added a test in the test_reader with a file to check view invalidation when a file schema changes.

src/fenic/api/catalog.py
german-typedef: Removed from the API.

src/fenic/_backends/local/catalog.py
german-typedef: Modified the error message for the validation.

src/fenic/_backends/local/catalog.py
german-typedef: Validate now uses the public constructor (with a session state) for the sources (File/Table).

src/fenic/_backends/local/system_table_client.py
german-typedef: get_view now returns the LogicalPlan for the view.

tests/_backends/local/test_metrics.py
german-typedef: Moved to test fixtures for the data setup.

tests/_backends/local/io/test_reader.py
german-typedef: Removed.

tests/_backends/local/io/test_reader.py
german-typedef: Removed.

src/fenic/api/catalog.py
```diff
@@ -495,15 +496,101 @@ def create_table(
...
+            List[str]: A list of view names stored in the database.
+            Returns an empty list if no views are found.
+
+        Example:
+            >>> session.catalog.list_views()
+            ['view1', 'view2', 'view3'].
+        """
+        return self.catalog.list_views()
+
+    @validate_call(config=ConfigDict(strict=True))
```
rohitrastogi, line 513: We need to delete this line.

src/fenic/api/catalog.py
```diff
@@ -495,15 +496,101 @@ def create_table(
         """
         return self.catalog.create_table(table_name, schema, ignore_if_exists)
 
+    def describe_view(self, view_name: str) -> LogicalPlan:
```
rohitrastogi: describe_view should not be part of the public catalog API. It should be part of the BaseCatalog interface and its implementations. The public API for view should be through session.

src/fenic/_backends/local/system_table_client.py
```diff
@@ -0,0 +1,402 @@
...
+
+            logger.debug(f"Deleted views for {database_name}.{view_name}")
+            return True
+        except Exception as e:
+            raise CatalogError(
+                f"Failed to delete views for {database_name}.{view_name}"
+            ) from e
+
+    def delete_database_views(self, database_name: str) -> int:
+
```
rohitrastogi: please remove this space

tests/_backends/local/io/test_reader.py
```diff
@@ -973,3 +973,60 @@ def test_ingest_array_type(local_session, temp_dir):
...
+def test_view_csv_union(local_session, temp_dir):
```
rohitrastogi: nit: consider renaming to test_nested_views

src/fenic/_backends/local/catalog.py
```diff
@@ -444,6 +540,10 @@ def _build_qualified_table_name(self, table_identifier: TableIdentifier,
     ) -> str:
         return f'"{table_identifier.db}"."{table_identifier.table}"'
 
+    def validate_view(self, view_name: str, logical_plan: LogicalPlan, session_state: BaseSessionState) -> None:
```
rohitrastogi: This doesn't need to be part of the BaseCatalog impls. We can just use the utility directly, I think.

src/fenic/api/catalog.py
german-typedef: Done.

src/fenic/api/catalog.py
german-typedef: +1. Removed from the public API.

src/fenic/_backends/local/system_table_client.py
german-typedef: Done.

tests/_backends/local/io/test_reader.py
german-typedef: Makes sense, I've renamed the test.

src/fenic/_backends/local/catalog.py
german-typedef: Agreed, we already have the utility (there is no state that we need here).

PR #48, opened 2025-06-30
fix: use basic types for expressions

german-typedef:
**What**
For semantic classify we can pass either an enum or a list of strings that will be transformed into an enum. When the list of strings is passed, that enum is dynamically generated, causing problems when calling pickle to serialize the plan.
For semantic extract we can pass either a ExtractSchema or a pydantic model, in the case of a ExtractSchema, this is transformed into a pydantic model during runtime causing a problem while pickling the plan.

**The Fix**
The proposal is to only do the transformations during execution (we might do transformations to validate), but while the logical plan is generated there shouldn't be any dynamic calls.


bcallender:
* **#57** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/57?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#52** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/52?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#48** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/48?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/48?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tests/_backends/local/dataframe/test_semantic_cluster.py
```diff
@@ -30,6 +31,8 @@ def test_semantic_cluster_with_centroids(local_session):
         source.with_column("embeddings", semantic.embed(col("blurb")))
         .semantic.with_cluster_labels(col("embeddings"), 2, centroid_column="cluster_centroid")
     )
+    deserialized_df = _test_df_serialization(df, local_session._session_state)
```
rohitrastogi: I prefer keeping serde tests outside of the tests that maintain the business logic of the operations.
I believe we already have an existing test suite for serialization, can we add this there?

tests/_utils/serde_utils.py
```diff
@@ -0,0 +1,42 @@
...
+    plan._build_schema()
+
+    # Test equivalence
+    assert isinstance(deserialized_with_session_state, type(plan))
+    assert plan._repr() == deserialized_with_session_state._repr()
+    assert str(plan._build_schema()) == str(deserialized_with_session_state._build_schema())
+
+    # Test children if any
+    assert len(plan.children()) == len(deserialized_with_session_state.children())
+    # for orig_child, deser_child in zip(plan.children(), deserialized.children()):
```
rohitrastogi: remove commented code?

german-typedef: my bad, this shouldn't have been merged. thanks for the catch.

tests/_backends/local/dataframe/test_semantic_cluster.py
german-typedef: Sounds good, I've updated the serde suite with more tests to cover the semantic portion.

PR #53, opened 2025-07-03
feat: function registry for signature validation

bcallender:
# feat: Function registry system for centralized signature validation

## Summary

Establishes the foundation for a centralized function signature validation system that enables automatic type checking and return type inference for all scalar functions.

## Key Changes

- **Function Registry**: Centralized registry for storing and retrieving function signatures
- **Function Signature System**: Complete signature definition combining type validation and return type inference with multiple strategies (SAME_AS_INPUT, PROMOTED, DYNAMIC)  
- **ScalarFunction Base Class**: Abstract base class that automatically validates arguments and infers return types using the registry
- **Type Signature Hierarchy**: Flexible constraint system with `PositionalSignature` base class supporting exact values and custom validators
- **Extensible Type Expectation System**: Added `get_expected_types()` method to `TypeSignature` base class for future implicit casting support
- **Comprehensive Test Coverage**: Full test suite covering signature validation, return type strategies, and registry behavior

## Technical Implementation

- Return type strategies handle common patterns: fixed types, input-dependent types, promoted numeric types, and dynamic computation
- Type signature validation supports exact constraints, variadic patterns, and custom validation logic
- Foundation ready for implicit casting system in future PRs


bcallender:
* **#56** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/56?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#59](https://github.com/typedef-ai/fenic/pull/59) <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/59?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#77](https://github.com/typedef-ai/fenic/pull/77) <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/77?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#55** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/55?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#54** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/54?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#53** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/53?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/53?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/fenic/core/_logical_plan/signatures/registry.py
```diff
@@ -0,0 +1,62 @@
+"""Central registry for function signatures.
+
+This module provides a global registry where function signatures are stored
+and retrieved by function name.
+"""
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Dict, List, Optional
+
+if TYPE_CHECKING:
```
rohitrastogi: Please remove the empty TYPE_CHECKING block.

tests/_logical_plan/signatures/test_signatures.py
```diff
@@ -0,0 +1,282 @@
...
+        struct_type = StructType([StructField("field1", StringType)])
+        
+        with pytest.raises(ValidationError, match=r"test_func expects 2 arguments \(struct, field_name\), got 1"):
+            sig.validate([struct_type], "test_func")
+        
+        with pytest.raises(ValidationError, match=r"test_func expects 2 arguments \(struct, field_name\), got 3"):
+            sig.validate([struct_type, StringType, StringType], "test_func")
+
+
+if __name__ == "__main__":
```
rohitrastogi: I don't think this is required to run the test using Pytest.

src/fenic/core/_logical_plan/signatures/types.py
```diff
@@ -0,0 +1,308 @@
...
+                        expected=constraint,
+                        actual=actual_arg_type,
+                        context=f"{func_name} Argument {i}",
+                    )
+
+        # Apply custom validation if provided
+        if self.custom_validator:
+            self.custom_validator(actual_arg_types, func_name)
+
+    def get_expected_types(self, arg_types: List[DataType]) -> List[DataType]:
```
rohitrastogi: We dont need this anymore since we are deprioritizing implicit casting, right? If so, can we remove it?

src/fenic/core/_logical_plan/signatures/types.py
```diff
@@ -0,0 +1,308 @@
...
+                raise TypeMismatchError(
+                    expected=actual_array_type.element_type,
+                    actual=actual_element_type,
+                    context=f"{func_name} Argument 1",
+                )
+        
+        super().__init__([ArrayType, DataType], custom_validator=custom_validator, arg_names=["array", "element"])
+
+
+class StructWithStringKey(PositionalSignature):
```
rohitrastogi: Is this used? Will it be used more than once? If not, maybe not worth the cost of the abstraction. Instead, we can just avoid registering a signature for IndexExpr.

src/fenic/core/_logical_plan/signatures/registry.py
```diff
@@ -0,0 +1,62 @@
...
+
+from fenic.core._logical_plan.signatures.signature import FunctionSignature
+from fenic.core.error import InternalError
+
+
+@dataclass
+class FunctionRegistryEntry:
+    """Entry in the function registry."""
+    expression_class: type
+    expression_migrated: bool
```
rohitrastogi: I don't think `expression_migrated` is required if we have the optional signature as it can be derived from checking signature existence.

src/fenic/core/_logical_plan/signatures/registry.py
```diff
@@ -0,0 +1,62 @@
...
+    def get_signature(cls, func_name: str) -> FunctionSignature:
+        """Get a function signature by name."""
+        if func_name not in cls._functions:
+            raise InternalError(f"Unknown function: {func_name}")
+        if cls._functions[func_name].signature is None:
+            raise InternalError(f"No signature registered for function: {func_name}")
+        return cls._functions[func_name].signature
+
+    @classmethod
+    def is_registered(cls, func_name: str) -> bool:
```
rohitrastogi: It doesn't appear that is_registered, list_functions, get_expression_class, and has_expression_class are used anywhere. Are they required?

src/fenic/core/_logical_plan/signatures/scalar_function.py
```diff
@@ -0,0 +1,56 @@
...
+from fenic.core._logical_plan.expressions.base import LogicalExpr
+from fenic.core._logical_plan.signatures.registry import FunctionRegistry
+from fenic.core.types.datatypes import DataType
+from fenic.core.types.schema import ColumnField
+
+
+class ScalarFunction(LogicalExpr):
+    """Base class for scalar functions with signatures.
+    
+    Automatically handles children() method based on LogicalExpr arguments passed to __init__.
```
rohitrastogi: nit: The comment about handling children might be unnecessary—it’s already clear from the code. In my case, it actually made things seem more complicated than they were.

src/fenic/core/_logical_plan/signatures/types.py
```diff
@@ -0,0 +1,308 @@
...
+    def get_expected_types(self, arg_types: List[DataType]) -> List[DataType]:
+        """Get the expected types for implicit casting.
+
+        Default implementation returns arg_types (no specific expectation).
+        Override in subclasses that have specific type expectations.
+        """
+        return arg_types
+
+
+class PositionalSignature(TypeSignature):
```
rohitrastogi: I don't think the PositionalSignature abstraction adds much value and is too generic. Classes that currently extend PositionalSignature would be better off inheriting directly from TypeSignature. PositionalSignature tries to do too much, handling multiple use cases in a single abstraction. That, to me, suggests it should be broken down into more focused and specific classes.

src/fenic/core/_logical_plan/signatures/types.py
```diff
@@ -0,0 +1,308 @@
...
+
+class Numeric(TypeSignature):
+    """Arguments must be numeric types (IntegerType, FloatType, or DoubleType)."""
+
+    def __init__(self, expected_num_args: int = 1):
+        self.expected_num_args = expected_num_args
+
+    def validate(self, actual_arg_types: List[DataType], func_name: str) -> None:
+        if len(actual_arg_types) != self.expected_num_args:
+            raise ValidationError(
```
rohitrastogi: Is it possible for a user to pass the wrong number of arguments (here and for uniform)? I don't think it is based on the fact that the user API for non variadic signatures are constrained. If that's the case, maybe this is actually an InternalError because it's not the user's fault.

bcallender: the api is constrained. the user could, in theory, pass extra arguments (since we're not setting `extra="forbid"` in the ConfigDict for `validate_call()`, but since we would have to explicitly add them to `super.__init__()` call, i think you're right on this. I can change the error message

src/fenic/core/_logical_plan/signatures/types.py
bcallender: makes sense. i devolved the responsibilities back into the individual implementations. i thought they looked similar enough to consolidate the behavior, but it was a little complicated

src/fenic/core/_logical_plan/signatures/types.py
bcallender: sure, removed for now

src/fenic/core/_logical_plan/signatures/arithmetic.py
```diff
@@ -0,0 +1,24 @@
...
+    FunctionSignature,
+    ReturnTypeStrategy,
+)
+from fenic.core._logical_plan.signatures.types import Numeric
+
+
+def register_arithmetic_signatures():
+    """Register signatures for arithmetic operations."""
+    # Single registration for all arithmetic operations: numeric inputs, promoted return type
+    FunctionRegistry.register(
```
rohitrastogi: Are we sure want to register arithmetic? Don't we want to be able to tell the user which arithmetic op they did something wrong for?

Same for all of the comparison operators. Do those need a type signature for binaryexpr?

bcallender: I mean `need` is a strong qualifier. We can support more custom logging for this by tying the function name to the operator, if we want to keep it in the TypeSignature system (and same for the comparison expressions), but I see no problem with returning these to use `BinaryExpr` and forgoing their addition to `TypeSignature`, would probably make the log lines less confusing.

bcallender: i ended up just reverting the functionality to what it was before. i'm also probably just going to remove the registration of non-TypeSignature expressions (and remove the expressions from the registry altogether) if we're not going to dynamically reconstruct expressions from the function names for serde.

src/fenic/core/_logical_plan/expressions/arithmetic.py
rohitrastogi, line 1: nit: probably can revert changes to this file to make the diff less noisy

src/fenic/core/_logical_plan/expressions/basic.py
```diff
@@ -157,51 +157,34 @@ def children(self) -> List[LogicalExpr]:
...
+    def __init__(self, exprs: List[LogicalExpr]):
+        self.exprs = exprs
+        super().__init__(*exprs)
 
-    def to_column_field(self, plan: LogicalPlan) -> ColumnField:
+    def _infer_dynamic_return_type(self, arg_types: List[DataType]) -> DataType:
+        """Return StructType with fields based on argument names and types."""
         struct_fields = []
-        for arg in self.args:
+        for _i, (arg, arg_type) in enumerate(zip(self._children, arg_types, strict=False)):
```
rohitrastogi: nit: _i var is not required so we can remove the enumerate(). Also, IIUC, len(arg_types) == len(children), so we should set strict=True in zip to fail loudly if this is not the case.

src/fenic/core/_logical_plan/expressions/comparison.py
```diff
@@ -64,4 +64,4 @@ def _validate_types(self, plan: LogicalPlan):
 
     def to_column_field(self, plan: LogicalPlan) -> ColumnField:
         self._validate_types(plan)
-        return ColumnField(str(self), BooleanType)
+        return ColumnField(str(self), BooleanType)
```
rohitrastogi: nit: revert to clean up diff.

bcallender:
### Merge activity

* **Jul 14, 11:40 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/typedef-ai/fenic/53).
* **Jul 14, 11:41 PM UTC**: @bcallender merged this pull request with [Graphite](https://app.graphite.dev/github/pr/typedef-ai/fenic/53).

PR #63, opened 2025-07-07
feat: convert AggregateExpr -> AggregateFunction 

bcallender:
Following onto the previous PRs in this stack -- creates AggregateFunction as an marker class on top of ScalarFunction, but keeping all of the same validation logic. This lets us express the aggregate expressions in the type signature system, and still have logic elsewhere in the project that depends on Aggregate Expressions all extending from the same superclass for things like plan optimization and expr_converter dispatch.

### DIFF
--- a/src/fenic/_backends/local/transpiler/expr_converter.py
+++ b/src/fenic/_backends/local/transpiler/expr_converter.py
@@ -25,7 +25,6 @@ from fenic._backends.local.semantic_operators import Reduce as SemanticReduce
 from fenic._backends.local.template import TemplateFormatReader
 from fenic._backends.schema_serde import serialize_data_type
 from fenic.core._logical_plan.expressions import (
-    AggregateExpr,
     AliasExpr,
     AnalyzeSentimentExpr,
     ArithmeticExpr,
@@ -96,6 +95,7 @@ from fenic.core._logical_plan.expressions import (
     UDFExpr,
     WhenExpr,
 )
+from fenic.core._logical_plan.signatures import AggregateFunction
 from fenic.core._utils.extract import convert_extract_schema_to_pydantic_type
 from fenic.core._utils.schema import (
     convert_custom_dtype_to_polars,
@@ -270,8 +270,8 @@ class ExprConverter:
         else:
             return converted_expr.mean()
 
-    @_convert_expr.register(AggregateExpr)
-    def _convert_aggregate_expr(self, logical: AggregateExpr) -> pl.Expr:
+    @_convert_expr.register(AggregateFunction)
+    def _convert_aggregate_expr(self, logical: AggregateFunction) -> pl.Expr:
         # Special handling for AvgExpr
         if isinstance(logical, AvgExpr):
             return self._convert_avg_expr(logical)
--- a/src/fenic/core/_logical_plan/expressions/aggregate.py
+++ b/src/fenic/core/_logical_plan/expressions/aggregate.py
@@ -16,6 +16,7 @@ from fenic.core.types import (
     EmbeddingType,
 )
 
+
 class SumExpr(AggregateFunction):
     function_name = "sum"
 
--- a/src/fenic/core/_logical_plan/expressions/aggregate.py
+++ b/src/fenic/core/_logical_plan/expressions/aggregate.py
@@ -7,182 +7,94 @@ if TYPE_CHECKING:
 
 from fenic.core._logical_plan.expressions.base import LogicalExpr
 from fenic.core._logical_plan.expressions.basic import LiteralExpr
+from fenic.core._logical_plan.signatures.function_base import AggregateFunction
 from fenic.core.types import (
     ArrayType,
-    BooleanType,
     ColumnField,
+    DataType,
     DoubleType,
     EmbeddingType,
-    FloatType,
-    IntegerType,
-    StringType,
 )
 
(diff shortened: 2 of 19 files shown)


bcallender:
* **#63** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/63?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/63?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/fenic/core/_logical_plan/signatures/function_base.py
```diff
@@ -59,3 +59,14 @@ def children(self) -> List[LogicalExpr]:
     def __str__(self) -> str:
         args_str = ", ".join(str(arg) for arg in self._children)
         return f"{self.function_name}({args_str})"
+
+
+class AggregateFunction(ScalarFunction):
```
rohitrastogi: I think AggregateFunction being a child of ScalarFunction can get us in trouble. I understand _why_ (from a code perspective) it is valuable for an AggregateFunction to use the FunctionRegistry, but I don't think we want isinstance(ScalarFunction) checks to be true for aggregate functions (max, min, etc).

It'll be very nice when we move into more complex planning/query optimization to be able to use the marker to distinguish between window/scalar/aggregate functions.

bcallender: putting this here for transparency -- 

Had a discussion in-person, we agree on this not being ideal -- we will merge the downstack branches for `ScalarFunction` for now, then come back to this and see if we can come up with a better way of managing the type inheritance.

src/fenic/_backends/local/transpiler/expr_converter.py
```diff
@@ -270,7 +270,7 @@ def embedding_avg(series: pl.Series, embedding_dim: int) -> pl.Series:
             return converted_expr.mean()
 
     @_convert_expr.register(AggregateExpr)
-    def _convert_aggregate_expr(self, logical: AggregateExpr) -> pl.Expr:
+    def _convert_aggregate_expr(self, logical) -> pl.Expr:
```
rohitrastogi: We can add this annotation back, right?

src/fenic/core/_logical_plan/expressions/aggregate.py
```diff
@@ -5,167 +5,202 @@
...
-        return f"{self.agg_name}({str(self.expr)})"
+    @property
+    def validator(self) -> SignatureValidator:
+        return self._validator
 
     def children(self) -> List[LogicalExpr]:
-        return [self.expr]
+        return self._children
+
+    def __str__(self) -> str:
```
rohitrastogi: IIUC, the override for __str__ is not required since they are provided by the new mixin. Can we remove it so its clear?

src/fenic/core/_logical_plan/expressions/aggregate.py
```diff
@@ -5,167 +5,202 @@
...
+
 
+class AvgExpr(ValidatedDynamicSignature, AggregateExpr):
+    function_name = "avg"
 
-class SumExpr(AggregateExpr):
     def __init__(self, expr: LogicalExpr):
-        super().__init__("sum", expr)
+        self.expr = expr
+        self._children = [expr]
```
rohitrastogi: here and elsewhere, we don't need this _children member, right?

src/fenic/core/_logical_plan/expressions/base.py
```diff
@@ -48,6 +49,114 @@ def children(self) -> List[LogicalExpr]:
...
+class SemanticExpr(LogicalExpr):
+    """Marker class for semantic expressions that use LLM models."""
+    
+    @abstractmethod
+    def _validate_completion_parameters(self, plan: LogicalPlan):
+        """Common validation for semantic functions."""
+        pass
+
+
+class SemanticAggregateExpr(AggregateExpr, SemanticExpr):
```
rohitrastogi: I don't think we need this marker. Since we aren't doing an instanceof with it now, let's get rid of it.

rohitrastogi (commented):
This approach definitely works well. Only minor gripe I have with this is that I think generally speaking, mixins are supposed to only do one thing (validation, logging, printing, etc) instead of doing several things (in our case, _str_, children(), to_column_field()), but I think it is okay and not worth having a separate __str__ mixin for the standard print implementation.

src/fenic/core/_logical_plan/expressions/base.py
```diff
@@ -48,6 +49,114 @@ def children(self) -> List[LogicalExpr]:
...
+    dynamic return type inference.
+    
+    Expressions using this mixin must implement the validator property and children() method.
+    """
+    
+    function_name: str  # Type hint to indicate this must be provided by subclass
+    
+    @property
+    @abstractmethod
+    def validator(self):
```
rohitrastogi: can we add a type hint for the return type here and for validator() in ValidatedDynamicSignature?

bcallender:
### Merge activity

* **Jul 16, 6:57 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/typedef-ai/fenic/63).
* **Jul 16, 6:57 PM UTC**: @bcallender merged this pull request with [Graphite](https://app.graphite.dev/github/pr/typedef-ai/fenic/63).

PR #66, opened 2025-07-08
feat: Add full support for complex Pydantic models in `semantic.extract`; deprecate custom `ExtractSchema`

rohitrastogi:
This PR introduces several improvements and simplifications to the `semantic.extract` system:

1. **Support complex nested Pydantic models**  
   Validation now correctly walks Pydantic annotations, including support for nested models, lists, and optionals. Previously, only primitive types were handled correctly.  
   - **Unsupported types**: Non-optional `Union` types (not supported by Fenic’s strongly typed columnar system) and dictionaries (may be supported via a `JsonType` in the future; we don’t yet support map types).

2. **Improve the extraction system prompt**  
   The system prompt for `semantic.extract` has been rewritten to provide clearer guidance to the LLM. It now emphasizes extraction *only* from present data and discourages speculation or inference.

3. **Enhance model-to-prompt generation**  
   The logic for generating natural language descriptions from Pydantic types now correctly handles `Literal` and `Optional` types.

4. **Deprecate the custom `ExtractSchema` API**  
   With Pydantic now offering full feature parity—and more—the custom `ExtractSchema` has been deprecated.  
   - The Pydantic model approach is more familiar to users already working with structured outputs and tool-calling.
   - Removing `ExtractSchema` simplifies maintenance and allows for better support of advanced Python typing features.

5. **Update all demos and examples**  
   All references to `ExtractSchema` have been removed and replaced with Pydantic-based schemas.

6. **Refactor and clean up unit tests**  
   The `semantic.extract` test suite has been streamlined to ensure tests are focused, non-overlapping, and easier to maintain.

### DIFF
--- a/src/fenic/logging.py
+++ b/src/fenic/logging.py
@@ -4,6 +4,8 @@ import logging
 import sys
 from typing import Optional, TextIO
 
+NOISY_LOGGER_NAMES = ("openai", "httpx", "google_genai")
+
 
 def configure_logging(
     log_level: int = logging.INFO,
@@ -37,7 +39,7 @@ def configure_logging(
         root_logger.addHandler(handler)
 
         # Silence noisy dependencies
-        for noisy_logger_name in ("openai", "httpx"):
+        for noisy_logger_name in NOISY_LOGGER_NAMES:
             noisy_logger = logging.getLogger(noisy_logger_name)
             noisy_logger.setLevel(logging.ERROR)
 


src/fenic/core/_utils/extract.py
```diff
@@ -1,170 +1,72 @@
...
-
-    Args:
-        schema: The ExtractSchema defining the structure
-
-    Returns:
-        A Pydantic model type that matches the schema structure
-    """
-    return _create_nested_model(schema=schema)
 
+class ExtractSchemaValidationError(Exception):
```
rohitrastogi, line 8: note for reviewer: this is an important change

src/fenic/core/_utils/schema.py
```diff
@@ -255,3 +247,21 @@ def _convert_polars_dtype_to_custom_dtype(
         return StringType
     else:
         raise ValueError(f"Unsupported Polars data type: {polars_dtype}")
+
+def _unwrap_optional_type(annotation) -> type:
```
rohitrastogi, line 251: note for reviewer: important change

src/fenic/_backends/local/semantic_operators/extract.py
```diff
@@ -26,13 +27,25 @@
...
+
+        "The field schema below defines the structure of the information you are expected to extract.\n\n"
+
+        "How to read the field schema:\n"
+        "- Nested fields are expressed using dot notation (e.g., 'organization.name' means 'name' is a subfield of 'organization')\n"
+        "- Lists are denoted using 'list of [type]' (e.g., 'employees' is a list of [string])\n"
+        "- Type annotations are shown in parentheses (e.g., string, integer, boolean, date)\n\n"
+
+        "Extraction Guidelines:\n"
+        "1. Extract only what is explicitly present or clearly supported in the document—do not guess or extrapolate.\n"
```
rohitrastogi, line 41: note for reviewer: important change

src/fenic/_backends/local/semantic_operators/extract.py
```diff
@@ -96,69 +109,75 @@ def convert_pydantic_model_to_key_descriptions(schema: Type[BaseModel]) -> str:
...
 
     def recurse(schema: Type[BaseModel], prefix: str = ""):
-        """Recursively extract field information from a model."""
         for field_name, field_info in schema.model_fields.items():
             full_field_name = f"{prefix}.{field_name}" if prefix else field_name
             annotation = field_info.annotation
-            description = getattr(field_info, "description", "") or ""
+            description = field_info.description or ""
+
+            # Unwrap Annotated
```
rohitrastogi, line 144: note for reviewer: important change

examples/document_extraction/document_extraction.py
```diff
@@ -66,73 +65,14 @@ def main(config: Optional[fc.SessionConfig] = None):
...
-    print()
-
-    # Method 2: Using Pydantic Model
-    print("Method 2: Pydantic Model Approach")
-    print("-" * 40)
-
     # Define Pydantic model for document metadata
-    # Note: Pydantic models for extraction support simple data types (str, int, float, bool, Literal)
-    # Complex types like lists must be represented as strings (e.g., comma-separated values)
     class DocumentMetadata(BaseModel):
```
bcallender, line 69: while you're updating these, can you move the class definitions out of the function? that will make these serializable with our current `cloudpickle` implementation that doesn't.

src/fenic/_backends/local/semantic_operators/extract.py
```diff
@@ -26,13 +27,25 @@
...
+        "- Type annotations are shown in parentheses (e.g., string, integer, boolean, date)\n\n"
+
+        "Extraction Guidelines:\n"
+        "1. Extract only what is explicitly present or clearly supported in the document—do not guess or extrapolate.\n"
+        "2. For list fields, extract all items that match the field description.\n"
+        "3. If a field is not found in the document, return null for single values and [] for lists.\n"
+        "4. Ensure all field names in your structured output exactly match the field schema.\n"
+        "5. Be thorough and precise—capture all relevant content without changing or omitting meaning.\n\n"
+
+        "Field Schema:\n"
```
bcallender, line 47: do we even still need to provide the schema here if we are enforcing the structured output via tool or `response_format` that accepts the underlying pydantic model and the user provided descriptions? does the model produce worse results if this is not provided in the prompt? unless this is for being able to support models for which we can't pass a structured schema to?

src/fenic/core/_utils/extract.py
```diff
@@ -1,170 +1,72 @@
...
+        raise ExtractSchemaValidationError(
             "Output schema cannot be empty. "
             "Please specify at least one output field."
         )
+
     for field_name, field_info in model.model_fields.items():
         if field_info.description is None:
-            raise ValueError(
-                f"Extract schema field {field_name} has no description.  Please specify a description for each field."
+            raise ExtractSchemaValidationError(
```
bcallender, line 24: especially if we are enforcing that each pydantic field _must_ have a description, it seems like even less reason to need to walk the class and add the field schema to the prompt.

src/fenic/core/_utils/extract.py
```diff
@@ -1,170 +1,72 @@
...
+    """Check a Pydantic model type   to ensure it is valid schema for a semantic extract."""
     # Check the field structure and the types for the pydantic model
     if len(model.model_fields.items()) == 0:
-        raise ValueError(
+        raise ExtractSchemaValidationError(
             "Output schema cannot be empty. "
             "Please specify at least one output field."
         )
+
     for field_name, field_info in model.model_fields.items():
```
bcallender: it seems that `model_fields()` is deprecated, we might want to use `model.__pydantic_fields__` directly to maintain future compatibility

src/fenic/core/_utils/extract.py
```diff
@@ -1,170 +1,72 @@
...
 
-    Returns:
-        The corresponding Python type
+    # Handle generic types (List, Optional, Union, etc.)
+    origin = get_origin(annotation)
+    args = get_args(annotation)
 
-    Raises:
-        ValueError: If the data type cannot be converted
+    if origin is list or origin is List:
```
bcallender, line 45: do we care about supporting set or tuple here? i can't really think of a use case for tuple, but set could be useful for removing duplicate values in the collection

bcallender (commented):
just a first pass, have to jump to a 1:1

src/fenic/_backends/local/semantic_operators/extract.py
rohitrastogi: I don't think that the descriptions provided in the BaseModel aren't actually used by structured outputs (so yes, results are worse without them). Guided decoding is just for the output types. You don't pay for the description tokens so we still need this to actually tell the model what to do.

src/fenic/core/_utils/extract.py
rohitrastogi: see above

src/fenic/core/_utils/extract.py
rohitrastogi: sets are not a json type. Users can always dedupe lists using the standard Fenic apis

examples/document_extraction/document_extraction.py
rohitrastogi: can be done in a follow up

src/fenic/core/_utils/extract.py
rohitrastogi: done

src/fenic/api/functions/semantic.py
```diff
@@ -92,66 +93,46 @@ def map(
...
     structured data based on the fields and descriptions provided. Useful for pulling out key entities,
     facts, or labels from documents.
 
     Args:
         column: Column containing text to extract from.
-        schema: An ExtractSchema containing fields of type ExtractSchemaField that define
-            the output structure and field descriptions or a Pydantic model that defines the output structure with
-            descriptions for each field.
-        model_alias: Optional alias for the language model to use for the mapping. If None, will use the language model configured as the default.
+        schema: A Pydantic model type that defines the output structure with descriptions for each field.
```
bcallender, line 119: it would be nice to be more clear to the user some of other common `BaseModel` validation failure issues (must have typed collections, no `Union` types other than `Optional` etc.)  , either here, or by providing negative examples

bcallender: i don't know if the docstring is the right place, or if we should add some sort of `Best Practices` section to the documentation (as a follow on) for each operator to describe dos and don'ts the best results, like not using overly nested structures, avoiding circular references in nested structures, etc

src/fenic/core/_utils/extract.py
bcallender: could be useful to have this as a `FenicError`, as a subtype of `ValidationError` or `PlanError`

src/fenic/core/_utils/extract.py
```diff
@@ -1,170 +1,72 @@
...
-        else:
-            annotated_fields[field.name] = Annotated[
-                _convert_primitive_dtype_to_pytype(field.data_type),
-                Field(description=field.description),
-            ]
-    return create_model(model_name, **annotated_fields)
+    # If we get here, it's an unsupported type
+    raise ExtractSchemaValidationError(
+        f"Unsupported data type in extract schema field {field_name}: {annotation}"
+    )
```
bcallender, line 73: maybe a hint what is allowed for the user, like 
```
    f"Supported types: str, int, float, bool, List[T], Optional[T], Literal, and nested Pydantic models."
```
?

bcallender (approved):
this makes sense to me, should be quite helpful. it is a breaking change to the API though. i think we're still early enough that we can get away with doing it in a minor release, but we'll have to think carefully about it in the future

rohitrastogi:
I think we're free to make breaking api changes until 1.0

src/fenic/core/_utils/extract.py
rohitrastogi: FYI, this isn't a user facing error. It's just a temporary error that we use to catch and transform into a ValidationError in the API layer.

src/fenic/api/functions/semantic.py
rohitrastogi: updated the docstring

PR #75, opened 2025-07-10
fix: fix bugs in text.extract related to delimiter sequence parsing and also disallow empty column names in templates

rohitrastogi:
## Refactor `text.extract` - Fix Bugs, Improve Performance & Maintainability

This PR improves `text.extract` by fixing critical parsing bugs and refactoring template parsing to make it more maintainable. It also massively improves test coverage.

### Key Bug Fixes

1. **Fixed delimiter matching logic**: `TemplateFormatReader` was incorrectly checking against all possible future delimiters simultaneously instead of matching them sequentially. This caused extraction failures when delimiter text appeared inside field content before the actual delimiter.
  - **Example**: Template `"[${timestamp}] ${level} - ${component}: ${message}"` would fail when reading timestamp if the data contained " - " before the closing bracket
  
2. **Disallowed empty column names**: Templates like `"prefix${}suffix"` previously created unnamed columns, which aren't allowed in DataFrames.

### Functional Improvements

- **Better type system integration**: JSON fields now return `JsonType` instead of `StringType`, enabling use with JSON-specific functions like `jq()`. Now returns `None` for invalid JSON/quoted fields instead of raising exceptions, providing better consistency with other fallible methods.
- Before, implementation only worked on a single line. Now, treat new lines as any other character in the template or input string.

### Refactoring Template Parsing

**Previous approach**: Single-pass state machine mixing parsing logic with state management  
**New approach**: Clean two-phase pipeline:

1. **Phase 1 - Tokenization**: Break format string into alternating TEXT/COLUMN tokens using efficient string operations
2. **Phase 2 - Processing**: Convert tokens into structured delimiter/column/escaping data

**Key invariant**: `len(delimiters) == len(columns) + 1` (delimiters surround columns)

### Refactoring Template Extraction

- **Sequential delimiter matching**: Instead of checking all possible delimiters for each character, now looks for the specific next delimiter for each field


src/fenic/api/functions/text.py
```diff
@@ -33,39 +33,57 @@
...
+
+    Matches each string in the input column against a template pattern with named
+    placeholders. Each placeholder can specify a format rule to handle different
+    data types within the text.
 
     Args:
-        template: Template string with fields marked as ``${field_name:format}``
         column: Input text column to extract from
+        template: Template string with placeholders as ``${field_name}`` or ``${field_name:format}``
+                 Available formats: none, csv, json, quoted, regex, qualified
```
german-typedef: nit: do you have an example of regex and qualified?

tests/_backends/local/functions/test_string_functions.py
```diff
@@ -25,73 +26,383 @@ def test_textract_basic(local_session):
...
+            '[2024-01-15 10:30:45] "ERROR" - DatabaseManager: "Connection timeout occurred" {"retry_count": 3, "duration_ms": 5000}',
+            '[2024-01-15 10:31:02] "INFO" - AuthService: "User login successful" {"user_id": 12345, "session_duration": 1800}',
+            '[2024-01-15 10:31:15] "WARN" - PaymentProcessor: "Rate limit ""exceeded""" {"requests_per_minute": 150, "limit": 100}',
+            '[2024-01-15 10:31:30] "DEBUG" - CacheManager: "Cache miss for key" {"cache_hit_ratio": 0.85, "keys_checked": 200}',
         ]
     }
     df = local_session.create_dataframe(data)
-    template = "Name: ${name:csv}, Age: ${age:none}"
 
+    template = '[${timestamp}] ${level:csv} - ${component}: ${message:csv} ${metadata:json}'
```
german-typedef, line 51: question: for my own understanding, could "level" have "none" as the format for this particular case?

src/fenic/api/functions/text.py
rohitrastogi: those actually aren't implemented yet, so thanks for catching this.

rohitrastogi: i removed them from the docstring.

cpard (approved):

tests/_backends/local/functions/test_string_functions.py
rohitrastogi: if you used "none", it would include the quotes in the extracted column. you'd get ['"ERROR"', ....] instead of ["ERROR"]

PR #59, opened 2025-07-07
feat: Implicit Casting for Type Signature System

bcallender:

PR #56, opened 2025-07-03
feat: convert semantic/embedding exprs to ScalarFunction

bcallender:
# feat: Convert semantic and embedding functions with specialized validation

## Summary

Completes the function migration by converting semantic (LLM-based) and embedding operations to use ScalarFunction validation, introducing specialized validation for AI/ML operations with completion parameter validation.

## Key Changes

- **Semantic Function Signatures**: Registered 7 semantic functions (`semantic.map`, `semantic.extract`, `semantic.predicate`, etc.) with AI-specific type constraints and parameter validation
- **Embedding Function Signatures**: Registered embedding similarity computation with flexible input pattern handling
- **SemanticFunction Base Class**: Specialized ScalarFunction subclass that validates LLM completion parameters (model_alias, temperature, max_tokens) against session configuration
- **Complex Type Patterns**: Advanced signature patterns using `VariadicUniform` for multi-argument semantic operations and `OneOf` for flexible embedding inputs
- **AI Parameter Validation**: Automatic validation of LLM parameters against model capabilities and session limits

## Technical Implementation

- **SemanticFunction Class**: Inherits from ScalarFunction and adds completion parameter validation
- **Advanced Signatures**: Uses `VariadicUniform(StringType, expected_min_args=2)` for semantic operations requiring multiple string inputs
- **Embedding Flexibility**: `OneOf` signatures handle different embedding input patterns (single arrays, array pairs for similarity)

## Optimizer Integration

- Updated semantic filter rewrite optimizer to recognize `SemanticFunction` instead of deprecated `SemanticExpr`
- Maintains all existing semantic optimization behaviors with new type system
- Cleaner semantic expression recognition pattern

## Specialized Validation Benefits

- Model parameter validation prevents runtime failures with invalid model configurations
- Type safety for complex multi-argument semantic operations  
- Consistent error messages for AI-specific validation failures
- Foundation for implicit casting in AI workflows (string → MarkdownType, arrays → EmbeddingType)

## Migration Impact

- Semantic expressions simplified by removing duplicate validation
- Embedding operations get automatic type checking for complex input patterns
- Model parameter validation moved to specialized base class for reusability

PR #55, opened 2025-07-03
feat: convert json/markdown functions to ScalarFunction

bcallender:
# feat: Convert JSON and markdown functions to ScalarFunction validation

## Summary

Continues the migration pattern by converting JSON and markdown processing functions to use the centralized signature validation system, expanding coverage to structured and markup data operations.

## Key Changes

- **JSON Function Signatures**: Registered 3 JSON functions (`json.get`, `json.get_list`, `json.extract`) with flexible type handling for JSON operations
- **Markdown Function Signatures**: Registered 2 markdown functions (`markdown.to_html`, `markdown.extract_text`) with markdown-specific type constraints
- **Expression Migration**: Converted `JsonGetExpr`, `JsonGetListExpr`, `JsonExtractExpr`, `MarkdownToHtmlExpr`, `MarkdownExtractTextExpr` to ScalarFunction pattern
- **Type Flexibility**: JSON functions handle mixed input types (JsonType, StringType) while markdown functions enforce MarkdownType constraints

## Technical Implementation

- JSON functions use `OneOf` signatures to accept both JsonType and StringType inputs, reflecting real-world usage patterns
- Markdown functions use `Exact` signatures for type safety with MarkdownType
- `DYNAMIC` return type strategy for `json.extract` (return type depends on extraction query)
- `SAME_AS_INPUT` for `markdown.to_html` (MarkdownType → StringType conversion)

## Data Type Handling

- JSON operations support flexible input types for compatibility with different data sources
- Markdown operations maintain type safety while enabling structured document processing
- Proper return type inference for complex extraction operations

## Code Simplification

- Removed duplicate validation logic from JSON expressions
- Eliminated manual type checking from markdown expressions
- Consistent error handling across structured data operations

## Dependencies

Requires previous branches in stack

PR #54, opened 2025-07-03
feat: register/convert text functions using ScalarFunction

bcallender:
# feat: Convert text processing functions to ScalarFunction-based validation

## Summary

Migrates all text processing functions from manual type validation to the new centralized signature system, demonstrating the ScalarFunction pattern and providing immediate benefits for text operations.

## Key Changes

- **Text Function Signatures**: Registered _all_ expressions in `text.py` to use the new `FunctionRegistry` and inherit from `ScalarFunction` 
- **Automatic Type Validation**: Text functions now get automatic argument validation and return type inference through the registry
- **Test Coverage**: Updated existing text function tests to work with new validation system

## Technical Details

- Text functions use `Exact` type signatures for precise string input requirements
- `SAME_AS_INPUT` return type strategy for case conversion functions (preserves StringType)
- Fixed return types for length (IntegerType) and contains (BooleanType) operations
- Expressions simplified by removing duplicate validation logic

## Migration Pattern

```python
# Before: Manual validation in each expression
class UpperExpr(LogicalExpr):
    def to_column_field(self, plan):
        # Manual type checking logic
        if arg_type != StringType:
            raise TypeMismatchError(...)
        return ColumnField(name=str(self), data_type=StringType)

# After: Automatic validation via ScalarFunction
class UpperExpr(ScalarFunction):
    function_name = "text.upper"
    # Registry handles all validation and type inference
```

## Benefits

- Eliminates duplicate validation code
- Consistent error messages across all text functions  
- Foundation for implicit casting (string-like types → StringType)
- Easier to add new text functions with guaranteed consistency

## Dependencies

Requires `07-03-feat_function_registry_for_signature_validation`

PR #77, opened 2025-07-12
feat: semantic operation instructions can now reference arbitrary column expressions

bcallender:
### TL;DR

Added backwards-compatible support for dynamic column bindings in semantic functions, allowing for complex expressions and just-in-time transformations in semantic operations. For now, only `semantic.map()`, `semantic.predicate()`, and `semantic.reduce()` support this new functionality, `semantic.join()` will be added in a follow up PR. 

```python
# Example with bindings for concatenated columns
from fenic.api.functions.text import concat

template = InstructionTemplate(
                "Is {user_post} inappropriate?",
                user_post=concat(col("username"), lit(": "), col("content"))
)
semantic.predicate(instruction=template)


# Example with nested semantic operations
template = InstructionTemplate(
   "Summarize {category} content: {title} - {content}"
    category=semantic.classify("content", ["AI", "ML", "Data Science"])
)
semantic.reduce(template)
```

### What changed?

- Added made the `instruction` parameter to `semantic.map()`, `semantic.predicate()`, and `semantic.reduce()` functions accept a new `InstructionTemplate` class, which encapsulates the validation/resolution of bindings to `LogicalExpr`
- InstructionTemplate supports placeholder deduplication to optimize performance when the same placeholder appears multiple times
- Improved validation to ensure alias names match placeholder keys, and will automatically alias column expressions to match their placeholder key names

### How to test?

The PR includes comprehensive tests for all modified functions:

Run the new test files:
- `tests/_backends/local/functions/test_semantic_map.py`
- `tests/_backends/local/functions/test_semantic_predicate.py`
- `tests/_backends/local/functions/test_semantic_reduce.py`
- `tests/api/functions/test_instruction_template.py`

### Why make this change?

This improves the flexibility of semantic functions by:

1. Enabling just-in-time transformations of data before it's passed to LLM operations
2. Supporting complex expressions like concatenations, computations, and nested semantic operations
3. Allowing for more sophisticated prompt engineering with dynamically computed values
4. Optimizing performance by ensuring expressions are only evaluated once, even when referenced multiple times

These changes make it easier to build complex semantic workflows without having to create intermediate columns.


bcallender:
* **#63** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/63?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#78** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/78?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#77** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/77?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/77?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#56** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/56?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#59](https://github.com/typedef-ai/fenic/pull/59) <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/59?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#55** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/55?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#54** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/54?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#53** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/53?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/fenic/api/functions/semantic.py
```diff
@@ -44,6 +50,9 @@ def map(
             The instruction must include placeholders in curly braces that reference one or more column names.
             These placeholders will be replaced with actual column values during prompt construction during
             query execution.
+        bindings: Optional mapping of placeholder names to column expressions.
+            Only required when placeholder names don't match existing column names.
+            Enables just-in-time transformations like concatenations, computations, or nested semantic operations.
```
rohitrastogi: Personally don't feel the need to mention "Enables ..." - The fact that the signature is a Column is enough to indicate that columns can be derived because that's the way it works for every other function.

src/fenic/api/functions/semantic.py
```diff
@@ -417,4 +473,49 @@ def summarize(
...
+
+            # Check if the expression is already aliased
+            if isinstance(logical_expr, AliasExpr):
+                # Validate that the alias matches the key
+                if logical_expr.name != placeholder_key:
+                    raise ValidationError(f"Alias name must match the key. Expected '{placeholder_key}', got '{logical_expr.name}'")
+                exprs.append(logical_expr)
+            else:
+                # Check if the expression has a name attribute that doesn't match
+                if hasattr(logical_expr, 'name') and logical_expr.name != placeholder_key:
```
rohitrastogi: Why is this check required if we are going to auto alias anyways? Also, as a rule of thumb, I think we should avoid hasattr unless only when required because if you refactor member names, hasattrs are not automatically refactored, leading to annoying bugs.

tests/_backends/local/functions/test_semantic_map.py
```diff
@@ -80,3 +84,195 @@ def test_semantic_map_with_nulls(local_session):
...
+        "name": pl.String,
+        "city": pl.String,
+        "weather_report": pl.String,
+    }
+    assert len(result) == 2
+    for val in result["weather_report"].to_list():
+        assert isinstance(val, str) and len(val) > 0
+
+
+def test_semantic_map_optimization_works(local_session):
```
rohitrastogi: This test doesn't test anything new (given its the way every other expr works), so consider removing it.

tests/_backends/local/functions/test_semantic_map.py
```diff
@@ -80,3 +84,195 @@ def test_semantic_map_with_nulls(local_session):
...
+        "first_name": pl.String,
+        "last_name": pl.String,
+        "city": pl.String,
+        "greeting": pl.String,
+    }
+    assert len(result) == 2
+    for val in result["greeting"].to_list():
+        assert isinstance(val, str) and len(val) > 0
+        # Should contain the full name
+        assert "Alice Smith" in val or "Bob Jones" in val
```
rohitrastogi: This may cause flaky tests since the model output is non-deterministic (here and elsewhere)

tests/_backends/local/functions/test_semantic_map.py
```diff
@@ -80,3 +84,195 @@ def test_semantic_map_with_nulls(local_session):
...
+    df = source.select(
+        col("first_name"),
+        col("last_name"),
+        semantic.map(
+            instruction=prompt,
+            bindings=nested_expr
+        ).alias("desc")
+    )
+    result = df.to_polars()
+    assert result.schema == {
```
rohitrastogi, line 133: Here and elsewhere we should assert that the Fenic schema is also correct.

src/fenic/core/_logical_plan/expressions/semantic.py
```diff
@@ -67,17 +66,15 @@ class SemanticMapExpr(SemanticFunction):
     def __init__(
             self,
             instruction: str,
+            instruction_exprs: list[LogicalExpr],
```
rohitrastogi: nit: For documentation's sake, we can probably just make this Union[ColumnExpr, AliasExpr] instead of LogicalExpr

tests/_backends/local/functions/test_semantic_map.py
```diff
@@ -80,3 +84,195 @@ def test_semantic_map_with_nulls(local_session):
...
+        "greeting": pl.String,
+    }
+    assert len(result) == 2
+    for val in result["greeting"].to_list():
+        assert isinstance(val, str) and len(val) > 0
+        # Should contain the full name
+        assert "Alice Smith" in val or "Bob Jones" in val
+
+
+def test_semantic_map_placeholder_deduplication_and_single_evaluation(local_session):
```
rohitrastogi: FYI, just because we have the same Fenic Expr for duplicated columns, doesn't mean that the expression will only be evaluated once during execution. AFAIK, Polars doesn't do common subexpression elimination in Eager mode and does in Lazy mode. If we continue to use eager mode as our execution engine, we have to implement CSE ourselves.

tests/_backends/local/functions/test_semantic_map.py
```diff
@@ -80,3 +84,195 @@ def test_semantic_map_with_nulls(local_session):
...
+        ).alias("greeting")
+    )
+    result = df.to_polars()
+    assert len(result) == 2
+    for val in result["greeting"].to_list():
+        assert isinstance(val, str) and len(val) > 0
+    plan = df._logical_plan
+    semantic_map_expr = None
+    for expr in plan.exprs():
+        if hasattr(expr, 'expr') and isinstance(expr.expr, SemanticMapExpr):
```
rohitrastogi: as mentioned above, best to avoid hasattr/getattr. We should be able to write an equivalent test just using isinstance checks.

tests/_backends/local/functions/test_semantic_map.py
bcallender: that makes sense. this test was there to ensure that the expression at least doesn't appear twice in the children, which was an issue i was seeing before I added deduplication in the step that extracts the placeholders from the instruction, more of a regression test to ensure that we would catch it if that behavior changed. i can change the name to be less confusing though.

src/fenic/api/functions/semantic.py
bcallender: i think this is vestigial from a time before i did the auto aliasing, removed.

tests/_backends/local/functions/test_semantic_map.py
bcallender: fixed, across all the tests

rohitrastogi:
Also qq: is semantic.join in scope for this work?

bcallender:
In general -- yes, but it will be part of a separate PR, since that involves rewriting a bunch of the type checking logic in SemanticJoin's `_validate_columns`

PR #78, opened 2025-07-12
feat: semantic map can now generate content with pydantic schema

bcallender:
### TL;DR

Added schema parameter to `semantic.map()` to support structured output using Pydantic models.

### What changed?

- Added an optional `schema` parameter to the `semantic.map()` function that accepts a Pydantic `BaseModel` to define the structure of the output
- Updated the function signature to pass the schema as `response_format` to the underlying implementation
- Modified the return type inference to dynamically determine the return type based on the provided schema
- Changed the function registry to use `ReturnTypeStrategy.DYNAMIC` instead of a fixed `StringType`
- Added support for the `MapExample` to be constructed with a `BaseModel` as the output if the schema is set. 
- Added documentation and example showing how to use the schema parameter with a Pydantic model

### How to test?

```python
from pydantic import BaseModel, Field

# Define a schema for structured output
class EmailResponse(BaseModel):
    subject: str = Field(description="A concise, professional email subject line")
    body: str = Field(description="A complete, empathetic email body")
    urgency_level: str = Field(description="Assessment of how quickly this requires follow-up")

# Use the schema with semantic.map
result = semantic.map(
    "Generate a customer service email response for: {support_ticket}", 
    schema=EmailResponse
)
```

### Why make this change?

This enhancement allows users to get structured, typed outputs from semantic.map operations instead of just strings. By defining a Pydantic model with field descriptions, users can guide the model to generate responses in a specific format with specific components, making the outputs more consistent and easier to work with programmatically.


bcallender:
* **#78** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/78?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/78?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

rohitrastogi:
Driveby: We need to add the field descriptions to the prompt conditionally like we do semantic.extract because descriptions are not used in structured outputs.

bcallender:
yep, already on it

rohitrastogi:
I think we need to update the examples API to accept instances of BaseModels instead of just string outputs.

src/fenic/core/_logical_plan/expressions/semantic.py
```diff
@@ -80,18 +81,36 @@ def __init__(
...
             examples._validate_with_instruction(instruction)
             self.examples = examples
 
         # Pass all parsed column expressions to signature validation
         super().__init__(*self.exprs)
 
+    def _validate_example_response_format(self, example_collection: MapExampleCollection):
+        for example in example_collection.examples:
+            if self.response_format is None and not isinstance(example.output, str):
+                raise ValidationError("If a `schema` is not provided to `semantic.map`, "
```
bcallender: should this be `ValidationError` or `InvalidExampleCollectionError`?

src/fenic/_backends/local/semantic_operators/map.py
```diff
@@ -28,7 +32,19 @@ class Map(BaseMultiColumnInputOperator[str, str]):
         "The instruction will reference the context fields using square brackets [LIKE_THIS]. "
         "Each context field will be labeled with its name in square brackets, matching the references in the instruction. "
         "Your response should fulfill the instruction by appropriately integrating each of the referenced context fields without using any external information. "
-        "Your response should not include unnecessary preamble or explanation. " # anthropic LOVES a friendly preamble.
+        "Your response should not include unnecessary preamble or explanation."
+    )
+
+    STRUCTURED_SYSTEM_PROMPT_PREFIX = (
+        "You are an AI assistant designed to follow instructions and generate structured output. "
```
rohitrastogi: I'm not sure we need two different system prompt prefixes for this given that we will thread in the output schema for structured outputs.

src/fenic/_backends/local/semantic_operators/utils.py
```diff
@@ -65,3 +83,197 @@ def filter_invalid_embeddings_expr(embedding_column: str) -> pl.Expr:
...
+        return validated_model.model_dump(mode="json")
+    except Exception as e:
+        logger.warning(
+            f"invalid model output: {json_resp} for {operator_name}: {e}",
+            exc_info=True,
+        )
+        return None
+
+
+def build_schema_prompt_section(
```
rohitrastogi: What do you think of having two separate free functions for build_schema_prompt_section (one for extract, one for map), which  can  live in the extract/map files respectively. This keeps all of the prompt generation logic for an operator neatly encapsulated in the same file.

src/fenic/core/_logical_plan/expressions/semantic.py
rohitrastogi: Let's say InvalidExampleCollectionError since the context of the error is in example building.

tests/_backends/local/semantic_operators/test_map.py
```diff
@@ -45,7 +45,7 @@ def test_build_prompts(self, local_session):
         expected = [
             [
                 {
-                    "content": sem_map.SYSTEM_PROMPT,
+                    "content": sem_map.build_system_message(),
```
rohitrastogi: Now that this does logic, I think I'd prefer the test assert against the actual string output instead of the actual function call. It's also useful for documentation for folks to see how the prompts are actually generated.

src/fenic/_backends/local/semantic_operators/map.py
rohitrastogi: Well, it will never hurt to give more specific instructions.
I wonder if we can re-use more of the prompt prefix and do conditional interpolation of the ("generate structured output..."). Should be easy enough to do without Jinja, but we also have jinja as a dependency now that we can use internally when munging the prompts.

src/fenic/core/_utils/schema.py
```diff
@@ -265,3 +265,92 @@ def _unwrap_optional_type(annotation) -> type:
 
     # Not Optional
     return annotation
+
+
+# =============================================================================
```
rohitrastogi: nit: I'm thinking that this schema.py file has become something of a dumping ground for utilities and it's been a bit annoying to maintain. What do you think about creating a new utility file called structured_outputs.py for this?

src/fenic/core/types/semantic_examples.py
```diff
@@ -497,3 +526,26 @@ def _as_df_input(self) -> List[Dict[str, Any]]:
             )
 
         return rows
+
+# when we add semantic.extract examples, this signature can change to
+# def validate_single_example_output_type(existing_examples: list[Union[MapExample, ExtractExample]], example: Union[MapExample, ExtractExample]):
+def _validate_single_example_output_type(existing_examples: list[MapExample], example: MapExample):
```
rohitrastogi, line 540: nit: I think this function can be made more readable if we use a guard clause to return early if there are no existing examples to reduce nesting.

src/fenic/_backends/local/semantic_operators/map.py
bcallender: re did this with jinja, with separate templates in each operator, as suggested below

src/fenic/_backends/local/semantic_operators/utils.py
bcallender: see above

tests/_backends/local/semantic_operators/test_map.py
bcallender: done

src/fenic/core/_utils/schema.py
bcallender: done

src/fenic/core/_utils/structured_outputs.py
```diff
@@ -1,41 +1,52 @@
-"""Utility functions for working with structured extraction schemas."""
 import typing
 from typing import List, Union, get_args, get_origin
 
 from pydantic import BaseModel
 
 
-class ExtractSchemaValidationError(Exception):
-    """Error raised when an extract schema is invalid."""
+class SemanticSchemaValidationError(Exception):
```
rohitrastogi: nit: maybe OutputFormatValidationError?

src/fenic/core/_utils/structured_outputs.py
```diff
@@ -1,41 +1,52 @@
...
+    """Error raised when a semantic operation schema is invalid."""
 
-def validate_extract_schema_structure(
+
+def validate_semantic_schema_structure(
```
rohitrastogi: nit: maybe validate_output_format?

rohitrastogi: Don't we need to call this in api/ for semantic.map if an output format is provided?

src/fenic/core/types/semantic_examples.py
```diff
@@ -191,12 +191,36 @@ class MapExampleCollection(BaseExampleCollection[MapExample]):
...
     Examples in this collection can have multiple input variables, each mapped
-    to their respective values, with a single output string representing the
-    expected transformation result.
+    to their respective values, with a single output string or structured model
+    representing the expected transformation result.
     """
 
     example_class = MapExample
 
+    def create_example(self, example: MapExample) -> "MapExampleCollection":
```
rohitrastogi: we can use this as an opportunity to remove the quotes.

src/fenic/core/types/semantic_examples.py
```diff
@@ -191,12 +191,36 @@ class MapExampleCollection(BaseExampleCollection[MapExample]):
...
+            example: The MapExample to add.
+
+        Returns:
+            Self for method chaining.
+
+        Raises:
+            InvalidExampleCollectionError: If the example output type doesn't match
+                the existing examples in the collection.
+        """
+        if not isinstance(example, self.example_class):
```
rohitrastogi: nit: might just be clearer to do the isinstance check against MapExample instead of example_class because this method is for the MapExampleCollection.

tests/_backends/local/semantic_operators/test_map.py
```diff
@@ -94,6 +105,79 @@ def test_build_prompts(self, local_session):
...
+            model=local_session._session_state.get_language_model(),
+            temperature=0,
+            max_tokens=512,
+            response_format=MovieReviewAnalysis
+        )
+        expected = [
+            [
+                {
+                    "content": str.strip(dedent("""\
+                    You are an AI assistant designed to follow instructions. Your task is to generate structured responses using the provided schema, based on instructions that reference one or more context fields.Each input message will have two sections:
```
rohitrastogi: fix spacing in prompt between "fields" and "Each input message"

src/fenic/core/_utils/structured_outputs.py
bcallender: yep, good catch

PR #80, opened 2025-07-14
chore: enable rust unit tests in ci using cargo test with feature flag

rohitrastogi:
# Fix PyO3 Linking Issues During Rust Unit Tests

This PR resolves linker errors that occur when running `cargo test` on the `polars-plugin` crate by disabling PyO3's `extension-module` feature during test builds.

## Problem

The `polars-plugin` crate uses PyO3 to build a Python extension module. By default, it enables the `extension-module` feature, which is necessary for tools like `maturin` to build shared libraries (`*.so`/`*.pyd`) that are importable from Python.

However, enabling `extension-module` tells PyO3:
* To assume it is running inside a Python process
* To rely on Python symbols (e.g. `_PyExc_SystemError`) being available at runtime

This works fine for extension builds, but causes issues when running `cargo test`, because:
* The test binary is not loaded by Python
* No Python interpreter is running
* Python symbols are undefined at link or runtime

Typical errors include:

```
undefined reference to _PyExc_SystemError
symbol not found: _PyList_New
```

## Solution

To fix this, we disable the `extension-module` feature during testing. This switches PyO3 to its default behavior where:
* Python is initialized explicitly by the Rust binary (if needed)
* Python symbols are linked at build time via `libpython`

This allows Rust tests to compile and run without requiring a running Python interpreter or `maturin`.

Reference: [PyO3 FAQ – Linker issues](https://pyo3.rs/v0.25.1/faq.html#i-cant-run-cargo-test-or-i-cant-build-in-a-cargo-workspace-im-having-linker-issues-like-symbol-not-found-or-undefined-reference-to-_pyexc_systemerror)

## Changes

1. **Cargo.toml** The crate already uses feature-based control of the extension module:

   ```toml
   [features]
   extension-module = ["pyo3/extension-module"]
   default = ["extension-module"]
   ```
   
   This lets us disable Python extension mode via `--no-default-features`.

2. **GitHub Actions** - Run Rust tests in CI now that we can
   * Updated CI to run `cargo test --no-default-features` to avoid linking to Python during tests
   * Sets `LD_LIBRARY_PATH` to locate `libpython` at build time
   * Limits test runs to Rust-related file changes to avoid running it needlessly
   
 3. Update CONTRIBUTING.md with instructions on running Rust only unit tests locally.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -86,7 +86,7 @@ jobs:
       - name: Run Rust unit tests
         if: steps.rust_changed.outputs.changed == 'true'
         working-directory: rust
-        run: cargo test --no-default-features --locked
+        run: cargo test --lib --no-default-features --locked
       # yamllint disable-line rule:line-length
       - name: run ${{ matrix.test }} tests with ${{ matrix.dependency == 'true' && 'synced' || matrix.dependency }} dependencies
         shell: bash
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -86,7 +86,9 @@ jobs:
       - name: Run Rust unit tests
         if: steps.rust_changed.outputs.changed == 'true'
         working-directory: rust
-        run: cargo test --lib --no-default-features --locked
+        run: |
+          export LD_LIBRARY_PATH=$(python -c "import sysconfig; print(sysconfig.get_config_var('LIBDIR'))"):$LD_LIBRARY_PATH
+          cargo test --no-default-features --locked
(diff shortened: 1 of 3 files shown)


bcallender (approved):
the changes overall make sense to me. @rohitrastogi have you seen if the results from the `cargo build` get cached before we move on to the next step, or do we end up redoing the compilation/downloading of dependencies in the fenic build step?

rohitrastogi:
They are compiled. They have to be recompiled due to differences in how the polars plugins are built (with extension-module using maturin, without extension-module for rust unit tests).

PR #87, opened 2025-07-16
feat: Add jinja template validation and variable extraction to be used in jinja rendering column function and more

rohitrastogi:
## Summary

This module provides validation logic and variable extraction for Jinja templates used in Fenic’s semantic operators. It ensures that templates adhere to strict rules to guarantee clean integration with Fenic expressions, safe rendering, and explicit variable dependency tracking.

## Template Rules

1. **Allowed Syntax:**  
Templates may only contain variable placeholders, attribute and fixed-index access, conditional logic (`if` statements), and loops over collections (`for` loops). Other features such as filters, function calls, macros, or assignments are prohibited and should be implemented using Fenic expressions.

2. **Variable Usage:**  
   - Top-level variables (those passed into the template) may be referenced anywhere in the template, including inside loops.   
   - The special variable `loop` (used in Jinja for loop metadata such as `loop.index`) is disallowed.

3. **Indexing Constraints:**  
   Index operations (e.g., `items[0]`, `order['id']`) must use constant integer or string indices. Dynamic indices based on variables or expressions are not allowed.

4. **Literals Restriction:**  
   Literal values (numbers, strings, booleans) cannot be used directly in expressions except as fixed indices in indexing operations. Users should pass literals through Fenic's literal mechanism.

## Behavior

- The validation process parses the template’s AST, verifying all nodes conform to the allowed syntax and rules.  
- It identifies loop variables and resolves their references to the underlying arrays they iterate over.  
- Variable usages are collected across three contexts:  
  - **Output:** variables rendered in the template output.  
  - **Condition:** variables used in `if` statement conditions.  
  - **Iteration:** variables used as loop iteration sources and loop variables.  
- A schema tree is built reflecting the expected data structure — marking variables as arrays, structs (objects), or booleans depending on usage patterns. This enables Fenic to understand what data shape to supply when rendering the template.  
- Validation errors provide clear, line-numbered messages for violations such as unsupported constructs, inconsistent variable typing, disallowed indexing, or forbidden use of the `loop` variable.

## Variable Extraction Details

- Variable paths are extracted recursively from AST nodes, supporting nested fields (e.g., `user.name`) and fixed indexing (e.g., `orders[0]`).  
- Loop variables are resolved to their original array sources, so that accesses like `order.id` inside `{% for order in orders %}` are mapped to `orders[*].id` in the schema tree.  
- Only variables that affect output or control flow for output variables are included in the final schema, filtering out unused variables.  
- This approach allows Fenic to enforce consistent variable types (e.g., a variable used as an array cannot also be used as a struct) and detect usage errors before runtime.


src/fenic/core/_logical_plan/utils.py
```diff
@@ -56,3 +71,98 @@ def validate_completion_parameters(
...
+            # Check if we're inside the corresponding For node
+            parent = node.parent
+            while parent and not isinstance(parent, nodes.For):
+                parent = getattr(parent, 'parent', None)
+            if not isinstance(parent, nodes.For) or parent.target.name == node.name:
+                return  # This usage is inside a valid loop
+
+            # If not inside its own loop body, it's invalid
+            raise ValidationError(
+                f"Jinja template error: Cannot use loop variable '{node.name}' in expressions (line {line_no}). Loop variables are only allowed in loop contexts."
```
bcallender: replace the repeated `Jinja template error:` with a dedicated exception type for errors related to the user's provided Jinja Template (`JinjaTemplateError` or `JinjaUnsupportedTemplateFeature`) that prepends it?

rohitrastogi: Don't think its necessary. It's  just user input validation like the jq query or anything else. I'm just adding the prefix to add a bit of context.

bcallender (approved):
lgtm!

PR #90, opened 2025-07-17
fix: make catalog use sc method to implement does_table_exist

german-typedef:
### TL;DR

Use a simple catalog way to determine if a table is available.

### What changed?

- Modified `CloudCatalog` initialization to accept direct parameters instead of depending on `CloudSessionState`
- Added a new `_get_table` method to improve table existence checking
- Updated `_does_table_exist` to use the new method instead of fetching all tables
- Updated tests to accommodate the new structure

PR #91, opened 2025-07-17
fix: use the cloud catalog when doing save_as_table in a cloud session

german-typedef:


german-typedef:
* **#91** <a href="https://app.graphite.com/github/pr/typedef-ai/fenic/91?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#94](https://github.com/typedef-ai/fenic/pull/94) <a href="https://app.graphite.com/github/pr/typedef-ai/fenic/94?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#147](https://github.com/typedef-ai/fenic/pull/147) <a href="https://app.graphite.com/github/pr/typedef-ai/fenic/147?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/typedef-ai/fenic/91?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#90** <a href="https://app.graphite.com/github/pr/typedef-ai/fenic/90?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/fenic/_backends/cloud/execution.py
```diff
@@ -164,16 +169,67 @@ def save_as_table(
...
+                file_format="PARQUET")
+            logger.debug(f"Table {table_identifier_str} created with result: {result}")
+
         # TODO (DY): check that current catalog and schema (if specified in table_name) match session state
-        table_identifier = TableIdentifier(
-            catalog=self.session_state.catalog,
-            schema=self.session_state.schema,
+        table_identifier_proto = TableIdentifierProto(
+            catalog=table_identifier.catalog,
+            schema=table_identifier.db,
```
bcallender, line 191: at some point we need to clean up some of these naming conventions -- we're calling this a `schema`, a `database` and a `namespace` in various different places.

src/fenic/_backends/cloud/execution.py
```diff
@@ -164,16 +169,67 @@ def save_as_table(
...
+                        f"Existing schema: {saved_schema} "
+                        f"Your DataFrame schema: {plan_schema} "
+                        f"To fix this: "
+                        f"1) Use mode='overwrite' to replace the table with your DataFrame's schema, "
+                        f"2) Modify your DataFrame to match the existing table's schema, "
+                        f"3) Use a different table name.")
+        else:
+            logger.debug(f"Creating table {table_identifier_str} with location: {logical_plan.location}")
+            # Create the table in the catalog.
+            result =self.session_state.catalog.create_table(
```
bcallender: could you add `**kwargs` to the base `create_table` method in `BaseCatalog` so the type validation works between the subclass with an extra parameter?

src/fenic/_backends/cloud/execution.py
```diff
@@ -164,16 +169,67 @@ def save_as_table(
     ) -> QueryMetrics:
         """Execute the logical plan and save the result as a table."""
         logger.debug(f"Saving plan {logical_plan} as table: {table_name}")
+
+        if isinstance(logical_plan, TableSink):
```
bcallender: if (somehow) `logical_plan` were not a `TableSink` we could just fail fast right? it might make sense to instead do something like 

```python
if not isinstance(logical_plan, TableSink):
     raise InternalError("Non TableSink LogicalPlan passed to `save_as_table`")

if not logical_plan.location: # field will definitely be present since we have now checked that logical_plan is a TableSink
                raise ValidationError(
                    f"Cannot save to table '{table_name}' - location is required. "
                    f"Provide a location.")
```

so that later when you call `logical_plan.location` it is guaranteed to be there.

bcallender (approved):
looks fine, just a few minor nits

src/fenic/_backends/local/physical_plan/sink.py
```diff
@@ -80,13 +80,15 @@ def __init__(
...
         schema: Schema,
+        location: Optional[str] = None,
     ):
         super().__init__(
             children=[child], cache_info=cache_info, session_state=session_state
         )
         self.table_name = table_name
         self.mode = mode
         self.schema = schema
+        self.location = location
```
rohitrastogi: We should have a separate physical plan node for Typedef managed tables. We shouldn't be adding the location metadata to the DuckDBTableSinkExec. The DuckDBTableSink and the managed typedef table sink have different implementations.

rohitrastogi: Actually, we shouldn't need to create  the PhysicalPlan object for the managed table sink here in Fenic at all - that node should only exist in the cloud service.

src/fenic/api/io/writer.py
```diff
@@ -46,6 +47,7 @@ def save_as_table(
                  - append: Appends data to table if it exists
                  - overwrite: Overwrites existing table
                  - ignore: Silently ignores operation if table exists
+            location: location where the table will be saved (only available for cloud execution)
```
rohitrastogi: I don't think we want save_as_table to accept a location. We should make the create_catalog API accept an S3 warehouse path (sort of like Spark). DBs in the catalog and tables in the dbs can be stored relative to the catalog path.

src/fenic/_backends/cloud/execution.py
german-typedef: Agreed, I think we started mixing some terms, but now that we have a library that's probably the way to go.

src/fenic/_backends/cloud/execution.py
german-typedef: Updated, I discussed this with Rohit, the approach we are taking is using the catalog warehouse location (s3 path as the basis) and then having the table location as the base of the key in that bucket.

src/fenic/_backends/cloud/execution.py
german-typedef: No longer need the table location to be part of the Sink node.

src/fenic/api/io/writer.py
german-typedef: `create_catalog` takes now the location (the S3 warehouse path) tables will be stored relative to the catalog path.

german-typedef: create_table does not need a location.

src/fenic/_backends/local/physical_plan/sink.py
german-typedef: We don't need it here, we already have a physical node in the Engine (separate PR), I'll update that one.

src/fenic/_backends/cloud/catalog.py
```diff
@@ -112,11 +113,18 @@ def list_catalogs(self) -> List[str]:
...
+    def create_catalog(
+            self,
+            catalog_name: str,
+            location: str,
+            ignore_if_exists: bool = True) -> bool:
         """Create a new catalog."""
         if compare_object_names(catalog_name, DEFAULT_CATALOG_NAME):
             raise CatalogError("Cannot create a catalog with the default name")
 
+        if urlparse(location).scheme not in CLOUD_SUPPORTED_SCHEMES:
```
rohitrastogi: Can we parse the scheme once and store it in a variable?

src/fenic/_backends/cloud/execution.py
```diff
@@ -164,16 +168,57 @@ def save_as_table(
     ) -> QueryMetrics:
         """Execute the logical plan and save the result as a table."""
         logger.debug(f"Saving plan {logical_plan} as table: {table_name}")
+
```
rohitrastogi: Isn't this validation logic common between local and catalog? If so, maybe we can push it into the API layer?

src/fenic/core/_logical_plan/plans/sink.py
```diff
@@ -91,6 +91,7 @@ def __init__(
                  - append: Appends data to table if it exists
                  - overwrite: Overwrites existing table
                  - ignore: Silently ignores operation if table exists
+            location: location where the table will be saved (only available for cloud execution)
```
rohitrastogi: This isn't required, right?

src/fenic/_backends/cloud/execution.py
```diff
@@ -164,16 +168,57 @@ def save_as_table(
...
+        table_identifier = TableIdentifier.from_string(table_name).enrich(
+            self.session_state.catalog.get_current_catalog(),
+            self.session_state.catalog.get_current_database(),
+        )
+
+        # If the table doesn't exist, create it, this has to be done in the user's context.
+        table_identifier_str = str(table_identifier)
```
rohitrastogi: Why do we need to call the str method? Don't we enrich/resolve table identifiers in the catalog methods?

src/fenic/core/_logical_plan/plans/sink.py
german-typedef: I've removed the comment.

src/fenic/_backends/cloud/execution.py
german-typedef: no need, remanent of the previous change.

src/fenic/_backends/cloud/execution.py
german-typedef: agreed, moved it out to a common way to validate.

src/fenic/_backends/cloud/catalog.py
german-typedef: Updated.

german-typedef:
My bad, this was out for a while. I've addressed all the comments.

src/fenic/_backends/cloud/catalog.py
bcallender: it seems like the location is still being parsed twice.

german-typedef: Good catch! forgot to push.
Done.

src/fenic/_backends/cloud/catalog.py
```diff
@@ -57,12 +58,12 @@
 
 logger = logging.getLogger(__name__)
 
-
 @dataclass(frozen=True)
 class CatalogKey:
     catalog_name: str
     catalog_id: UUID
 
+CLOUD_SUPPORTED_SCHEMES = ["s3"]
```
bcallender, line 66: we'll have to adjust this in #157

src/fenic/_backends/cloud/catalog.py
```diff
@@ -700,3 +700,8 @@ def _get_schema_type_to_pyarrow(schema_type: str):
             return pa.float64()
         else:
             return schema_type
+
+    @staticmethod
+    def _get_table_location_from_table_identifier(table_identifier: TableIdentifier) -> str:
```
bcallender, line 705: wouldn't the location be based on the base location in the catalog? if the user provides a bucket or prefix as the catalog's base location, all of the paths for dbs and tables should be relative to that base location, i would think.

src/fenic/_backends/cloud/execution.py
```diff
@@ -165,15 +168,33 @@ def save_as_table(
...
+        table_exists, query_metrics = self._validate_table_existance(logical_plan, table_name, mode)
+        if not table_exists:
+            raise CloudExecutionError(
+                f"Cannot save to table '{table_name}' - it does not exist. "
+                f"Choose a different approach: "
+                f"1) Create the table in question "
+                f"2) Use a different table name.")
+        elif table_exists and query_metrics:
+            # trunk-ignore-begin(bandit/B101)
+            assert mode == "ignore", "only mode to fulfill this invariant is ignore."
```
bcallender, line 180: particular reason to choose assert here and not raise an exception?

PR #93, opened 2025-07-17
refactor: remove need for window function in split part implementation with part number expr

rohitrastogi:
^


bcallender (approved):
as long as we're okay with removing the ability to specify a replacement_count this looks fine to me. it certainly simplifies the converter implementation.

rohitrastogi:
We were never accepting user input for the replacement_count (we were always hardcoding to -1), so this is not a breaking api change.

PR #96, opened 2025-07-21
feat: add jinja expr and validate jinja template against input exprs

rohitrastogi:


src/fenic/core/_logical_plan/expressions/text.py
```diff
@@ -845,3 +854,59 @@ def validator(self):
...
+    """Expression for evaluating a Jinja template.
+
+    This expression creates a new string column with the result of evaluating the Jinja template.
+
+    Args:
+        exprs: The input string column expressions
+        template: The Jinja template to evaluate
+    """
+
+    def __init__(self, exprs: List[Union[ColumnExpr, AliasExpr]], template: str):
```
bcallender, line 868: I thought about creating a `NamedExpr` Union for these two, since we are using this Union quite a bit, what do you think?

tests/_logical_plan/test_jinja_validation.py
```diff
@@ -0,0 +1,770 @@
...
+{%- endfor -%}
+{%- endif -%}
+{%- endfor -%}
+
+{%- if admin.logged_in -%}
+{%- for notification in admin.notifications -%}
+{%- endfor -%}
+{%- endif -%}
+{% if should_greet %}
+    {{ hello }}
```
bcallender, line 329: Does this test the newly fixed case of the dead code elimination where it would eliminate control variables that were attached to a section with a non-variable output?

bcallender (approved):

tests/_logical_plan/test_jinja_validation.py
rohitrastogi: yes, previously since should_greet and hello did not share common paths, should_greet would have been eliminated.

src/fenic/core/_logical_plan/expressions/text.py
rohitrastogi: will consider for a follow up

PR #97, opened 2025-07-21
feat: add rust utility to convert arrow array values into minijinja contexts

rohitrastogi:
This PR implements an efficient Rust-based converter for transforming Apache Arrow arrays into minijinja values, enabling high-performance template rendering in our query engine.

### Arrow to Jinja Value Conversion
- Added `arrow_scalar_extractor.rs` with a generic trait-based system for converting Arrow arrays to both JSON and minijinja values
- Supports comprehensive type coverage:
  - All Arrow primitive types (integers, floats, booleans, strings) supported in Fenic
  - Complex types (structs, lists, arrays)
  - Deeply nested structures (arrays of structs, structs containing arrays, etc.)

### Performance Optimizations
- Replaced slower `anyvalue_to_json` approach with direct Arrow-to-value conversion. `anyvalue` required unnecessary allocations and also and use this the arrow scalar extractor in tojson casting.

### Infrastructure
- Added `minijinja` dependency to `Cargo.toml`
- Implemented comprehensive test coverage for all supported type conversions
- Structured code with clear trait boundaries for extensibility


german-typedef (approved):

rust/src/arrow_scalar_extractor.rs
```diff
@@ -0,0 +1,1004 @@
...
+    fn convert<V: ArrowToValue>(&self, array: &dyn Array, row_idx: usize) -> PolarsResult<V> {
+        // Check for null only for primitive types
+        if Self::is_primitive_type(array.dtype()) && array.is_null(row_idx) {
+            return Ok(V::from_null());
+        }
+
+        match array.dtype() {
+            ArrowDataType::Utf8 => {
+                let str_array =
+                    array
```
bcallender: it felt like this repetition could be consolidated so I asked claude -- could a macro like this work to handle these cases?

```rust
  macro_rules! downcast_array {
      ($array:expr, $type:ty, $name:literal) => {
          $array.as_any().downcast_ref::<$type>()
              .ok_or_else(|| PolarsError::ComputeError(
                  format!("Failed to downcast to {}", $name).into()
              ))?
      };
  }
```

rust/src/arrow_scalar_extractor.rs
```diff
@@ -0,0 +1,1004 @@
...
+                | ArrowDataType::UInt64
+                | ArrowDataType::Float32
+                | ArrowDataType::Float64
+        )
+    }
+
+    // Single generic implementation that works for both JinjaValue and JsonValue
+    fn convert<V: ArrowToValue>(&self, array: &dyn Array, row_idx: usize) -> PolarsResult<V> {
+        // Check for null only for primitive types
+        if Self::is_primitive_type(array.dtype()) && array.is_null(row_idx) {
```
bcallender, line 222: just curious, why do structs get their internal values set to null/empty instead of just being marked entirely as null? it also seems like the JSON implementation diverges when it is handling the null fields, is that because of different ways we handle the resultant row in the `JinjaExpr`?

rust/src/arrow_scalar_extractor.rs
rohitrastogi: done

rust/src/arrow_scalar_extractor.rs
rohitrastogi: We have to treat null structs differently based on the output context:

For JSON output:
We return a true null struct (i.e., null), preserving the null semantics (this is also the existing behavior with Anyvalue. I noticed that when we cast from Json to Struct, we cast null Json as non-null structs with null fields. This is a problem that should be fixed separately so we get consistent roundtrip behavior.

For Jinja rendering:
Returning null would cause runtime errors because Jinja expects the struct’s fields to exist in the context.
Instead, we return a map/object with the same fields as the struct, but with each field set to an empty string ("").
This ensures Jinja templates can safely access all fields without panicking at runtime.

PR #98, opened 2025-07-21
feat: add implementation for text.jinja() column function

rohitrastogi:
This PR completes the Jinja template implementation by adding the Rust rendering engine and exposing it through a Python API, enabling efficient template-based string generation in Fenic.

### Rust Implementation
- **`jinja/mod.rs`**: Polars plugin entry point that registers the Jinja expression
- **`jinja/render.rs`**: Core rendering engine that:
  - Leverages the Arrow-to-Jinja converter from PR 2
  - Handles template compilation
  - Processes arrays row-by-row with minimal allocations
  - Includes extensive test coverage for real-world scenarios

### Python API
- **`text.jinja()` function**: Clean kwargs-based API for template rendering
  ```python
  df.select(text.jinja(
      "Hello {{ name }}{% if vip %} (VIP){% endif %}!",
      name=col("customer_name"),
      vip=col("is_vip")
  ))
  ```
- **Polars plugin integration**: Seamless integration with Polars' expression system
- **Transpiler support**: Added `JinjaExpr` handling for query plan serialization


rust/src/jinja/render.rs
```diff
@@ -0,0 +1,524 @@
...
+    // Process each row
+    for row_idx in 0..struct_series.len() {
+        // Build context for this row
+        let mut ctx = BTreeMap::new();
+
+        // Extract values for all fields
+        for field_name in &field_names {
+            match struct_array.field_by_name(field_name) {
+                Ok(field_series) => {
+                    // Handle multiple chunks if necessary
```
bcallender, line 41: when would this be the case? when the field value is itself a series that is larger than the chunk size?

rust/src/jinja/render.rs
```diff
@@ -0,0 +1,524 @@
...
+use polars::prelude::*;
+use std::collections::BTreeMap;
+
+use crate::arrow_scalar_extractor::ArrowScalarConverter;
+
+/// Core implementation of Jinja template rendering
+pub fn render(inputs: &[Series], template: &str) -> PolarsResult<Series> {
+    // Create minijinja environment and compile template
+    let env = Environment::new();
+    let template = env.template_from_str(template).map_err(|e| {
```
bcallender, line 11: this is run for each row right? would it be possible to cache the "compiled" template to avoid needing to recompute this each row?

src/fenic/api/functions/text.py
```diff
@@ -925,3 +928,103 @@ def byte_length(column: ColumnOrName) -> Column:
...
+        - If/elif/else conditionals: {% if condition %}...{% endif %}
+        - Loop variables: {{ loop.index }}, {{ loop.first }}, etc.
+        - Constants: {{ "literal string" }}, {{ 42 }}
+
+    Not Supported (use column expressions instead):
+        - **Filters**: {{ name|upper }} → Use upper_name=fc.upper(col("name"))
+        - **Function calls**: {{ len(items) }} → Use item_count=fc.array_size(col("items"))
+        - **Operators**: {% if price > 100 %} → Use is_expensive=(col("price") > 100)
+        - **Arithmetic**: {{ price * quantity }} → Use total=col("price") * col("quantity")
+        - **Dynamic indexing**: {{ items[i] }} → Use item=(fc.col("items").get_item(col("index)))
```
bcallender: nit: missing closing `"`

src/fenic/api/functions/text.py
```diff
@@ -925,3 +928,103 @@ def byte_length(column: ColumnOrName) -> Column:
...
+        prompt_template = '''
+        Answer the user's question.
+
+        {% if context %}
+        Context: {{ context }}
+        {% endif %}
+
+        {% if examples %}
+
+    Examples:
```
bcallender: mixed usage of `Example` and `Examples` will lead to inconsistent results in the documentation -- only `Example` will result in the nice admonition box, `Examples` will render it in with the rest of the description.

src/fenic/api/functions/text.py
```diff
@@ -925,3 +928,103 @@ def byte_length(column: ColumnOrName) -> Column:
...
+        # Generate prompts with varying context based on query type
+        result = df.select(
+            text.jinja(
+                prompt_template,
+                # Direct columns
+                query=col("user_question"),
+                context=col("retrieved_context"),  # Can be null for some rows
+
+                # Column expression for conditional logic
+                style=F.when(col("query_type") == "technical", "detailed and technical")
```
bcallender: `fc`?

bcallender (approved):

rust/src/jinja/render.rs
rohitrastogi: It's not computed for each row. It's computed once and then reused for each row.

rust/src/jinja/render.rs
rohitrastogi: This is just how Polars works. Polars Series are themselves list of chunked arrays which are themselves lists of contiguous Arrow arrays. Polars doesn't force a series to be contiguous in memory. Of course, you can do this if you want using their rechunk api. Give this a read for more info: https://docs.rs/polars/latest/polars/chunked_array/struct.ChunkedArray.html

src/fenic/api/functions/text.py
rohitrastogi: nice catch, done.

src/fenic/api/functions/text.py
rohitrastogi: something weird happened when examples was part of the template. fixed.

src/fenic/api/functions/text.py
rohitrastogi: done.

PR #103, opened 2025-07-23
feat: add support for basic fuzzy string ratio column functions

rohitrastogi:
## Description

This PR adds fuzzy string matching capabilities to the text module, providing 6 similarity algorithms across 3 comparison modes to serve as primitives for text matching, deduplication, and record linkage workflows.

We implement this functionality as Polars Plugins by wrapping the [rapidfuzz-rs crate](https://docs.rs/rapidfuzz/0.5.0/rapidfuzz/). The `compute_fuzzy_ratio` function uses the plugin directly, while `compute_fuzzy_token_sort_ratio` and `compute_fuzzy_token_set_ratio` are composed using Polars expressions on top of the plugin.

### What's New

**Three fuzzy matching functions:**
- `text.compute_fuzzy_ratio()` - Direct string similarity comparison
- `text.compute_fuzzy_token_sort_ratio()` - Order-independent comparison after token sorting
- `text.compute_fuzzy_token_set_ratio()` - Set-based comparison ignoring duplicates and order

**Six similarity algorithms:**
- **Indel**: Pure insertion-deletion distance (no substitutions)
- **Levenshtein**: Classic edit distance
- **Damerau-Levenshtein**: Edit distance with transpositions
- **Jaro**: Character proximity-based similarity
- **Jaro-Winkler**: Jaro with prefix boost
- **Hamming**: Position-by-position comparison with auto-padding

### Usage Examples

```python
# Direct similarity (default: indel)
text.compute_fuzzy_ratio(col("name1"), col("name2"), method="jaro_winkler")
text.compute_fuzzy_ratio(col("product"), "iPhone 15", method="levenshtein")

# Order-independent comparison
text.compute_fuzzy_token_sort_ratio(col("address1"), col("address2"))
# "new york city" vs "city new york" → 100%

# Set-based comparison
text.compute_fuzzy_token_set_ratio(col("description1"), col("description2"))
# "fuzzy fuzzy was a bear" vs "fuzzy was a bear but not a dog" → 100%
```

### Limitations
We don't currently support `partial_*` functions or [WRatio](https://rapidfuzz.github.io/RapidFuzz/Usage/fuzz.html#rapidfuzz.fuzz.WRatio), which depend on [partial_ratio](https://rapidfuzz.github.io/RapidFuzz/Usage/fuzz.html#rapidfuzz.fuzz.partial_ratio). The Rust crate doesn't support these, and they're not easily expressible as Polars expressions. Implementation would require writing a fast partial_ratio kernel in Rust.

BEGIN_COMMIT_OVERRIDE
feat: add support for basic fuzzy string ratio column functions
END_COMMIT_OVERRIDE

### DIFF
--- a/src/fenic/api/functions/text.py
+++ b/src/fenic/api/functions/text.py
@@ -1045,10 +1045,13 @@ def compute_fuzzy_ratio(column: ColumnOrName, other: Union[Column, str], method:
         - 100 indicates the strings are identical
         - 0 indicates maximum dissimilarity (as defined by the method)
 
+    Based on https://rapidfuzz.github.io/RapidFuzz/Usage/fuzz.html#rapidfuzz.fuzz.ratio
+
     Args:
         column: A string column or column name. This is the left-hand side of the comparison.
         other: A second string column or literal string. This is the right-hand side of the comparison.
         method: A string indicating which similarity method to use. Must be one of:
+            - `"indel"`: Indel distance — counts only insertions and deletions (no substitutions); based on the Longest Common Subsequence.
             - `"levenshtein"`: Levenshtein distance (edit distance)
             - `"damerau_levenshtein"`: Damerau-Levenshtein distance (includes transpositions)
             - `"jaro"`: Jaro similarity, accounts for transpositions and proximity
@@ -1087,6 +1090,8 @@ def compute_fuzzy_token_sort_ratio(column: ColumnOrName, other: Union[Column, st
     them back into a string, then applies the specified similarity metric.
     Useful for comparing strings where word order doesn't matter.
 
+    Based on https://rapidfuzz.github.io/RapidFuzz/Usage/fuzz.html#rapidfuzz.fuzz.token_sort_ratio
+
     Args:
         column: First string column to compare
         other: Second string column or literal string to compare against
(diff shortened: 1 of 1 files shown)


src/fenic/api/functions/text.py
```diff
@@ -1027,3 +1032,115 @@ def jinja(
...
+    and a literal string) for each row. It supports multiple well-known string similarity metrics,
+    including Levenshtein, Damerau-Levenshtein, Jaro, Jaro-Winkler, and Hamming.
+
+    The returned score is a similarity percentage between 0 and 100, where:
+        - 100 indicates the strings are identical
+        - 0 indicates maximum dissimilarity (as defined by the method)
+
+    Args:
+        column: A string column or column name. This is the left-hand side of the comparison.
+        other: A second string column, column name, or literal string. This is the right-hand side of the comparison.
```
german-typedef: nit: is other really a string column or column name? from the definition it seems that is just either a Column or a string.

rohitrastogi: Yeah, it can't be a column name. Good catch, updating.

src/fenic/api/functions/text.py
```diff
@@ -1027,3 +1032,115 @@ def jinja(
...
+        column: First string column to compare
+        other: Second string column or literal string to compare against
+        method: Similarity algorithm to use after token sorting
+
+    Returns:
+        Column with similarity scores between 0 and 100
+
+    Example:
+        ```python
+        # df.select(compute_fuzzy_token_sort_ratio("name", "Smith John", "levenshtein"))
```
german-typedef: nit: is the example correct?

german-typedef:
@rohitrastogi Just a couple of nits, one of them you already fixed. Could you check the code example for `compute_fuzzy_token_sort_ratio` ?

src/fenic/api/functions/text.py
rohitrastogi: no, fixed.

rohitrastogi:
> @rohitrastogi Just a couple of nits, one of them you already fixed. Could you check the code example for `compute_fuzzy_token_sort_ratio` ?

fixed.

PR #104, opened 2025-07-24
feat: replace pylance kmeans implementation with scikit-learn and expose num_init and max_iter params in api

rohitrastogi:
Replace Lance's KMeans implementation with scikit-learn and expose `n_init` and `max_iter` parameters in the fenic API.

## Background

We initially used Lance's "hidden" KMeans implementation because it was relatively lightweight and easy to install. In contrast, faiss-cpu has installation issues on Apple Silicon Macs and generally requires Anaconda, which would negatively impact the developer and installation experience of fenic. Lance also had a smaller dependency footprint (though Lance isn't small either due to the package including support for the entire Lance file format) than scikit-learn, making it an appealing choice early on (approximately 30 MB smaller, or about ~35% the size of Lance itself). It also has first class support for Arrow FixedSizedLists.

However, after using it in real-world clustering scenarios, we found that its silhouette score performance was consistently worse than scikit-learn. This is likely because scikit-learn supports k-means++ centroid initialization and exposes the n_init parameter, which allows running KMeans multiple times with different initializations and selecting the best result.

While the underlying Rust implementation used by Lance does support a parameter [similar to n_init](https://github.com/lancedb/lance/blob/main/rust/lance-index/src/vector/kmeans.rs#L544), it is not exposed in the [Python bindings](https://github.com/lancedb/lance/blob/bed04d3f77ced32a0212ade733835cb67ac09463/python/src/utils.rs#L82). It also lacks a random seed parameter, meaning repeated clustering runs produce highly variable results. Because there is no way to stabilize or repeatably improve clustering quality (as n_init allows), some runs yield significantly worse silhouette scores, making it unideal for fenic.

## Changes

This PR:
1. Replaces Lance's KMeans implementation with scikit-learn's KMeans
2. Exposes two important parameters in the fenic API:
   - `n_init`: Number of times KMeans will run with different centroid seeds (default: 1). This param allows the algorithm to try multiple initializations and select the best result based on inertia.
   - `max_iter`: Maximum number of iterations for a single run (default: 300). Controls convergence behavior and computational cost.

These parameters give users control over the clustering quality vs performance trade-off and enable more stable, reproducible results.

## Future Considerations

Some better long-term solutions include:
- Building a lightweight custom KMeans implementation that works directly over PyArrow FixedSizeLists, without the full scikit-learn dependency set.
- Creating PyO3 bindings for [linfa-clustering](https://crates.io/crates/linfa-clustering), though linfa-clustering doesn't support PyArrow FixedSizeList.
- Contributing to Lance to expose n_iter and implement k-means++ initialization, or even vendoring just the Rust clustering logic and building new PyO3 bindings if feasible.


rohitrastogi:
> The code change lgtm, but I do have a question -
> 
> Do we want n_init to be configurable? Do we want the user to be able to choose performance/lower query latency over better clustering quality?

Good call, made num_init and max_iter configurable with sensible defaults.

PR #105, opened 2025-07-24
feat: Add support for webvtt transcript format

### LINKED ISSUE #71: Add support for VTT format transcripts in parse_transcript() column function
## Add Support for WebVTT Transcript Format

Currently, `parse_transcript()` supports:
- SRT transcript files
- A loosely standardized "generic" transcript format emitted by some podcasting platforms

To improve ingestion from a wider variety of sources, it would be useful for **Fenic** to support additional formats like [**WebVTT**](https://developer.mozilla.org/en-US/docs/Web/API/WebVTT_API/Web_Video_Text_Tracks_Format).

The `parse_transcript()` function is implemented as a [**Polars plugin**](https://marcogorelli.github.io/polars-plugins-tutorial/vec_of_option/) in Rust. The existing [parser implementations](https://github.com/typedef-ai/fenic/blob/616f54104d897f0968939223071065b328e80495/rust/src/transcript/parsers/srt.rs#L52) uses [`memchr`](https://docs.rs/memchr/latest/memchr/) for efficient parsing.

---

### Proposed Changes

1. **Extend the transcript parser** to support **WebVTT** in Rust, emitting `UnifiedTranscriptEntry` structs.
   - One possible approach: convert VTT to SRT internally and reuse existing SRT parsing logic.

2. **Update the [TranscriptFormatType](https://github.com/typedef-ai/fenic/blob/616f54104d897f0968939223071065b328e80495/src/fenic/core/types/enums.py#L31) enum** in the query frontend used in the [user facing API layer ](https://github.com/typedef-ai/fenic/blob/616f54104d897f0968939223071065b328e80495/src/fenic/api/functions/text.py#L419)to include a new `Vtt` variant.

3. **Plumb the format information** through to the execution layer:
   - Extend [TsParseExpr](https://github.com/typedef-ai/fenic/blob/616f54104d897f0968939223071065b328e80495/src/fenic/core/_logical_plan/expressions/text.py#L512) to handle the new format
   - Pass the format type into the Rust backend via the [expr_converter](https://github.com/typedef-ai/fenic/blob/616f54104d897f0968939223071065b328e80495/src/fenic/_backends/local/transpiler/expr_converter.py#L375)

---

This change would improve support for ingesting transcripts from platforms that natively emit VTT, such as YouTube and web-based video players.

YoungVor:
feat: Add basic support for WebVTT format to transcript parser

### rust WebVTT parser 
* expects specific timestamp format and WebVTT header
* skips optional cue settings following timestamp range
* returns raw content in cue entries, stripping all html or style tags through to the output
* Auto-indexes cue entries, ignoring identifiers as they can be arbitrary characters

### WebVTT version of transcripts for benchmarks
* webvtt parser is slower than generic and srt, transcript-1 running at 80us vs 30us

## Not in scope, potential improvements
* Translating html format tags like newline or space 
* Opimize by replacing HTML and speaker tag regexes  

Closes #71


rust/src/transcript/parsers/webvtt.rs
```diff
@@ -0,0 +1,528 @@
...
+    fn test_webvtt_with_bom_and_without_identifiers() {
+        // concat the BOM character
+        let webvtt_content = format!(
+            "\u{FEFF}{}",
+            r#"WEBVTT
+00:00:01.000 --> 00:00:05.000
+No identifier here.
+
+00:00:06.000 --> 00:00:10.000
+<v Bob>Bob speaking without identifier.
```
rohitrastogi: nit: Bob speaking with identifier.

rust/src/transcript/parsers/webvtt.rs
```diff
@@ -0,0 +1,528 @@
...
+
+    #[test]
+    fn test_various_cue_identifiers() {
+        let webvtt_data = r#"WEBVTT
+
+1
+00:00:01.000 --> 00:00:03.000
+Simple numeric identifier
+
+cue-2
```
rohitrastogi, line 372: I think here (and for SRT though I think SRT requires indices) we should just auto increment the indices instead of parsing them as is, wdyt?

YoungVor: Yeah, that's probably simplest.  We could also provide an identifier string to the UnifiedEntry struct

rust/src/transcript/parsers/webvtt.rs
YoungVor: i meant cue identifier, rather than speaker tag.  WIll update to be clearer

rust/src/transcript/parsers/webvtt.rs
YoungVor: I've plumbed identifier separately and auto-generated the index for webvtt

YoungVor:
@rohitrastogi Removed regexes for timestamp parsing.  As a future optimization, we can replace the regex for tag parsing as well

tests/_backends/local/functions/test_transcripts.py
```diff
@@ -37,6 +37,64 @@ def test_parse_transcript_srt_format(local_session):
...
+    entry1 = entries[0]
+    assert entry1["index"] == 1
+    assert entry1["speaker"] is None  # WebVTT doesn't currently parse the speakers
+    assert entry1["start_time"] == 1.0  # 00:00:01,000 = 1 second
+    assert entry1["end_time"] == 4.0   # 00:00:04,000 = 4 seconds
+    assert entry1["duration"] == 3.0   # 4 - 1 = 3 seconds
+    assert entry1["content"] == "Hello, world!"
+    assert entry1["format"] == "webvtt"
+
+    # Check second entry
```
rohitrastogi, line 81: we should have a speaker for element 2, right? can we assert for it?

YoungVor: whoops!  Forgot to update the python unit tests after adding the speaker.  Done

PR #106, opened 2025-07-25
fix: split plan creation from validation

german-typedef:
### TL;DR

Refactored the logical plan architecture to separate the logical plan node tree from the session state, improving serialization and plan manipulation.

### What changed?

- Created a new `LogicalPlanNode` class hierarchy that's independent of session state
- Modified `LogicalPlan` to be a container that holds both a `LogicalPlanNode` tree and session state
- Updated all plan operations to work with the new node-based architecture
- Refactored expression evaluation to take an optional session state parameter
- Modified plan serialization to only serialize the node tree, not the session state

### DIFF
--- a/src/fenic/core/_logical_plan/expressions/base.py
+++ b/src/fenic/core/_logical_plan/expressions/base.py
@@ -4,13 +4,14 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from enum import Enum
-from typing import TYPE_CHECKING, List
+from typing import TYPE_CHECKING, List, Optional
 
 if TYPE_CHECKING:
-    from fenic.core._logical_plan import LogicalPlan
+    from fenic.core._logical_plan.plans.node import LogicalPlanNode
     from fenic.core._logical_plan.signatures import SignatureValidator
     from fenic.core.types.datatypes import DataType
 
+from fenic.core._interfaces.session_state import BaseSessionState
 from fenic.core.types import ColumnField
 
 
@@ -40,8 +41,8 @@ class LogicalExpr(ABC):
         pass
 
     @abstractmethod
-    def to_column_field(self, plan: LogicalPlan) -> ColumnField:
-        """Returns the schema field for the expression within the given plan."""
+    def to_column_field(self, node: LogicalPlanNode, session_state: Optional[BaseSessionState] = None) -> ColumnField:
+        """Returns the schema field for the expression within the given plan node."""
         pass
 
     @abstractmethod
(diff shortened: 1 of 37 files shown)


german-typedef:
* **#106** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/106?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/106?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/fenic/_backends/local/execution.py
```diff
@@ -60,6 +60,7 @@ def show(self, plan: LogicalPlan, n: int = 10) -> Tuple[str, QueryMetrics]:
         """Execute a logical plan and return a string representation of the sample rows of the DataFrame and query metrics."""
         self.session_state._check_active()
         physical_plan = self.transpiler.transpile(plan)
+        print(f"Physical plan: {physical_plan}")
```
YoungVor: Do we always want to print the physical plan on a show()?  Or make it optional?

YoungVor:
>> Refactored expression evaluation to take an optional session state parameter

Hey @german-typedef , can you walk me through why this is necessary?  I guess, some to_column evaluations and schema validation depend on the model you are using, is that why there's a dependency on session state?

When session state was part of the plan, we didn't need to pass it to the eval.  But I'm also confused about why its optional to pass the session_state when doing the to_column_expression

german-typedef:
@YoungVor Correct, there are a few places where the to_column require the session state. I left it as Optional just in case, but I think we can make it not optional. Let's get together and do a quick walk through.

src/fenic/core/_logical_plan/expressions/base.py
```diff
@@ -128,22 +129,27 @@ def children(self) -> List[LogicalExpr]:
         pass
 
     @abstractmethod
-    def _infer_dynamic_return_type(self, arg_types: List[DataType], plan: LogicalPlan) -> DataType:
+    def _infer_dynamic_return_type(
+        self,
+        arg_types: List[DataType],
+        node: LogicalPlanNode,
+        session_state: BaseSessionState) -> DataType:
```
YoungVor, line 136: shouldn't session_state by optional for all the _infer_dynamic_return_type definitions, since we get session_state from to_column_field (optional) and validator will pass that session_state variable?

src/fenic/core/_logical_plan/signatures/function_signature.py
```diff
@@ -49,12 +50,13 @@ def __init__(
     def validate_and_infer_type(
         self,
         args: List[LogicalExpr],
-        plan: LogicalPlan,
-        dynamic_return_type_func: Optional[Callable[[List[DataType], LogicalPlan], DataType]] = None
+        node: LogicalPlanNode,
+        session_state: BaseSessionState,
```
YoungVor, line 54: Shouldn't this be optional, since we call from to_column_field which hints session_state is Optional[session_state]?

src/fenic/_backends/local/execution.py
german-typedef: My bad, removing it. This shouldn't be there. I just did it for testing.

PR #112, opened 2025-07-28
fix: Check UDF return value and validate its not a logical type

YoungVor:
# Summary

Imporve UDF type checking - check UDF return type before mapping the result to the polars series/column output, and validate that the UDF is not returning a logical type since that's currently unsupported.

# What Changed

* UDF polars wrapper also checks the return type to see if we can cast it to one of our DTypes
* udf() will validate that the return type is not a logical type
* Moved can_cast and added is_logical_type to utils type_inference
* Added unit tests for logical type validation and for checking a mismatching return type

# Out of scope
Add support for returned LogicalTypes in UDFs (see [issue #114](https://github.com/typedef-ai/fenic/issues/114))

### DIFF
--- a/src/fenic/_backends/local/transpiler/expr_converter.py
+++ b/src/fenic/_backends/local/transpiler/expr_converter.py
@@ -105,7 +105,8 @@ from fenic.core._utils.schema import (
     convert_custom_dtype_to_polars,
     convert_pydantic_type_to_custom_struct_type,
 )
-from fenic.core.error import InternalError
+from fenic.core._utils.type_inference import can_cast, infer_dtype_from_pyobj
+from fenic.core.error import ExecutionError, InternalError
 from fenic.core.types.datatypes import (
     ArrayType,
     BooleanType,
@@ -339,8 +340,9 @@ class ExprConverter:
         struct = pl.struct(
             [self._convert_expr(arg) for arg in logical.args]
         )
-        converted_udf = _convert_udf_to_map_elements(
-            logical.func
+        converted_udf = _convert_udf_to_map_elements_and_validate_return_type(
+            logical.func,
+            logical.return_type,
         )
         return struct.map_elements(
             converted_udf,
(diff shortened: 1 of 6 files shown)


src/fenic/_backends/local/transpiler/expr_converter.py
```diff
@@ -1171,11 +1173,20 @@ def _convert_udf_to_map_elements(udf: Callable) -> Callable:
...
         A function that operates on a row (Struct) and applies the UDF to the unwrapped values.
+    
+    Raises:
+        ExecutionError: If the type of the UDF's return value does not match the expected return type.
     """
 
     def adapted_udf(row):
+        def udf_with_validate_return_type(udf: Callable, return_type: DataType, *args, **kwargs) -> DataType:
+            result = udf(*args, **kwargs)
+            dtype = infer_dtype_from_pyobj(result)
```
rohitrastogi, line 1184: this changed isn't required. polars already handles the coercion and will try and coerce if possible. we don't need to add our own can cast check.

src/fenic/core/_utils/type_inference.py
```diff
@@ -96,3 +104,63 @@ def _find_common_supertype(type1: DataType, type2: DataType, path="") -> DataTyp
         return StructType(merged_fields)
 
     raise TypeInferenceError(f"Incompatible types: {type1} vs {type2}", path)
+
+def is_logical_type(type: DataType) -> bool:
```
rohitrastogi, line 108: It may be cleaner to just add a marker type for logical types (like we do for _StringBackedType) so we can just do isinstance of logical type.

We can actually just get rid of stringbackedtype and change it to logicaltype and then also have embeddingtype inherit from it.

src/fenic/_backends/local/transpiler/expr_converter.py
rohitrastogi: I think for this PR, all we need is the change in the api layer that validates the user provided return type.

YoungVor: got it, will remove the wrapper function

src/fenic/core/_utils/type_inference.py
YoungVor: Makes sense!

YoungVor: Will keep is_logical_type though, so we can determine if an Array or Struct contains a LogicalType

rohitrastogi: yeah, makes sense!

YoungVor:
moved to https://github.com/typedef-ai/fenic/pull/117

PR #116, opened 2025-07-29
Add support for Cohere embeddings

YoungVor:


YoungVor:
* **#116** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/116?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/116?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#122** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/122?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

examples/test_cohere.py
rohitrastogi, line 1: not sure we want to commit this.

examples/transcripts.py
rohitrastogi, line 1: also, not sure if we should be checking this in

tests/conftest.py
```diff
@@ -174,6 +175,63 @@ def examples_session_config(app_name) -> SessionConfig:
         ),
     )
 
+@pytest.fixture
+def examples_cohere_session_config(app_name) -> SessionConfig:
```
bcallender: Why don't I merge the fix I have for the examples config into main -- then you can just add this to the main `local_session_config` fixture when cohere is passed as the model provider

src/fenic/core/_inference/model_catalog.py
```diff
@@ -771,6 +782,69 @@ def _initialize_google_gla_models(self):
...
+
+        # embed-english-v3.0 - English-only model
+        self._add_model_to_catalog(
+            ModelProvider.COHERE,
+            "embed-english-v3.0",
+            EmbeddingModelParameters(
+                input_token_cost=0.10 / 1_000_000,  # $0.10 per 1M tokens
+                allowed_output_dimensions=1024,  # Fixed dimensions
+                max_input_size=512,
+                default_dimensionality=1024,
```
bcallender: you don't need to specify the default if only one dimension is allowed.

tests/examples/test_examples.py
```diff
@@ -25,8 +25,15 @@ def get_example_scripts():
...
 @pytest.mark.parametrize("script_path", get_example_scripts())
-def test_example_script(script_path, examples_session_config):
+def test_example_script_google(script_path, examples_google_session_config):
     """Test that each example script's main function runs without errors."""
     module = import_module_from_path(script_path)
     assert hasattr(module, "main"), f"Script {script_path} does not have a main() function"
-    module.main(examples_session_config)  # Run the main function
+    module.main(examples_google_session_config)  # Run the main function
+
+@pytest.mark.parametrize("script_path", get_example_scripts())
```
bcallender: yeah, lets get #118 in (or you can restack onto it) -- then we can pass arguments to pytest to configure the model provider.

src/fenic/api/session/config.py
```diff
@@ -592,7 +595,84 @@ class Profile(BaseModel):
             ge=1024,
         )
 
-EmbeddingModel = Union[OpenAIEmbeddingModel, GoogleVertexEmbeddingModel, GoogleDeveloperEmbeddingModel]
+CohereEmbeddingTaskType = Literal[
+    "search_document",
+    "search_query",
+    "classification",
+    "clustering",
+    "image", # note: used by embed v3.0.  v4.0 will default to search_document
```
bcallender: can we even support `image` if we are passing `texts` in the embeddings client?

src/fenic/core/_inference/model_catalog.py
```diff
@@ -771,6 +782,69 @@ def _initialize_google_gla_models(self):
             ),
         )
 
+    # TODO: lookup actual pricing from Cohere
```
bcallender: still TODO?

tests/conftest.py
YoungVor: sounds good!

src/fenic/core/_inference/model_catalog.py
YoungVor: Yes, I found the latest pricing on cohere's embed4 on their site, but had trouble finding their pricing for embed3.  I do see embed3 multi-lingual pricing on platform providers like AWS and Azure, so I think we can use that

tests/examples/test_examples.py
YoungVor: sounds good!

src/fenic/core/_inference/model_catalog.py
YoungVor: double checked the pricing with LiteLLM too

src/fenic/api/session/config.py
YoungVor: i'm not sure what will happen if you pass text in the API. You're right, better not to give the option

tests/examples/test_examples.py
YoungVor: discussed on a separate thread, I created a separate PR for changing the pytest arguments

src/fenic/core/_inference/model_catalog.py
YoungVor: gotcha.  Thanks!

src/fenic/_inference/cohere/cohere_batch_embeddings_client.py
```diff
@@ -0,0 +1,197 @@
...
+            model=model,
+            model_provider=ModelProvider.COHERE,
+            rate_limit_strategy=rate_limit_strategy,
+            queue_size=queue_size,
+            max_backoffs=max_backoffs,
+            token_counter=TiktokenTokenCounter(model_name=model),
+        )
+        
+        # Initialize Cohere client - reads COHERE_API_KEY from environment
+        api_key = os.environ.get("COHERE_API_KEY")
```
bcallender, line 58: nit: The pattern we have in each of the other clients (except a special case for google backwards compatibility) is not to load the API key from the env, but instead just create the client like `self._client = cohere.ClientV2()`. If the env var is set, with the correct name, the library should come up. Otherwise, the library itself will throw an error when we instantiate it.

src/fenic/_inference/cohere/cohere_batch_embeddings_client.py
```diff
@@ -0,0 +1,197 @@
...
+                "embedding_types": ["float"], # We only support float embeddings
+            }
+            
+            # Add output dimensionality if specified
+            if profile_config.output_dimensionality:
+                embed_params["output_dimension"] = profile_config.output_dimensionality
+            
+            # Make the API call
+            response = self._client.embed(**embed_params)
+            if not hasattr(response, 'embeddings') or not hasattr(response.embeddings, 'float') or not len(response.embeddings.float) > 0:
```
bcallender: why all of the `hasattr` here? this `EmbedByTypeResponse` class that the function is return ing looks to be well-typed, with only some optional fields, which `hasattr` won't help with.

src/fenic/_inference/cohere/cohere_batch_embeddings_client.py
```diff
@@ -0,0 +1,197 @@
...
+        except cohere.InternalServerError as e:
+            # Server error - retryable
+            return TransientException(e)
+        except cohere.ServiceUnavailableError as e:
+            # Service unavailable - retryable
+            return TransientException(e)
+        except (cohere.BadRequestError, cohere.UnauthorizedError, cohere.ForbiddenError) as e:
+            # Client errors - not retryable
+            return FatalException(e)
+        except Exception as e:
```
bcallender, line 137: make sure to look into which http library the cohere client uses. For the Gemini genai client, the `httpx` exceptions that occur when you have a spotty connection for example come through as bare exceptions, but they're technically transient. It might make sense to make these `Transient` -- if we see enough of them, the batch will fail anyway.

src/fenic/_inference/cohere/cohere_batch_embeddings_client.py
```diff
@@ -0,0 +1,197 @@
...
+            request: The request to generate a key for
+            
+        Returns:
+            A unique key for the request
+        """
+        # Include preset information in the key for proper deduplication
+        profile_config = self._profile_manager.get_profile_by_name(request.model_profile)
+        key_components = [
+            request.doc,
+            str(profile_config.output_dimensionality),
```
bcallender, line 154: that's a good point. can you also add this for gemini embeddings?

src/fenic/_inference/cohere/cohere_profile_manager.py
```diff
@@ -0,0 +1,73 @@
...
+        Returns:
+            Processed Cohere-specific profile configuration
+            
+        Raises:
+            ConfigurationError: If dimensionality is invalid
+        """
+        output_dimensionality = None
+        if profile.embedding_dimensionality:
+            if not self.model_parameters.validate_dimensions(profile.embedding_dimensionality):
+                raise ConfigurationError(
```
bcallender: I moved this validation to occur in `validate_models` in `config.py` (what I wrote the tests for that were just merged), so it happens earlier on in the process of configuration. My implementation *should* cover your use case too since the fields are named the same way, and you should be able to remove the check from here.

bcallender:
Could you also add the cohere module namespace to the list of `NOISY_LOGGER_NAMES` in `logging.py`? We don't want our logs to get full of cohere related things.

src/fenic/_inference/cohere/cohere_batch_embeddings_client.py
YoungVor: Good callout, I'll check the error behavior

src/fenic/_inference/cohere/cohere_batch_embeddings_client.py
YoungVor: got it.  I was following the google example, but no reason to do that here

src/fenic/_inference/cohere/cohere_batch_embeddings_client.py
YoungVor: will do!

YoungVor:
>  Could you also add the cohere module namespace to the list of NOISY_LOGGER_NAMES in logging.py? We don't want our logs to get full of cohere related things.

Ah, that's useful!  Yes, will add it there.  Adding anthropic too.

src/fenic/_inference/cohere/cohere_profile_manager.py
YoungVor: makes sense!  I'll change it and add a second unit test for cohere dimensions next to the ones you just added to confirm.

YoungVor: @bcallender I realized that we were also allowing any aritrary args to go through, which means if the user mistypes, they will never know and just get default behavior.  Added Pydantic check to ensure each Profile only allows the given args, plus a unit test for each

src/fenic/_inference/cohere/cohere_batch_embeddings_client.py
YoungVor: They do use httpx, but I checked the behavior and they do get used when the corresponding 400, 401, and 403 get returned.  I don't think these should be Transient, but we could make the default catch-all a Transient (I wasn't sure which you meant).  Made the change!

https://github.com/cohere-ai/cohere-python/blob/e47e143faf688c633359d5009c784cfb2867ce79/src/cohere/embed_jobs/raw_client.py

src/fenic/_inference/cohere/cohere_batch_embeddings_client.py
YoungVor: good point.  I'll tell claude that ;) 

Removed the uneeded check

src/fenic/_backends/local/model_registry.py
```diff
@@ -8,6 +8,9 @@
     OpenAIBatchChatCompletionsClient,
     OpenAIBatchEmbeddingsClient,
 )
+from fenic._inference.cohere.cohere_batch_embeddings_client import (
```
rohitrastogi: Please do not import this at the top level. This must be a conditional import in `_initialize_embedding_model`, otherwise fenic init will fail if fenic[cohere] is not installed.

src/fenic/_backends/local/model_registry.py
```diff
@@ -202,6 +206,14 @@ def _initialize_embedding_model(self, model_config: ResolvedModelConfig) -> Embe
                     profiles=model_config.profiles,
                     default_profile_name=model_config.default_profile
                 )
+            elif isinstance(model_config, ResolvedCohereModelConfig):
+                rate_limit_strategy = UnifiedTokenRateLimitStrategy(rpm=model_config.rpm, tpm=model_config.tpm)
+                client = CohereBatchEmbeddingsClient(
```
rohitrastogi, line 221: conditional import.

src/fenic/_backends/local/model_registry.py
YoungVor: Thanks for the catch!  I see this is related to your slack last night, makes sense

src/fenic/_backends/local/model_registry.py
YoungVor: done!

src/fenic/_inference/cohere/cohere_batch_embeddings_client.py
```diff
@@ -0,0 +1,193 @@
...
+from fenic.core.metrics import RMMetrics
+
+
+class CohereBatchEmbeddingsClient(ModelClient[FenicEmbeddingsRequest, List[float]]):
+    """Client for making batch requests to Cohere's embeddings API."""
+
+    def __init__(
+        self,
+        rate_limit_strategy: UnifiedTokenRateLimitStrategy,
+        model: str = "embed-v4.0",
```
rohitrastogi: Consider not having a default value here and for the other models. This can lead to bugs where we unintionally pass none to the model (due to a bug in processing user input) and us thus accidentally sett the model to embed-v4.0 even though null should have propagated, eventually causing a failure

rohitrastogi (commented):
lgtm, but will defer to @bcallender for the stamp

src/fenic/_inference/cohere/cohere_batch_embeddings_client.py
YoungVor: that makes sense, will remove

Issue #114, opened 2025-07-29
Support returning LogicalTypes in UDFs


YoungVor:
### Support Returning LogicalTypes in UDFs

Currently, UDFs in Fenic do not support returning LogicalTypes. We should extend the UDF system to allow functions to return logical types defined in Fenic.

Fenic UDFs are implemented by wrapping Polars' `map_elements` expression. However, Polars has no awareness of Fenic’s logical types—specifically `JsonType` and `EmbeddingType`. Although these logical types are internally represented using Polars types (`JsonType` as `pl.Utf8` and `EmbeddingType` as `pl.Array`), they enforce additional invariants:

* `JsonType` must be a valid JSON string.
* `EmbeddingType` must be a vector with a fixed dimensionality, depending on the embedding model.

Since Polars isn't aware of these constraints, it cannot enforce them.

### Proposed Change

Introduce a wrapper in `_convert_udf_to_map_elements` that validates the output of the UDF to ensure it conforms to the expected LogicalType and its invariants.

This will allow developers to safely return logical types from UDFs while preserving the guarantees those types provide.

PR #117, opened 2025-07-29
fix: Validate udf does not return LogicalType

YoungVor:
# Summary

Imporve UDF type checking - Validate that the UDF is not returning a logical type since that's currently unsupported.

# What Changed

* Added LogicalType to dtype class heirarchy
* is_logical_type to check whether Array or Structs contain LogicalType
* udf() will validate that the return type is not a logical type
* Added unit tests for logical type validation and for checking a mismatching return type

# Out of scope
Add support for returned LogicalTypes in UDFs (see [issue #114](https://github.com/typedef-ai/fenic/issues/114))

moved from [112](https://github.com/typedef-ai/fenic/pull/112)


src/fenic/api/functions/builtin.py
```diff
@@ -303,6 +304,9 @@ def _udf_wrapper(*cols: ColumnOrName) -> Column:
 
         return _udf_wrapper
 
+    if is_logical_type(return_type):
+        raise ValidationError("return_type must be a LogicalType")
```
rohitrastogi: I think this should be "not a LogicalType" 
Also, the concept of a LogicalType is totally hidden from the user, so this needs to be more user friendly error.

Perhaps we can just say that we don't support the specific return type the user provided.

Also, maybe we can raise Unimplemented instead of Validation to indicate the work is coming.

src/fenic/core/types/datatypes.py
```diff
@@ -73,8 +73,8 @@ class _PrimitiveType(DataType):
     pass
 
 
-class _StringBackedType(DataType):
-    """Marker class for all string-backed logical types."""
+class LogicalType(DataType):
```
rohitrastogi: This should be _LogicalType to indicate that its private. Users don't need to be aware of a LogicalType's existence.

src/fenic/api/functions/builtin.py
YoungVor: Ha, yes, message doesn't make sense at all.

Definitely, all those points make sense!  Will make the changes.

src/fenic/core/types/datatypes.py
YoungVor: makes sense.

src/fenic/core/_utils/type_inference.py
```diff
@@ -96,3 +97,13 @@ def _find_common_supertype(type1: DataType, type2: DataType, path="") -> DataTyp
         return StructType(merged_fields)
 
     raise TypeInferenceError(f"Incompatible types: {type1} vs {type2}", path)
+
+def is_logical_type(type: DataType) -> bool:
```
rohitrastogi: nit: consider moving this to core/types/datatypes.py

src/fenic/core/error.py
```diff
@@ -49,6 +49,12 @@ class ValidationError(FenicError):
     pass
 
 
+class UnimplementedError(FenicError):
```
rohitrastogi: we don't need a fenic error. we can just use the standard Python NotImplementedError, like we do in other parts of the codebase. eg. https://github.com/typedef-ai/fenic/blob/1d7c388005bbe097bf510b8169c855c83a5a865e/src/fenic/api/lineage.py#L131

YoungVor: ah, makes sense.  Done

src/fenic/core/_utils/type_inference.py
YoungVor: done

src/fenic/core/types/datatypes.py
```diff
@@ -488,6 +488,17 @@ def is_dtype_numeric(dtype: DataType) -> bool:
     return dtype in (IntegerType, FloatType, DoubleType)
 
 
+def is_logical_type(type: DataType) -> bool:
```
rohitrastogi: please make this a private method

rohitrastogi: nvmd.

PR #120, opened 2025-07-31
feat: MCP server example

german-typedef:
Starting from @cpard  example and moving it into the examples folder.
A few changes:
1. Moved from print statements to use logging (so log errors get redirected to the stderr)
2. `populate_tables.py` Made the code more modular (calling functions instead of a full script)
3. `server.py` Using a singleton for the Fenic session instead of a global
4. Fix all trunk issues

### DIFF
--- a/src/fenic/_backends/local/model_registry.py
+++ b/src/fenic/_backends/local/model_registry.py
@@ -136,7 +136,7 @@ class SessionModelRegistry:
             raise InternalError(f"Language Model with alias '{alias.name}' not found in configured models: {sorted(list(self.language_model_registry.models.keys()))}")
         return language_model_for_alias
 
-    def get_embedding_model(self, alias: Optional[str] = None) -> EmbeddingModel:
+    def get_embedding_model(self, alias: Optional[ResolvedModelAlias] = None) -> EmbeddingModel:
         """Get an embedding model by alias or return the default model.
 
         Args:
@@ -152,9 +152,9 @@ class SessionModelRegistry:
             raise InternalError("Requested embedding model, but no embedding models are configured.")
         if alias is None:
             return self.embedding_model_registry.default_model
-        embedding_model_for_model_alias = self.embedding_model_registry.models.get(alias)
+        embedding_model_for_model_alias = self.embedding_model_registry.models.get(alias.name)
         if embedding_model_for_model_alias is None:
-            raise InternalError(f"Embedding Model with model name '{alias}' not found in configured models: {sorted(list(self.embedding_model_registry.models.keys()))}")
+            raise InternalError(f"Embedding Model with model name '{alias.name}' not found in configured models: {sorted(list(self.embedding_model_registry.models.keys()))}")
         return embedding_model_for_model_alias
 
     def shutdown_models(self):
--- a/src/fenic/_backends/local/session_state.py
+++ b/src/fenic/_backends/local/session_state.py
@@ -63,7 +63,7 @@ class LocalSessionState(BaseSessionState):
     def get_language_model(self, alias: Optional[ResolvedModelAlias] = None) -> LanguageModel:
         return self._model_registry.get_language_model(alias)
 
-    def get_embedding_model(self, alias: Optional[str] = None) -> EmbeddingModel:
+    def get_embedding_model(self, alias: Optional[ResolvedModelAlias] = None) -> EmbeddingModel:
         return self._model_registry.get_embedding_model(alias)
 
     def get_model_metrics(self) -> tuple[LMMetrics, RMMetrics]:
--- a/tests/_backends/local/functions/test_embed.py
+++ b/tests/_backends/local/functions/test_embed.py
@@ -15,6 +15,7 @@ from fenic import (
     semantic,
     text,
 )
+from fenic._backends.local.session_state import LocalSessionState
 from fenic.api.session import (
     OpenAILanguageModel,
     SemanticConfig,
@@ -22,6 +23,7 @@ from fenic.api.session import (
     SessionConfig,
 )
 from fenic.core._inference.model_catalog import ModelProvider
+from fenic.core._logical_plan.resolved_types import ResolvedModelAlias
 from fenic.core.error import TypeMismatchError, ValidationError
 
 
@@ -332,3 +334,8 @@ def test_cosine_similarity_special_cases(local_session):
     similarities = result["cosine_sim"].to_list()
 
     assert all(np.isnan(sim) for sim in similarities)
+
+def test_fetch_embedding_model_by_alias(local_session):
+    session_state: LocalSessionState = local_session._session_state
+    embedding_model = session_state.get_embedding_model(alias=ResolvedModelAlias(name="embedding"))
+    assert embedding_model


examples/mcp/docs-server/fenic_getting_started_guide.md
```diff
@@ -0,0 +1,355 @@
+# Getting Started with Fenic
```
cpard: What is the difference between this file and the README.md? Why we need both?

examples/mcp/docs-server/server.py
```diff
@@ -168,7 +204,7 @@ def search(query: str, max_results: int = 30) -> str:
 
           # Filter only public API elements
           df = df.filter(
-              (fc.col("is_public") == True) &
+              (fc.col("is_public")) &
               (~fc.col("qualified_name").contains("._"))
           )
```
cpard: The current implementation of searching has an issue that makes it not ideal for agents. Agents like Claude Code tend to search for multiple terms at once so they will submit a query like this: "dataframe load text" or even better if it's instructed that the search takes regexp as input "dataframe | load | text". The current implementation ignores that and tries to match the whole string which obviously fails. Before we merge this, we should change the search behavior to make it more agent friendly.

german-typedef: Sounds good!
I'll work on the change.

examples/mcp/docs-server/fenic_getting_started_guide.md
german-typedef: The file focuses more on the first few lines of code that you'll write with Fenic, the initial Key setup is the same, the rest of the file is somewhat different.

examples/mcp/docs-server/server.py
german-typedef: Updated. I've split the search to be able to account for multiple terms vs. regex.

examples/mcp/docs-server/fenic_getting_started_guide.md
```diff
@@ -0,0 +1,355 @@
...
+- Automatic batch optimization for API calls
+- Built-in retry logic and rate limiting
+- Token counting and cost tracking
+
+#### Semantic Operations\*\*
+
+- `semantic.analyze_sentiment` - Built-in sentiment analysis
+- `semantic.classify` - Categorize text with few-shot examples
+- `semantic.extract` - Transform unstructured text into structured data with schemas
+- `semantic.group_by` - Group data by semantic similarity
```
rohitrastogi: this api is now with_cluster_labels

examples/mcp/docs-server/fenic_getting_started_guide.md
```diff
@@ -0,0 +1,355 @@
...
+
+df = fc.DataFrame(data)
+print(df)
+```
+
+### Reading Data
+
+```python
+# Read from CSV
+df = fc.read_csv("data.csv")
```
rohitrastogi: these aren't the correct read apis
it's session.read.csv/session.read.parquet

examples/mcp/docs-server/fenic_getting_started_guide.md
```diff
@@ -0,0 +1,355 @@
...
+
+# Read from JSON
+df = fc.read_json("data.json")
```
rohitrastogi: we dont have a json reader yet

examples/mcp/docs-server/fenic_getting_started_guide.md
```diff
@@ -0,0 +1,355 @@
...
+    fc.semantic.extract(fc.col("text"), ProductReview)
+)
+```
+
+#### Semantic Filtering
+
+```python
+# Filter rows using natural language predicates
+positive_reviews = df.filter(
+    fc.semantic.predicate("Is this review positive? {text}")
```
rohitrastogi: need to update these to use new jinja work

examples/mcp/docs-server/fenic_getting_started_guide.md
cpard: I would actually prefer this document to be dynamically generated when the pipeline for populating the data is run. Otherwise the getting started will have to be updated manually every time we make changes

cpard (changes requested):
It's going towards the right direction! A few additional comments.

1. The fenic_getting_started_guide.md should be generated when the populate_tables.py is run to ensure that it reflects the latest version of the API. 

2. For the search functionality, is it possible to add a few tests to make sure that the behavior of the functions is the one expected and have a point of reference against the queries the models will be running when in use?

examples/mcp/docs-server/populate_tables.py
```diff
@@ -0,0 +1,239 @@
...
+def _load_fenic_api():
+    """Load the Fenic API using Griffe."""
+    logger.info("Loading Fenic API with Griffe...")
+    # Load fenic API using Griffe
+    loader = griffe.GriffeLoader()
+    fenic_api = loader.load("fenic")
+    return fenic_api
+
+
+def _extract_api_elements(module: griffe.Module, parent_path: str = "") -> List[Dict[str, Any]]:
```
cpard, line 62: server.py needs to be cleaned up. The populate_tables logic is also present in there.

examples/mcp/docs-server/populate_tables.py
```diff
@@ -0,0 +1,239 @@
...
+
+def _setup_session() -> fc.Session:
+    # Use the same directory setup as the MCP server
+    logger.info("Setting up session...")
+    work_dir = os.environ.get("FENIC_WORK_DIR", os.path.expanduser("~/.fenic"))
+    os.makedirs(work_dir, exist_ok=True)
+    os.chdir(work_dir)
+    logger.info(f"Creating documentation tables in: {work_dir}")
+
+    # Configure fenic session (same as MCP server)
```
cpard: Now with fenic 0.3.0 we can omit adding semantic config when is not needed and this is the case with the server. It will simplify a lot setup of the MCP server. Please, remove this part from the server.

examples/mcp/docs-server/fenic_getting_started_guide.md
german-typedef: LOL
Thanks! This was the AI generated doc, need a few changes.

examples/mcp/docs-server/fenic_getting_started_guide.md
german-typedef: Working on them.

examples/mcp/docs-server/fenic_getting_started_guide.md
german-typedef: The doc is now generated dynamically when populate_tables is executed, using the latest version of the API.

examples/mcp/docs-server/fenic_getting_started_guide.md
german-typedef: Doc is generated dynamically when the API tables are generated.

examples/mcp/docs-server/populate_tables.py
german-typedef: Per our conversation, this has been separated.

examples/mcp/docs-server/populate_tables.py
german-typedef: Done.

examples/mcp/docs-server/server.py
```diff
@@ -0,0 +1,472 @@
...
+    session = FenicSession().get_session()
+    overview = session.table("fenic_summary").select("project_summary").to_pydict()["project_summary"]
+    structure = session.table("hierarchy_df").filter((fc.col("is_public")) & (fc.col("type") != "attribute") & (~fc.col("name").starts_with("_"))).select("qualified_name", "name", "type", "depth", "path_parts").to_pydict()
+    tree = tree_to_string(build_tree(structure))
+    result = f"## Fenic Project Overview\n\n{overview}\n\n## Fenic API Tree\n\n{tree}"
+    return result
+
+@mcp.tool()
+def get_started() -> str:
+    """Get a guide on How to get started with Fenic."""
```
rohitrastogi: do we need better descriptions for these tools?

german-typedef: I've updated the description.

examples/mcp/README.md
cpard, line 1: The content here assumes that mcp is a top level directory and part of the fenic architecture kind of thing. Now that it's under examples, the copy doesn't make that much sense, especially the part about how to add a new MCP server.

my question is, do we want to have this in examples or we want to have as it was before, where mcp is a top level construct? @rohitrastogi @bcallender what do you guys think?

cpard (approved):
overall looks good, assuming the generated getting started/how to guide will be accurate. We need to test this in practice as I'm also not sure how much value it adds to the agent compared to letting the agent figure out things using the search tool. I left a comment about the readme document and if we want to keep this under examples or make MCPs something that should be part of the project but I'd like to hear from the rest of the team what they think.

examples/mcp/README.md
rohitrastogi: the prompts for the getting started guide are so constrained, that if you do want to generate them, you might as well hard code them or retrieve them from the docstrings statically (by id somehow). i agree with kostas that the agent should just make search requests just-in-time for particular apis instead of loading in a summary at the beginning (if this summary is wrong, the entire agent context will be poisoned)

bcallender: Yeah I don't think we will be adding additional MCP servers -- it makes more sense that this is just an example of how one might use `fenic` to curate and search a dataset for an MCP use-case. We are ideally going to have some actual separate implementation of this that is more refined that we expose as an MCP server that users can hit with Claude or similar and ask questions of.

german-typedef: Thanks for the feedback. Given this, we can do the following:
1. I agree with @bcallender , we should have this as an example on Fenic's usage, with that I'll propose we remove the examples/mcp/README.md
2. Regarding the getting started guide, agreed with @rohitrastogi about the possibility of this being wrong and causing the context to poison the rest of the execution. I'll remove it.

PR #123, opened 2025-08-01
draft: jinjify sem ops

rohitrastogi:
todos:
- add tests for expression eq
- update demos to use new apis

### DIFF
--- a/src/fenic/_backends/local/physical_plan/utils.py
+++ b/src/fenic/_backends/local/physical_plan/utils.py
@@ -1,3 +1,5 @@
+from typing import Optional, Union
+
 import polars as pl
 
 
@@ -49,3 +51,62 @@ def _build_target_dtype(dtype: pl.DataType) -> pl.DataType:
         ]
         return pl.Struct(new_fields)
     return dtype
+
+
+# =============================================================================
+# Semantic Join-related utilities
+# =============================================================================
+
+def normalize_column_before_join(
+    df: pl.DataFrame,
+    col: Union[str, pl.Expr],
+    alias: str
+) -> tuple[pl.DataFrame, Optional[str]]:
+    """Normalize a column for join operations by applying a consistent alias.
+
+    This method handles both existing columns (string names) and derived
+    expressions. Derived expressions are computed and added to the DataFrame,
+    while existing columns are simply renamed.
+
+    Args:
+        df: DataFrame to normalize
+        col: Column specification - either a column name (str) or expression (pl.Expr)
+        alias: Target alias for the normalized column
+
+    Returns:
+        Tuple of:
+        - Modified DataFrame with normalized column
+        - Original column name if col was a string, None if it was an expression
+    """
+    if isinstance(col, pl.Expr):
+        # Add derived column with alias
+        return df.with_columns(col.alias(alias)), None
+    else:
+        # Rename existing column
+        return df.rename({col: alias}), col
+
+def restore_column_after_join(
+    df: pl.DataFrame,
+    original_name: Optional[str],
+    alias: str
+) -> pl.DataFrame:
+    """Restore column to original state after join operation.
+
+    This reverses the normalization from normalize_column_before_join:
+    - Renames aliased columns back to their original names
+    - Drops temporary columns that were created from expressions
+
+    Args:
+        df: DataFrame to restore
+        original_name: Original column name (None if column was derived)
+        alias: Current alias of the column
+
+    Returns:
+        DataFrame with restored column state
+    """
+    if original_name:
+        # Restore original column name
+        return df.rename({alias: original_name})
+    else:
+        # Drop temporary derived column
+        return df.drop(alias)
--- a/src/fenic/_backends/local/semantic_operators/base.py
+++ b/src/fenic/_backends/local/semantic_operators/base.py
@@ -2,6 +2,7 @@ from abc import ABC, abstractmethod
 from typing import Generic, List, Optional, TypeVar
 
 import polars as pl
+from jinja2 import Template
 
 from fenic._inference.language_model import InferenceConfiguration, LanguageModel
 from fenic._inference.types import FewShotExample, LMRequestMessages
@@ -89,7 +90,7 @@ class BaseOperator(Generic[ModelResponseType, OperatorOutputType], ABC):
 
     request_sender: RequestSender[ModelResponseType]
 
-    def __init__(self, request_sender: RequestSender[ModelResponseType]):
+    def __init__(self, input: pl.Series, request_sender: RequestSender[ModelResponseType], examples: Optional[BaseExampleCollection]):
         """Initializes the component with a request sender.
 
         Args:
@@ -97,6 +98,8 @@ class BaseOperator(Generic[ModelResponseType, OperatorOutputType], ABC):
                 Component responsible for sending requests and retrieving model responses.
         """
         self.request_sender = request_sender
+        self.input = input
+        self.examples = examples
 
 
     def execute(self) -> pl.Series:
@@ -109,12 +112,60 @@ class BaseOperator(Generic[ModelResponseType, OperatorOutputType], ABC):
         responses = self.request_sender.send_requests(prompts)
         return pl.Series(self.postprocess(responses))
 
-    @abstractmethod
     def build_request_messages_batch(self) -> List[Optional[LMRequestMessages]]:
-        """Construct the LLM prompts from the operator's input.
+        messages_batch = []
+        for document in self.input:
+            if not document:
+                messages_batch.append(None)
+            else:
+                messages_batch.append(
+                    self.build_request_messages(document)
+                )
+        return messages_batch
+
+    def build_request_messages(self, input: str) -> LMRequestMessages:
+        """Construct a prompt from a single input and optional in-context examples."""
+        return LMRequestMessages(
+            system=self.build_system_message(),
+            user=self.build_user_message(input),
+            examples=self.build_examples() if self.examples else [],
+        )
+
+    def build_examples(self) -> List[FewShotExample]:
+        """Convert the stored example collection into prompt message pairs.
 
         Returns:
-            A list of Prompts or None entries, one per row.
+            List of (UserMessage, AssistantMessage) tuples.
+        """
+        formatted_examples: List[FewShotExample] = []
+
+        for example in self.examples.examples:
+            formatted_examples.append(
+                self.build_example(example)
+            )
+
+        return formatted_examples
+
+    @abstractmethod
+    def build_system_message(self) -> str:
+        """Construct the system message for the prompt.
+        This typically includes task instructions or context.
+        """
+        pass
+
+    @abstractmethod
+    def build_user_message(self, input: str) -> str:
+        """Construct the user message for the prompt.
+        This typically includes the input text or a rendered version of the input.
+        """
+        pass
+
+    @abstractmethod
+    def build_example(self, example) -> FewShotExample:
+        """Convert the stored example collection into prompt message pairs.
+
+        Returns:
+            FewShotExample.
         """
         pass
 
@@ -154,50 +205,18 @@ class BaseSingleColumnInputOperator(
             examples (Optional[BaseExampleCollection]):
                 Optional labeled examples to include in the prompt for few-shot learning.
         """
-        super().__init__(request_sender)
-        self.input = input
-        self.examples = examples
+        super().__init__(input, request_sender, examples)
 
-    def build_request_messages_batch(self) -> List[LMRequestMessages | None]:
-        messages_batch = []
-        for document in self.input:
-            if not document:
-                messages_batch.append(None)
-            else:
-                messages_batch.append(
-                    self.build_request_messages(
-                        document, self.build_examples() if self.examples else []
-                    )
-                )
-        return messages_batch
-
-    def build_request_messages(
-        self, input: str, examples: List[FewShotExample]
-    ) -> LMRequestMessages:
-        """Construct a prompt from a single input and optional in-context examples."""
-        return LMRequestMessages(
-            system=self.build_system_message(),
-            user=self.build_user_message(input),
-            examples=examples,
-        )
-
-    def build_examples(self) -> List[FewShotExample]:
+    def build_example(self, example) -> FewShotExample:
         """Convert the stored example collection into prompt message pairs.
 
         Returns:
             List of (UserMessage, AssistantMessage) tuples.
         """
-        formatted_examples: List[FewShotExample] = []
-
-        for example in self.examples.examples:
-            formatted_examples.append(
-                FewShotExample(
-                    user=self.build_user_message(example.input),
-                    assistant=self.convert_example_to_assistant_message(example),
-                )
-            )
-
-        return formatted_examples
+        return FewShotExample(
+            user=self.build_user_message(example.input),
+            assistant=self.convert_example_to_assistant_message(example),
+        )
 
     def convert_example_to_assistant_message(self, example) -> str:
         """Extract the assistant message text from a labeled example."""
@@ -234,6 +253,7 @@ class BaseMultiColumnInputOperator(
         self,
         input: pl.DataFrame,
         request_sender: RequestSender[ModelResponseType],
+        jinja_template: Template,
         examples: Optional[BaseExampleCollection],
     ):
         """Initializes the request component with input data, a request sender, and optional examples.
@@ -246,51 +266,19 @@ class BaseMultiColumnInputOperator(
             examples (Optional[BaseExampleCollection]):
                 Optional labeled examples to include for few-shot prompting.
         """
-        super().__init__(request_sender)
-        self.input = input
-        self.examples = examples
+        super().__init__(input, request_sender, examples)
+        self.jinja_template = jinja_template
 
-    def build_request_messages_batch(self) -> List[LMRequestMessages | None]:
-        messages_batch = []
-        for input in self.input.iter_rows(named=True):
-            if not input or not all(input.values()):
-                messages_batch.append(None)
-            else:
-                messages_batch.append(
-                    self.build_request_messages(
-                        input, self.build_examples() if self.examples else []
-                    )
-                )
-        return messages_batch
-
-    def build_request_messages(
-        self,
-        input: dict[str, str],
-        formatted_examples: List[FewShotExample],
-    ) -> LMRequestMessages:
-        """Construct a prompt from a dict of column values and in-context examples."""
-        return LMRequestMessages(
-            system=self.build_system_message(),
-            user=self.build_user_message(input),
-            examples=formatted_examples,
-        )
-
-    def build_examples(self) -> List[FewShotExample]:
+    def build_example(self, example) -> FewShotExample:
         """Convert example input/output pairs to message format for in-context learning.
 
         Returns:
             List of (UserMessage, AssistantMessage) tuples.
         """
-        formatted_examples = []
-        for example in self.examples.examples:
-            formatted_examples.append(
-                FewShotExample(
-                    user=self.build_user_message(example.input),
-                    assistant=self.convert_example_to_assistant_message(example),
-                )
-            )
-
-        return formatted_examples
+        return FewShotExample(
+            user=self.build_example_user_message(example.input),
+            assistant=self.convert_example_to_assistant_message(example),
+        )
 
     def convert_example_to_assistant_message(self, example):
         """Extract assistant message text from a labeled example."""
@@ -305,8 +293,7 @@ class BaseMultiColumnInputOperator(
         """
         pass
 
-    @abstractmethod
-    def build_user_message(self, input: dict[str, str]) -> str:
+    def build_example_user_message(self, input: dict[str, str]) -> str:
         """Construct the user message from a row of named input values.
 
         Args:
@@ -315,4 +302,8 @@ class BaseMultiColumnInputOperator(
         Returns:
             A string containing the user message content.
         """
-        pass
+        return self.jinja_template.render(input)
+
+    def build_user_message(self, rendered_input: str) -> str:
+        """Construct the user message from a rendered input string."""
+        return rendered_input
--- a/src/fenic/_backends/local/semantic_operators/join.py
+++ b/src/fenic/_backends/local/semantic_operators/join.py
@@ -1,16 +1,13 @@
 import logging
-import re
-from typing import List, Optional
+from typing import Optional
 
 import polars as pl
 
+import fenic._backends.local.polars_plugins  # noqa: F401
 from fenic._backends.local.semantic_operators.predicate import Predicate
-from fenic._backends.local.semantic_operators.utils import (
-    uppercase_instruction_placeholder,
-)
 from fenic._constants import (
-    EXAMPLE_LEFT_KEY,
-    EXAMPLE_RIGHT_KEY,
+    LEFT_ON_KEY,
+    RIGHT_ON_KEY,
 )
 from fenic._inference.language_model import LanguageModel
 from fenic.core._logical_plan.resolved_types import ResolvedModelAlias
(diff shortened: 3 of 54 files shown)


src/fenic/_backends/local/physical_plan/utils.py
```diff
@@ -49,3 +51,62 @@ def _build_target_dtype(dtype: pl.DataType) -> pl.DataType:
         ]
         return pl.Struct(new_fields)
     return dtype
+
+
+# =============================================================================
+# Semantic Join-related utilities
+# =============================================================================
+
+def normalize_column_before_join(
```
YoungVor, line 60: Do these need to be private _normalize_column_before_join?  Or because its in _backend, they won't show up in the doc strings?

src/fenic/_backends/local/semantic_operators/base.py
```diff
@@ -154,50 +205,18 @@ def __init__(
...
-            user=self.build_user_message(input),
-            examples=examples,
-        )
-
-    def build_examples(self) -> List[FewShotExample]:
+    def build_example(self, example) -> FewShotExample:
         """Convert the stored example collection into prompt message pairs.
 
         Returns:
             List of (UserMessage, AssistantMessage) tuples.
```
YoungVor: Returns: FewShotExample

src/fenic/_backends/local/semantic_operators/base.py
```diff
@@ -246,51 +266,19 @@ def __init__(
...
-            user=self.build_user_message(input),
-            examples=formatted_examples,
-        )
-
-    def build_examples(self) -> List[FewShotExample]:
+    def build_example(self, example) -> FewShotExample:
         """Convert example input/output pairs to message format for in-context learning.
 
         Returns:
             List of (UserMessage, AssistantMessage) tuples.
```
YoungVor: FewShotExample

src/fenic/_backends/local/semantic_operators/base.py
```diff
@@ -109,12 +112,60 @@ def execute(self) -> pl.Series:
...
+            system=self.build_system_message(),
+            user=self.build_user_message(input),
+            examples=self.build_examples() if self.examples else [],
+        )
+
+    def build_examples(self) -> List[FewShotExample]:
+        """Convert the stored example collection into prompt message pairs.
+
+        Returns:
+            List of (UserMessage, AssistantMessage) tuples.
```
YoungVor: nit: List of FewShotExamples

src/fenic/_backends/local/semantic_operators/join.py
```diff
@@ -1,43 +1,40 @@
...
+    LEFT_ON_KEY,
+    RIGHT_ON_KEY,
 )
 from fenic._inference.language_model import LanguageModel
 from fenic.core._logical_plan.resolved_types import ResolvedModelAlias
 from fenic.core.types import JoinExampleCollection, PredicateExampleCollection
 
 logger = logging.getLogger(__name__)
 
+# TODO(rohitrastogi): Make this a guid so it doesn't collide with any column names in a user dataframe.
```
YoungVor, line 18: Do you plan on adding the uid in the current PR?

src/fenic/api/dataframe/semantic_extensions.py
```diff
@@ -142,50 +140,53 @@ def with_cluster_labels(
...
+        left_on: Column,
+        right_on: Column,
         examples: Optional[JoinExampleCollection] = None,
         model_alias: Optional[Union[str, ModelAlias]] = None
     ) -> DataFrame:
         """Performs a semantic join between two DataFrames using a natural language predicate.
 
-        That evaluates to either true or false for each potential row pair.
+        This method evaluates a boolean predicate for each potential row pair between the two DataFrames,
+        including only those pairs where the predicate evaluates to True.
```
YoungVor, line 145: nit:  Can we include 'predicate' in the jinja_template var name?  Or say 'predicate in the jinja_template' in the comment?  Just to make it clear that the jinja_template contains the predicate

YoungVor (commented):
Haven't finished my pass yet, will finish my pass later today. Posting a few minor nits

src/fenic/core/_logical_plan/expressions/basic.py
```diff
@@ -164,9 +174,11 @@ def to_column_field(self, plan: LogicalPlan, session_state: BaseSessionState) ->
                 f"get_item cannot be applied to type {expr_type}. Supported types: ArrayType, StructType."
             )
 
-
     def children(self) -> List[LogicalExpr]:
-        return [self.expr]
+        return [self.expr, self.index]
+
+    def _eq_specific(self, other: IndexExpr) -> bool:
```
bcallender: could we put this simple implementation into a mixin class the `LogicalExpr`s could inherit? If an expr is a `UnparameterizedExpr` or something, then it would get this behavior for free

src/fenic/_backends/local/semantic_operators/sim_join.py
```diff
@@ -10,14 +10,17 @@
 from fenic._constants import VECTOR_INDEX_DIR
 from fenic.core.types.enums import SemanticSimilarityMetric
 
+# TODO(rohitrastogi): Make these guids so they don't collide with any column names in a user dataframe.
 # LanceDB column names
 DISTANCE_COL_NAME = "_distance"
 # IMPORTANT: Lance expects a column named "vector" in the table.
 VECTOR_COL_NAME = "vector"
```
YoungVor, line 16: should we add more __ to the system column names?  (Until the guids)

src/fenic/core/_logical_plan/jinja_validation.py
```diff
@@ -177,6 +183,46 @@ def from_jinja_template(cls, template: str) -> VariableTree:
         # Build the final schema tree
         return cls._build_schema_tree(relevant_accesses)
 
+    def filter_used_expressions(self, exprs: List[Union[ColumnExpr, AliasExpr]]) -> List[Union[ColumnExpr, AliasExpr]]:
+        """Filters expressions to only those used in the template, validating all template variables are defined.
+
+        Args:
+            exprs: List of expression objects to filter
+            logger: Optional logger for warnings (if None, warnings are not logged)
```
bcallender: still in args?

src/fenic/api/functions/semantic.py
```diff
@@ -36,39 +36,41 @@
 
 @validate_call(config=ConfigDict(arbitrary_types_allowed=True, strict=True))
 def map(
-    instruction: str,
-    examples: Optional[MapExampleCollection] = None,
-    schema: Optional[type[BaseModel]] = None,
+        jinja_template: str,
+        /,
+        *,
+        examples: Optional[MapExampleCollection] = None,
```
YoungVor, line 43: what happens if the user tries to include one of these variable names in the jinja, like { examples } or { temperature }?  I guess, it would probably fail pydantic validation because they try to cast examples as a columnOrString

src/fenic/core/types/semantic_examples.py
```diff
@@ -32,7 +35,7 @@ class MapExample(BaseModel):
     string or structured model used in a semantic.map operation.
     """
 
-    input: Dict[str, str]
+    input: Dict[str, Any]
```
bcallender, line 38: ~~does it really _have_ to be `Any`?~~

I looked further in the PR -- yes I can see why it is any now. But it is `Any` with an asterisk right? It has to be `Any` that can be converted to a `DataType` with `infer_dtype_from_pyobj` right? Asking for a friend who works on serde.

bcallender: Actually -- if this is any, they could put a `BaseModel` here with the same layout as the struct they're passing in (say, it was the result of a previous operation) instead of making a dict with the same fields. I don't think that would be supported here right? since `infer_dtype_from_pyobj` doesn't deal with base models, you would need `convert_pydantic_type_to_custom_struct_type`

tests/_backends/local/functions/test_semantic_predicate.py
```diff
@@ -159,6 +144,204 @@ def test_semantic_predicate_without_models():
...
+                    }
+                ]
+            }
+        ]
+    })
+
+    result = source.select(
+        semantic.predicate(
+            dedent("""\
+                Given the following resume and job requirements, determine if the candidate is a good fit for the job.
```
bcallender, line 216: this is pretty slick

tests/_backends/local/semantic_operators/test_join.py
```diff
@@ -1,16 +1,46 @@
...
+    3. Does the tone match what {{ right_on }} enthusiasts generally expect?
+    4. Are there elements that might disappoint someone specifically seeking {{ right_on }} content?""").strip()
+
+    GOOD_WILL_HUNTING = "Good Will Hunting is a 1997 American drama film directed by Gus Van Sant and written by Ben Affleck and Matt Damon. It stars Robin Williams, Damon, Affleck, Stellan Skarsgård and Minnie Driver. The film tells the story of janitor Will Hunting, whose mathematical genius is discovered by a professor at MIT."
+    SPIDER_MAN = "Spider-Man is a 2002 American superhero film based on the Marvel Comics character Spider-Man. Directed by Sam Raimi from a screenplay by David Koepp, it is the first installment in Raimi's Spider-Man trilogy."
+
+    GOOD_WILL_HUNTING_RENDERED_TEMPLATE = jinja2.Template(TEMPLATE).render(
+        left_on="Good Will Hunting is a 1997 American drama film directed by Gus Van Sant and written by Ben Affleck and Matt Damon. It stars Robin Williams, Damon, Affleck, Stellan Skarsgård and Minnie Driver. The film tells the story of janitor Will Hunting, whose mathematical genius is discovered by a professor at MIT.",
```
bcallender: use `GOOD_WILL_HUNTING`? unless you're testing that the render itself works

tests/_backends/local/semantic_operators/test_join.py
```diff
@@ -1,16 +1,46 @@
...
+        right_on="Drama",
+    )
+    SPIDER_MAN_RENDERED_TEMPLATE = jinja2.Template(TEMPLATE).render(
+        left_on="Spider-Man is a 2002 American superhero film based on the Marvel Comics character Spider-Man. Directed by Sam Raimi from a screenplay by David Koepp, it is the first installment in Raimi's Spider-Man trilogy.",
```
bcallender: and `SPIDER_MAN`?

src/fenic/core/_logical_plan/expressions/base.py
```diff
@@ -131,17 +159,17 @@ def children(self) -> List[LogicalExpr]:
     @abstractmethod
     def _infer_dynamic_return_type(self, arg_types: List[DataType], plan: LogicalPlan, session_state: BaseSessionState) -> DataType:
         """Must be implemented by subclass for dynamic return type inference.
-        
+
```
bcallender, line 161: this formatting change is so confusing...trunk flags me for trailing whitespace literally all the time, I can't understand how it could have gotten through ci in the first place.

bcallender (commented):
I'm still working my way through it, wanted to give you something to at least start with

src/fenic/_backends/local/transpiler/expr_converter.py
```diff
@@ -311,6 +314,14 @@ def _convert_aggregate_expr(self, logical: AggregateExpr) -> pl.Expr:
...
+            group_context_names = list(logical.group_context_exprs.keys()) if logical.group_context_exprs else None
+            ascending = logical.ascending if logical.order_by_expr else None
+            polars_exprs = [self._convert_expr(logical.input_expr).alias(DATA_COLUMN_NAME)]
+            for name, expr in logical.group_context_exprs.items():
+                polars_exprs.append(self._convert_expr(expr).alias(name))
+            if logical.order_by_expr:
+                polars_exprs.append(self._convert_expr(logical.order_by_expr).alias(SORT_KEY_COLUMN_NAME))
+            struct = pl.struct(polars_exprs)
 
             def sem_reduce_fn(batch: pl.Series) -> str:
```
bcallender: one thing that confuses me, why does this have a `str` return type if `execute()` returns a ` pl.Series`, of the `list[str]` results? I know it's because at the end of the day, the result of the aggregation will be a single `str` in each column row, i'm just confused as to how it _actually_ works

src/fenic/api/functions/text.py
```diff
@@ -941,7 +941,7 @@ def byte_length(column: ColumnOrName) -> Column:
 def jinja(
     jinja_template: str,
     /,
-    **columns: ColumnOrName
+    **columns: Column
```
YoungVor, line 945: could we never handle column names as the value?

src/fenic/core/_logical_plan/expressions/base.py
```diff
@@ -50,6 +50,34 @@ def children(self) -> List[LogicalExpr]:
...
+        if len(self_chldren) != len(other_children):
+            return False
+        return all(
+            child1 == child2
+            for child1, child2 in zip(self_chldren, other_children, strict=False)
+        )
+
+    @abstractmethod
+    def _eq_specific(self, other: LogicalExpr) -> bool:
+        """Returns True if the expression is equal to the other expression.
```
YoungVor: nit: I'm not sure if the comment summary is fully accurate - since we rely on checking the types first in __eq__.  Not sure how better to express what _eq_specific does though- maybe True if the expression characteristics are equal?  I'm probably over thinking it, just a thought.

src/fenic/core/_logical_plan/expressions/basic.py
```diff
@@ -164,9 +174,11 @@ def to_column_field(self, plan: LogicalPlan, session_state: BaseSessionState) ->
                 f"get_item cannot be applied to type {expr_type}. Supported types: ArrayType, StructType."
             )
 
-
     def children(self) -> List[LogicalExpr]:
-        return [self.expr]
+        return [self.expr, self.index]
```
YoungVor, line 179: this was a bug?  Did it mean before that we couldn't evaluate a complex expression to use as the index?

src/fenic/core/_logical_plan/expressions/embedding.py
```diff
@@ -89,3 +92,20 @@ def to_column_field(self, plan: LogicalPlan, session_state: BaseSessionState) ->
...
     def children(self) -> List[LogicalExpr]:
         return self._children
+
+    def _eq_specific(self, other: EmbeddingSimilarityExpr) -> bool:
+        # Check metric (always needs to be compared)
+        if self.metric != other.metric:
+            return False
+
+        # Check the type of self.other vs other.other
+        if isinstance(self.other, LogicalExpr) != isinstance(other.other, LogicalExpr):
```
YoungVor, line 103: for my own understanding, I'm curious if is this needed, if other goes into children when its a LogicalExpr, and the children get checked?  

I guess there might be some weird cornercase where you end up getting equivalent children from multiple inequivalent expressions

src/fenic/_backends/local/physical_plan/utils.py
rohitrastogi: the latter.

src/fenic/_backends/local/semantic_operators/join.py
rohitrastogi: no, this should be good enough.

YoungVor (commented):
another few nits and questions, will finish a full pass next time around

src/fenic/core/_logical_plan/expressions/embedding.py
rohitrastogi: Yes exactly. I don't think it's that weird of a corner case given how many exprs just have a single expr child.

src/fenic/core/_logical_plan/expressions/basic.py
rohitrastogi: It was definitely bug. children() isn't used for evaluation of expressions (just validation of them). This was a hard bug to stumble across (you'd have to do some very specific things to trigger).

In any case, now that we have the equality expression tests, we're also thoroughly checking the children() function for each expr.

src/fenic/core/_logical_plan/expressions/base.py
rohitrastogi: fixed.

src/fenic/_backends/local/semantic_operators/sim_join.py
rohitrastogi: Removing the comment. DISTANCE_COL_NAME and VECTOR_COL_NAME are specific for Lance. If you examine the code, you'll notice that we send the bare minimum columns to LanceDB (we don't send anything that can be `vector` or `_distance`) so no risk of collisions.

src/fenic/_backends/local/transpiler/expr_converter.py
rohitrastogi: the signature is wrong, fixed it.

src/fenic/core/_logical_plan/expressions/base.py
rohitrastogi: yeah i have no idea. will prioritize fixing the trunk/ruff formatting issues after this goes in.

tests/_backends/local/semantic_operators/test_join.py
rohitrastogi: done.

tests/_backends/local/semantic_operators/test_join.py
rohitrastogi: done.

src/fenic/core/_logical_plan/jinja_validation.py
rohitrastogi: done.

src/fenic/api/dataframe/semantic_extensions.py
rohitrastogi: done.

src/fenic/_backends/local/semantic_operators/base.py
rohitrastogi: done.

src/fenic/_backends/local/semantic_operators/base.py
rohitrastogi: done.

src/fenic/_backends/local/semantic_operators/base.py
rohitrastogi: done.

src/fenic/core/_logical_plan/expressions/basic.py
rohitrastogi: good idea. done.

src/fenic/core/types/semantic_examples.py
rohitrastogi: Yeah, its a "constrained" Any - I suppose it's expressible as a Union, but even that is not fully expressive.

 I'm not sure the value in supporting BaseModel for the examples because the jinja templates have dynamic keys, so the BaseModels would also have to be dynamic.

src/fenic/api/functions/semantic.py
rohitrastogi: yes, pydantic validate call handles this. because the variable used in the template has no relationship to the actual expressions, the users can pick really any variable in the template, so I'm not worried about collisions with "reserved" variables like "examples", "temperature", etc.

src/fenic/api/functions/text.py
rohitrastogi: the logic was buggy (we previously didn't have tests for the 'else/name' condition, fixed now.

I think ll Jinja-adjacent APIs (jinja(), join() with jinja_template, etc.) should only accept Column, never ColumnOrName.

Allowing strings could be misconstrued as direct string substitution in the template, when they actually refer to column names. Using the `Column` type forces users to be explicit about their intent and prevents confusion about whether strings are literals or column references.

tests/_backends/local/functions/test_semantic_map.py
```diff
@@ -37,72 +39,111 @@ class PersonInfo(BaseModel):
 
 
 def test_semantic_map(local_session):
-    source = local_session.create_dataframe({"name": ["Alice"], "city": ["New York"]})
```
YoungVor, line 40: for my own understanding of the test organization, how are the functions/test_semantic_map different than the semantic_operators/test_map ?

src/fenic/core/types/semantic_examples.py
```diff
@@ -527,34 +618,82 @@ def _as_df_input(self) -> List[Dict[str, Any]]:
...
+        """Process an example dict, establishing or validating types."""
+        # Check for None values in any example
+        for key, value in example_dict.items():
+            if value is None:
+                raise InvalidExampleCollectionError(
+                    f"Example #{example_num}: None values are not allowed. "
+                    f"Field '{key}' is None."
+                )
+
+        if self.is_first_example:
```
YoungVor, line 646: this is cool, just like the pattern for infering schema when reading csv.

src/fenic/core/types/semantic_examples.py
```diff
@@ -527,34 +618,82 @@ def _as_df_input(self) -> List[Dict[str, Any]]:
...
+        # Check that all column types exist in examples
+        missing_fields = set(column_types.keys()) - set(example_types.keys())
+        if missing_fields:
             raise InvalidExampleCollectionError(
-                f"All BaseModel examples must be of the same type. "
-                f"Existing examples are {type(first_example.output).__name__}, "
-                f"but new example is {type(example.output).__name__}."
-            )
\ No newline at end of file
+                f"The following columns are used in the jinja template but missing from examples: "
```
YoungVor, line 687: does it make sense for examples to include types that aren't in our template?  That's different than checking the used_variables in the jinja template right?

src/fenic/core/_logical_plan/expressions/semantic.py
```diff
@@ -180,112 +194,146 @@ def _validate_completion_parameters(self, plan: LogicalPlan, session_state: Base
...
-        # Use mixin's implementation
-        return super().to_column_field(plan, session_state)
+        self._validate_completion_parameters(session_state)
+        data_types: Dict[str, DataType] = {}
+        for expr in self.exprs:
+            data_type = expr.to_column_field(plan, session_state).data_type
+            self.variable_tree.validate_jinja_variable(expr.name, data_type)
+            data_types[expr.name] = data_type
+        if self.examples:
+            self.examples._validate_against_column_types(data_types)
```
YoungVor, line 248: Do we need to validate the column types for classify examples too?

src/fenic/api/functions/semantic.py
```diff
@@ -49,9 +50,13 @@ def map(
     """Applies a generation prompt to one or more columns, enabling rich summarization and generation tasks.
 
     Args:
-        jinja_template: A Jinja2 template for the generation prompt. References column
+        prompt: A Jinja2 template for the generation prompt. References column
             values using {{ column_name }} syntax. Each placeholder is replaced with the
             corresponding value from the current row during execution.
+        strict: If True, when any of the provided columns has a None value for a row,
```
YoungVor, line 56: I see 'strict' getting checked in the expression equivalency.  And no change needed in example validation because you don't allow none there, right?

src/fenic/core/_logical_plan/expressions/base.py
```diff
@@ -199,3 +198,8 @@ def children(self) -> List[LogicalExpr]:
 
     def _eq_specific(self, other: BinaryExpr) -> bool:
         return self.op == other.op
+
+class UnparameterizedExpr:
```
YoungVor, line 202: Much cleaner, very nice!

YoungVor (approved):
A few more nits, but looks good to me as far as i can tell ;)

It was a good opportunity for me to get up to speed with a lot of code

pyproject.toml
```diff
@@ -24,7 +24,7 @@ dependencies = [
   "botocore>=1.38.28",
   "duckdb>=1.1.3",
   "numpy>=2.0.0",
-  "polars>=1.20.0",
+  "polars>=1.20.0,<=1.30.0",
```
bcallender, line 27: flat `<` i think?

rohitrastogi: No, it works for 1.30.0 (this is what's currently in the uv lock)

tests/_backends/local/functions/test_semantic_map.py
rohitrastogi: The tests under /semantic operators test the implementation details of the individual operators (e.g are prompts formatted correctly), whereas the tests under /functions test the integration with dataframes.

src/fenic/core/types/semantic_examples.py
rohitrastogi: We only ever compare the data types of example columns against expression types used in the plans.

src/fenic/core/_logical_plan/expressions/semantic.py
rohitrastogi: No, they are always strings and the classifyexample type enforces this. There's no Jinja for classification.

src/fenic/api/functions/semantic.py
rohitrastogi: Yes

german-typedef (approved):

PR #126, opened 2025-08-04
feat: add support for greatest/least column functions

rohitrastogi:
### Summary
This PR adds SQL-compatible `greatest()` and `least()` functions to find the maximum and minimum values across multiple columns within each row. These functions mirror the behavior of SQL's `GREATEST` and `LEAST` operations.

### What's New
- **`greatest(*cols)`** - Returns the maximum value among the provided columns for each row
- **`least(*cols)`** - Returns the minimum value among the provided columns for each row

### Example Usage
```python
# Find the maximum value across multiple columns
df.select(greatest(col("score1"), col("score2"), col("score3")).alias("best_score"))

# Find the minimum value across multiple columns  
df.select(least(col("price1"), col("price2"), col("price3")).alias("lowest_price"))
```

### Implementation Details
- Functions require at least 2 arguments
- All arguments must be of the same primitive type (IntegerType, FloatType, StringType, BooleanType, etc.)
- Complex types (lists, structs) are not supported and will raise `TypeMismatchError`
- Null handling follows SQL semantics - returns NULL if all values are NULL
- Transpiles to Polars' `max_horizontal()`/`min_horizontal()` for efficient execution

### DIFF
--- a/src/fenic/api/functions/builtin.py
+++ b/src/fenic/api/functions/builtin.py
@@ -15,6 +15,8 @@ from fenic.core._logical_plan.expressions import (
     CoalesceExpr,
     CountExpr,
     FirstExpr,
+    GreatestExpr,
+    LeastExpr,
     ListExpr,
     MaxExpr,
     MinExpr,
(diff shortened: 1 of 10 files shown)


src/fenic/api/functions/builtin.py
```diff
@@ -540,3 +535,78 @@ def coalesce(*cols: ColumnOrName) -> Column:
         Column._from_col_or_name(c)._logical_expr for c in flattened_args
     ]
     return Column._from_logical_expr(CoalesceExpr(flattened_exprs))
+
+@validate_call(config=ConfigDict(strict=True, arbitrary_types_allowed=True))
+def greatest(*cols: ColumnOrName) -> Column:
+    """Returns the greatest value from the given columns for each row.
+
+    This function mimics the behavior of SQL's GREATEST function. It evaluates the input columns
```
YoungVor, line 536: Nit: should mention that the columns must evaluate to a primitive type like string, bool, float, int

src/fenic/core/_logical_plan/signatures/basic.py
```diff
@@ -44,4 +44,14 @@ def register_basic_signatures():
                                                                   type_signature=ArrayWithMatchingElement(),
                                                                   return_type=BooleanType))
 
+    # Greatest - all arguments must be same type as first
+    FunctionRegistry.register("greatest", FunctionSignature(function_name="greatest",
```
bcallender, line 48: as discussed, we can eventually convert this such that it will accept any numeric type and perform return type promotion, but for now this should be fine

bcallender (approved):
yeah, as long as the documentation is made clear that it only accepts primitives, looks good to me

src/fenic/api/functions/builtin.py
rohitrastogi: done.

PR #132, opened 2025-08-06
feat: create `none()` and `empty()` column functions to allow users to populate columns with null/empty values

bcallender:
# Add `empty()` and `null()` functions for creating empty and null values

### TL;DR

Added two new functions to create empty and null values with specific data types, and improved error handling in the `lit()` function.

### What changed?

- Added `null()` function to create null values of a specific data type
- Added `empty()` function to create empty values (empty arrays, empty structs, or nulls for primitive types)
- Updated `lit()` to raise clear error messages when used with `None` or empty collections
- Added special handling for empty arrays in the polars expression converter
- Added comprehensive test cases for the new functions
- Exported the new functions in the public API

### Why make this change?

The `lit()` function couldn't handle empty collections or `None` values because it relied solely on type inference. These new functions provide a clear way to create empty or null values with specific data types, making the API more flexible and user-friendly. The improved error messages in `lit()` guide users to use the appropriate function for their use case.


bcallender:
* **#132** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/132?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/132?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/fenic/api/functions/core.py
```diff
@@ -29,21 +30,79 @@ def col(col_name: str) -> Column:
...
-    return Column._from_logical_expr(literal_expr)
+        if dtype is None:
+            raise ValidationError(f"`lit` failed to infer type for value `{value}`. "
+                                  f"If you are trying to create a literal with `None` or an empty list, "
+                                  f"you must specify the dtype explicitly.") from e
+        inferred_type = dtype
+    else:
+        # Handle dtype override/validation
+        if dtype is not None:
+            if value == {}:  # Empty dict: allow dtype override
```
YoungVor: nit - why not just not infer_dtype_from_pobj if the value is empty

src/fenic/api/functions/core.py
```diff
@@ -29,21 +30,79 @@ def col(col_name: str) -> Column:
...
+        df.select(
+            lit(None, dtype=IntegerType).alias("none")
+        )
+        ```
     Raises:
-        ValueError: If the type of the value cannot be inferred
+        ValidationError: If the type of the value cannot be inferred and no dtype is provided
     """
+    # Handle special cases with helpful error messages
+    if value == {} and dtype is None:
```
YoungVor: If we allow None as a value, I think you need to handle None separately from {}, right? They are not equivalent in python.

YoungVor: Although your unit test includes that, so maybe python is casting it?

src/fenic/api/functions/core.py
```diff
@@ -29,21 +30,79 @@ def col(col_name: str) -> Column:
...
+    Example: Create a literal with a struct
+        ```python
+        # Creating a struct from a populated dict
+        df.select(
+            lit({"a": 1, "b": 2}).alias("struct")
+        )
+
+        # Creating a struct from an empty dict. This will populate all fields of the struct with None
+        df.select(
+            lit({}, dtype=StructType([StructField(name='a', data_type=IntegerType)])).alias("struct")
```
YoungVor: dtype

src/fenic/api/functions/core.py
```diff
@@ -29,21 +30,79 @@ def col(col_name: str) -> Column:
...
     Raises:
-        ValueError: If the type of the value cannot be inferred
+        ValidationError: If the type of the value cannot be inferred and no dtype is provided
     """
+    # Handle special cases with helpful error messages
+    if value == {} and dtype is None:
+        raise ValidationError("`lit` failed to infer type for value `{}`. "
+                              "If you are trying to create a literal with an empty struct, "
+                              "you must specify the dtype explicitly. "
+                              "For example, `lit({}, dtype=StructType([StructField(name='c', data_type=IntegerType)]))`.")
```
YoungVor: dtype

src/fenic/api/functions/core.py
YoungVor: although, maybe it would be better to call the arg data_type so its consistent with our type class objects

src/fenic/api/functions/core.py
```diff
@@ -29,21 +30,79 @@ def col(col_name: str) -> Column:
...
+                                  f"you must specify the dtype explicitly.") from e
+        inferred_type = dtype
+    else:
+        # Handle dtype override/validation
+        if dtype is not None:
+            if value == {}:  # Empty dict: allow dtype override
+                inferred_type = dtype
+            elif inferred_type != dtype:  # Type mismatch: error
+                raise ValidationError(f"User provided dtype {dtype} does not match inferred type {inferred_type} for value `{value}` "
+                                      "If value is not None or an empty list, you need not specify a dtype.")
```
YoungVor: is there any functional difference in specifying a dtype, like forcing an int to be treated like a float?   Or is it always going to infer the dtype?

If there is not functional difference, maybe we should just enforce this (throw if dtype is set and arg is not None or {})

YoungVor (commented):
This is a good catch, we should be able to handle empty lists and None.  Thank you brandon!

I had a few thoughts about the api and some nits.

src/fenic/api/functions/core.py
bcallender: if the dict is empty, we will get back `StructType([]))` which wouldn't match the user's provided `dtype`. We can tell polars to create the column with the user provided dtype and pass the empty dict, polars will fill all the fields with None

src/fenic/api/functions/core.py
bcallender: i'm confused -- here we're checking if the provided value is an empty dict and a dtype was not provided?

src/fenic/api/functions/core.py
bcallender: that's an interesting point. it probably doesn't ever make sense to allow the user to pass a dtype at all, since we would just reject it anyway if it didn't match the inferred type.

src/fenic/api/functions/core.py
YoungVor: ahh, makes sense.

src/fenic/api/functions/core.py
YoungVor: right, you only care about the case where you can't infer because its none/empty

src/fenic/api/functions/core.py
YoungVor: ah, i meant don't you want to check (value == {} or value == None) and dype is None, but i realize now that you might have a schema for that column already

src/fenic/api/functions/core.py
```diff
@@ -28,22 +29,101 @@ def col(col_name: str) -> Column:
     """
     return Column._from_column_name(col_name)
 
+def null(data_type: DataType) -> Column:
+    """Creates a Column expression representing a null value of the given type.
+
+    No matter the data type, the value that will be populated in the column will be `None`.
```
YoungVor: nit: internally are we defining this as null, but it resolves to 'None' in python?

src/fenic/api/functions/core.py
```diff
@@ -28,22 +29,101 @@ def col(col_name: str) -> Column:
...
+        ```
+
+    """
+    return Column._from_logical_expr(LiteralExpr(None, data_type))
+
+def empty(data_type: DataType) -> Column:
+    """Creates a Column expression representing an empty value of the given type.
+
+    - If the data type is an array or struct, the empty value will be an empty array or struct.
+    - Otherwise, the empty value will be `None`, as if you had called `none()` on the data type.
```
YoungVor: nit - could also be more verbose and say that an empty(literal) is equivalent to a null (literal)

src/fenic/api/functions/core.py
```diff
@@ -28,22 +29,101 @@ def col(col_name: str) -> Column:
...
     Args:
         value: The literal value to create a column for
 
+
     Returns:
         A Column expression representing the literal value
-
     Raises:
-        ValueError: If the type of the value cannot be inferred
+        ValidationError: If the type of the value cannot be inferred and no dtype is provided
```
YoungVor: dtype referred from your older version

src/fenic/api/functions/core.py
```diff
@@ -28,22 +29,101 @@ def col(col_name: str) -> Column:
...
     """
+    if value is None:
+        raise ValidationError("Cannot create a literal with value `None`. Use `null()` instead.")
```
YoungVor: nit: null(...)

tests/_backends/local/functions/test_lit.py
```diff
@@ -130,3 +135,136 @@ def test_lit_struct_with_list(local_session):
...
+    ])
+    df = df.select(col("a"), null(nested_struct_type).alias("b"))
+    assert df.schema == Schema([
+        ColumnField(name="a", data_type=IntegerType),
+        ColumnField(name="b", data_type=nested_struct_type),
+    ])
+    result = df.to_polars()
+    assert result["a"][0] == 1
+    assert result["b"][0] is None
+    # Ensuring that when the empty struct is unnested, the fields are all None
```
YoungVor: nit: null struct is unnested

tests/_backends/local/functions/test_lit.py
```diff
@@ -130,3 +135,136 @@ def test_lit_struct_with_list(local_session):
...
+    ])
+    result = df.to_polars()
+    assert result["a"][0] == 1
+    assert result["b"][0] == {"c": None, "d": None}
+    # Ensuring that when the empty struct is unnested, the fields are all None
+    result = result.unnest("b")
+    assert result["c"][0] is None
+    assert result["d"][0] is None
+
+def test_empty_embedding_type(local_session):
```
YoungVor, line 249: Should we add null() for embedding and string_backed too?

src/fenic/api/functions/core.py
YoungVor: oh nevermind, that doesn't account for embedding and string-backed types.  Maybe worth adding "For all other datatypes (literals, embeddings, json, etc) the empty value will be None..."

YoungVor (commented):
A few minor suggestions but looks good!  Much cleaner than adding dtype param to lit

src/fenic/api/functions/core.py
bcallender: well that's a question of semantics. we don't have a `null` DataType, so when we actually transpile this to polars, we are doing `pl.lit(None, dtype=...)`. Inside the polars dataframe, the value is `null`. When you actually pull that out into python, it evaluates to `None`.

src/fenic/api/functions/core.py
bcallender: i've changed it to be a little more verbose and clear.

src/fenic/api/functions/core.py
```diff
@@ -28,22 +29,103 @@ def col(col_name: str) -> Column:
...
+
+    """
+    return Column._from_logical_expr(LiteralExpr(None, data_type))
+
+def empty(data_type: DataType) -> Column:
+    """Creates a Column expression representing an empty value of the given type.
+
+    - If the data type is `ArrayType(...)`, the empty value will be an empty array.
+    - If the data type is `StructType(...)`, the empty value will be an instance of the struct type with all fields set to `None`.
+    - For all other data types, the empty value will be `None`, as if you had called `null(data_type)`.
```
rohitrastogi: nit:  For all other data types, the empty value is `None` (equivalent to calling `null(data_type)`).
don't want to be using "you"

src/fenic/api/functions/core.py
```diff
@@ -28,22 +29,103 @@ def col(col_name: str) -> Column:
     """
     return Column._from_column_name(col_name)
 
+def null(data_type: DataType) -> Column:
+    """Creates a Column expression representing a null value of the given type.
```
rohitrastogi: nit:
"
Creates a Column expression representing a null value of the specified data type.

Regardless of the data type, the column will contain a null (`None`) value.
This function is useful for creating columns with null values of a particular type."

bcallender:
There is still a small change in the expr_converter -- explicitly calling `pl.lit([])` if we create the `LiteralExpr` with `ArrayType` and an empty list.

src/fenic/_backends/local/transpiler/expr_converter.py
```diff
@@ -168,6 +168,8 @@ def _literal_to_polars_expr(value: Any, data_type: DataType) -> pl.Expr:
 
             if isinstance(data_type, ArrayType):
                 elems = [_literal_to_polars_expr(v, data_type.element_type) for v in value]
+                if not elems:
+                    return pl.lit([], dtype=convert_custom_dtype_to_polars(data_type))
```
rohitrastogi, line 172: Why do we need this change? Can't we push this into the api layer?

rohitrastogi: Nvmd

bcallender:
### Merge activity

* **Aug 8, 4:43 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/typedef-ai/fenic/132).
* **Aug 8, 4:44 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/typedef-ai/fenic/132) rebased this pull request as part of a merge.
* **Aug 8, 4:45 PM UTC**: @bcallender merged this pull request with [Graphite](https://app.graphite.dev/github/pr/typedef-ai/fenic/132).

PR #134, opened 2025-08-07
feat: Add gpt-5 support

YoungVor:
### TL;DR

Added support for GPT-5 models and fixed temperature handling for reasoning models.

### What changed?

- Added GPT-5 model family to the model catalog:
  - `gpt-5` with snapshots (`gpt-5-2025-08-07`)
  - `gpt-5-mini` with snapshots (`gpt-5-mini-2025-08-07`)
  - `gpt-5-nano` with snapshots (`gpt-5-nano-2025-08-07`)
- Fixed temperature handling for reasoning models:
  - Removed temperature parameter for OpenAI reasoning models
  - Moved temperature check to the batch client instead of the core module
- Updated default test model from `gpt-4.1-nano` to `gpt-5-nano`

### How to test?

1. run semantic unit tests with each model

### Not in scope

- Support for gpt-5-chat-latest, since it doesn't support Structured Outputs
- Adding verbosity field
- Updating reasoning level fields

### DIFF
--- a/src/fenic/_inference/language_model.py
+++ b/src/fenic/_inference/language_model.py
@@ -1,3 +1,4 @@
+import logging
 from dataclasses import dataclass
 from typing import Optional
 
@@ -18,6 +19,7 @@ from fenic.core._inference.model_catalog import (
 from fenic.core.error import ConfigurationError
 from fenic.core.metrics import LMMetrics
 
+logger = logging.getLogger(__name__)
 
 @dataclass
 class InferenceConfiguration:
@@ -50,18 +52,23 @@ class LanguageModel:
     ) -> list[Optional[FenicCompletionsResponse]]:
         # Create batch requests
         requests = []
+
+        # Check model specific requirements for request params.
+        temperature_param = temperature if self.model_parameters.supports_custom_temperature else None
+        if not temperature_param:
+            logger.warning(f"Model {self.model} does not support custom temperature.  Ignoring temperature parameter.")
+
         for message_list in messages:
             # if there are no messages, set the request as None, so it can be skipped.
             if not message_list:
                 requests.append(None)
                 continue
-
             request = FenicCompletionsRequest(
                 messages=message_list,
                 max_completion_tokens=max_tokens,
                 top_logprobs=top_logprobs,
                 structured_output=response_format,
-                temperature=temperature,
+                temperature=temperature_param,
                 model_profile=model_profile,
             )
             requests.append(request)
--- a/src/fenic/_inference/openai/openai_batch_chat_completions_client.py
+++ b/src/fenic/_inference/openai/openai_batch_chat_completions_client.py
@@ -83,7 +83,6 @@ class OpenAIBatchChatCompletionsClient(ModelClient[FenicCompletionsRequest, Feni
         Returns:
             The response from the API or an exception
         """
-        # Get profile-specific parameters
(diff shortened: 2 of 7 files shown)


YoungVor:
* **#135** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/135?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#134** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/134?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/134?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/fenic/_inference/openai/openai_batch_chat_completions_client.py
```diff
@@ -84,6 +84,10 @@ async def make_single_request(
             The response from the API or an exception
         """
         # Get profile-specific parameters
+        if request.temperature and self._model_parameters.supports_reasoning:
+            # OpenAI does not support parameterizing temperature on models with reasoning.
+            del request.temperature
```
bcallender: do we want to `del` it or just set it to `None`? doing `del` will make it so `request.temperature` throws an error when accessed, no?

YoungVor: as far as i understood, after this logic we're just passing the 'request' dict as **args to the API call, so deleting it just means we aren't passing it to openai API call

YoungVor: ah, i see, I left the old profile check in CompletionsCore.  I think the right behavior is to check the model here, and just add to common_params if it exists in the request.  wdyt?

YoungVor: or, if its more idiomatic, we can just move the check into CompletionsCore.make_single_request(), since we can get the params there too

bcallender: You don't have access to common_params here yet, thats why i did it in the core. If you want to do it here, I would think the best way is to set it to `1` (if we're dealing with a reasoning model) and always add temperature to the `common_params` in core.

src/fenic/_inference/language_model.py
```diff
@@ -55,13 +55,13 @@ def get_completions(
             if not message_list:
                 requests.append(None)
                 continue
-
+            temperature_param = temperature if self.model_parameters.supports_custom_temperature else None
```
bcallender: nit: this can be done outside of the loop. that way we only do it once, and can even add a warning log telling the user that temperature is not supported on this model.

bcallender (approved):
yeah, I think this is much cleaner than trying to handle it client by client -- this way it works everywhere and we just have to change metadata. just a nit around the temperature checking.

src/fenic/_inference/language_model.py
YoungVor: good catch!  updated.

src/fenic/_inference/openai/openai_batch_chat_completions_client.py
YoungVor: spoke offline, agreed on an approach

PR #135, opened 2025-08-07
feat: Add verbosity and minimal reasoning params for gpt5

YoungVor:
### TL;DR

Added support for GPT-5 models with new verbosity parameter and minimal reasoning effort option.  Added Profile validation to ensure users can only use Profiles and params that apply to a given model.

### What changed?

- Added strict profile param checking to the runtime validation step depending on the model
- Added unit tests for the above
- Updated OpenAI dependency from 1.82.0 to 1.99.0 (included a bug fix, and supports new gpt5 params)
- Added support for `verbosity` parameter in OpenAI profiles (low/medium/high)
- Added support for `minimal` reasoning effort level for GPT-5 models
- Updated documentation and examples to reference GPT-5 models as well as O-series models
- Modified profile configuration to handle verbosity and reasoning effort appropriately based on model capabilities

### How to test?

1. Create a configuration with GPT-5 models using the new parameters:
```python
config = SemanticConfig(
    language_models={
        "gpt5": OpenAILanguageModel(
            model_name="gpt-5-mini",
            rpm=1_000,
            tpm=1_000_000,
            profiles={
                "fast": OpenAILanguageModel.Profile(reasoning_effort="minimal", verbosity="low"),
                "thorough": OpenAILanguageModel.Profile(reasoning_effort="high", verbosity="high")
            },
            default_profile="fast"
        )
    },
    default_language_model="gpt5"
)
```
Make sure the requests include the new params

2. Create profile with non GPT-5 model and add the reasoning_effort "minimal" and verbosity. Make sure the requests have reasoning_effort low and remove verbosity

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ dependencies = [
   "polars>=1.20.0,<=1.30.0",
   "tiktoken>=0.9.0",
   "lancedb>=0.22.0",
-  "openai>=1.82.0",
+  "openai>=1.101.0",
   "sqlglot>=26.25.3",
   "pandas>=2.2.2",
   "cloudpickle>=3.1.1",
--- a/src/fenic/api/session/config.py
+++ b/src/fenic/api/session/config.py
@@ -4,13 +4,14 @@ from __future__ import annotations
 
 from enum import Enum
 from pathlib import Path
-from typing import Literal, Optional, Union
+from typing import Any, Literal, Optional, Union
 
 from pydantic import BaseModel, ConfigDict, Field, model_validator
 
 from fenic.core._inference.model_catalog import (
     AnthropicLanguageModelName,
     CohereEmbeddingModelName,
+    CompletionModelParameters,
     EmbeddingModelParameters,
     GoogleDeveloperEmbeddingModelName,
     GoogleDeveloperLanguageModelName,
@@ -37,6 +38,7 @@ from fenic.core._resolved_session_config import (
     ResolvedOpenAIModelProfile,
     ResolvedSemanticConfig,
     ResolvedSessionConfig,
+    Verbosity,
 )
 from fenic.core.error import ConfigurationError, InternalError
 
@@ -197,6 +199,9 @@ class GoogleDeveloperLanguageModel(BaseModel):
                 It is very possible for the model to generate far more thinking tokens than the suggested budget, and for the
                 model to generate reasoning tokens even if thinking is disabled.
 
+        Raises:
+            ConfigurationError: If a profile is set with parameters that are not supported by the model.
+
         Example:
             Configuring a profile with a fixed thinking budget:
 
@@ -268,6 +273,7 @@ class GoogleVertexEmbeddingModel(BaseModel):
     profiles: Optional[dict[str, Profile]] = Field(default=None, description=profiles_desc)
     default_profile: Optional[str] = Field(default=None, description=default_profiles_desc)
 
+
     class Profile(BaseModel):
         """Profile configurations for Google Vertex embedding models.
 
@@ -296,7 +302,6 @@ class GoogleVertexEmbeddingModel(BaseModel):
 
         output_dimensionality: Optional[int] = Field(default=None, ge=768, le=3072, description="Dimensionality of the embedding created by this model")
         task_type: GoogleEmbeddingTaskType = Field(default="SEMANTIC_SIMILARITY", description="Type of the task")
-
 
 class GoogleVertexLanguageModel(BaseModel):
     """Configuration for Google Vertex AI models.
@@ -357,6 +362,9 @@ class GoogleVertexLanguageModel(BaseModel):
                 the prompt, set this to -1. Note that Gemini models take this as a suggestion -- and not a hard limit.
                 It is very possible for the model to generate far more thinking tokens than the suggested budget, and for the
                 model to generate reasoning tokens even if thinking is disabled.
+
+        Raises:
+            ConfigurationError: If a profile is set with parameters that are not supported by the model.
 
         Example:
             Configuring a profile with a fixed thinking budget:
@@ -458,13 +466,15 @@ class OpenAILanguageModel(BaseModel):
         """OpenAI-specific profile configurations.
 
         This class defines profile configurations for OpenAI models, allowing a user to reference
-        the same underlying model in semantic operations with different settings. For now, only
-        the reasoning effort can be customized.
+        the same underlying model in semantic operations with different settings.
 
         Attributes:
-            reasoning_effort: If configuring a reasoning model, provide a reasoning effort. OpenAI has separate o-series reasoning models,
-                for which thinking cannot be disabled. If an o-series model is specified, but no `reasoning_effort` is provided,
-                the `reasoning_effort` will be set to `low`.
+            reasoning_effort: Provide a reasoning effort. Only for gpt5 and o-series models.
+                If reasoning effort is not provided, the `reasoning_effort` will be set to `low` (for o-series models) or `minimal` (for gpt5 models).
+            verbosity: Provide a verbosity level. Only for gpt5 models.
+
+        Raises:
+            ConfigurationError: If a profile is set with parameters that are not supported by the model.
 
         Note:
             When using an o-series reasoning model, the `temperature` cannot be customized -- any changes to `temperature` will be ignored.
@@ -481,6 +491,10 @@ class OpenAILanguageModel(BaseModel):
         reasoning_effort: Optional[ReasoningEffort] = Field(
             default=None, description="The reasoning effort level for the profile"
         )
+        verbosity: Optional[Verbosity] = Field(
+            default=None, description="The verbosity level for the profile"
+        )
+
 
 class OpenAIEmbeddingModel(BaseModel):
     """Configuration for OpenAI embedding models.
@@ -578,8 +592,11 @@ class AnthropicLanguageModel(BaseModel):
         different thinking token budget settings to be applied to the same model.
 
         Attributes:
-            thinking_token_budget: If configuring a model that supports reasoning, provide a default thinking budget in tokens. If not provided,
+            thinking_token_budget: Provide a default thinking budget in tokens. If not provided,
                 thinking will be disabled for the profile. The minimum token budget supported by Anthropic is 1024 tokens.
+
+        Raises:
+            ConfigurationError: If a profile is set with parameters that are not supported by the model.
 
         Note:
             If `thinking_token_budget` is set, `temperature` cannot be customized -- any changes to `temperature` will be ignored.
@@ -855,7 +872,19 @@ class SemanticConfig(BaseModel):
                 language_model_name = language_model.model_name
                 language_model_provider = _get_model_provider_for_model_config(language_model)
 
+                completion_model_params = model_catalog.get_completion_model_parameters(language_model_provider,
+                                                                                 language_model_name)
+                if completion_model_params is None:
+                    raise ConfigurationError(
+                        model_catalog.generate_unsupported_completion_model_error_message(
+                            language_model_provider,
+                            language_model_name
+                        )
+                    )
                 if language_model.profiles is not None:
+                    if not completion_model_params.supports_profiles:
+                        raise ConfigurationError(
+                            f"Model '{model_alias}' does not support parameter profiles. Please remove the Profile configuration.")
                     profile_names = list(language_model.profiles.keys())
                     if language_model.default_profile is None and len(profile_names) > 0:
                         raise ConfigurationError(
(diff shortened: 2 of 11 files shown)


YoungVor:
* **#135** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/135?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/135?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#134** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/134?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/fenic/api/session/config.py
```diff
@@ -458,22 +459,29 @@ class Profile(BaseModel):
...
-            reasoning_effort: If configuring a reasoning model, provide a reasoning effort. OpenAI has separate o-series reasoning models,
-                for which thinking cannot be disabled. If an o-series model is specified, but no `reasoning_effort` is provided,
-                the `reasoning_effort` will be set to `low`.
+            reasoning_effort: If configuring a reasoning model, provide a reasoning effort.
+                - For gpt-5 and o-series reasoning models, if no `reasoning_effort` is provided, the `reasoning_effort` will be set to `low`.
+                - For models that don't support reasoning, the `reasoning_effort` will be ignored.
+                - gpt-5 models, support 'minimal' reasoning effort.  If 'minimal' is provided for non-gpt-5 models, the `reasoning_effort`
+                will be set to `low`.
+            verbosity: An api to control whether answers are short and to the point or long and comprehensive.
+                - If this field conflicts with explicit instructions, the explicit instructions will take precedence.
```
bcallender: What are `explicit instructions`?

bcallender (commented):
I feel like the complexity of the docstrings we're presenting to users when creating the profile indicates that we should probably just have different user facing `Profile` classes, one with `verbosity` and a different Literal that also includes `minimal`. That way, the users won't ever be confused. Something like `ReasoningModelProfile` (for o-series) and `GPT5Profile` (for gpt5). I'm not thrilled by that name, so feel free to come up with something better.  

We can still convert both into a `ResolvedOpenAIModelProfile` beyond the API layer. Then we don't have to do any checks in the `ProfileManager` to change stuff the user specified. 

This does mean that `config.py` gets kind of messy in `resolve_model`, but I think a little complexity there is better than doing a ton of checks in `ProfileManager`, and allows us to expose a cleaner API to the users. What do you think?

src/fenic/api/session/config.py
YoungVor: user instructions/system instructions.  i can go into more detail.  But OpenAI says if 'verbosity' is low, but the instructions specifically say to go into detail, the instruction takes precedence.

YoungVor:
> I feel like the complexity of the docstrings we're presenting to users when creating the profile indicates that we should probably just have different user facing `Profile` classes, one with `verbosity` and a different Literal that also includes `minimal`. That way, the users won't ever be confused. Something like `ReasoningModelProfile` (for o-series) and `GPT5Profile` (for gpt5). I'm not thrilled by that name, so feel free to come up with something better.
> 
> We can still convert both into a `ResolvedOpenAIModelProfile` beyond the API layer. Then we don't have to do any checks in the `ProfileManager` to change stuff the user specified.
> 
> This does mean that `config.py` gets kind of messy in `resolve_model`, but I think a little complexity there is better than doing a ton of checks in `ProfileManager`, and allows us to expose a cleaner API to the users. What do you think?

I was also considering multiple profiles.  I agree with you, I think its worth more complexity in the backend if its more clear for the user.

YoungVor:
@bcallender done, and I added an extra profile validation step to enforce which model supports which profile.

src/fenic/core/_inference/model_catalog.py
```diff
@@ -151,6 +156,37 @@ def supports_dimensions(self, requested_dimensions: int) -> bool:
     "o3-mini",
     "o4-mini",
 ]
+OpenAIProfileClassName = Literal["ReasoningModelProfile", "GPT5ModelProfile"]
```
bcallender: this is an interesting concept, but the implementation seems a little fragile. it might make more sense to add `supported_profile_classes` or something similar to `completion_model_parameters`, so we can just register them with the proper profiles and consolidate the profile validation logic in one place. that way, we can add additional profile types to the other providers in the future (and we don't have to have all of the snapshot versions listed again here).

YoungVor: Ah, that sounds much cleaner.  So I'll just add a supported_profile_classes list in the model parameters and grab them from model_catalog during openAI model init

bcallender: Well at that point it's not even OpenAI specific, we can ensure that all models that can have a profile are registered with the allowed ones

YoungVor: exactly!

YoungVor:
As per conversation with @bcallender , added back the Profile for gemini-2.5-pro.

Also updated openai version to fix recent their recent bug with verbosity

src/fenic/core/_inference/model_catalog.py
```diff
@@ -83,10 +92,12 @@ def __init__(
         allowed_output_dimensions: Union[int, list[int]],
         max_input_size: int,
         default_dimensionality: Optional[int] = None,
+        model_profile_class_name: Optional[str] = None,
     ):
         self.input_token_cost = input_token_cost
         self.output_dimensions = allowed_output_dimensions
         self.max_input_size = max_input_size
+        self.model_profile_class_name = model_profile_class_name
```
bcallender: I don't think doing this with string matching is a sustainable approach. I get why it's happening, you can't import the types from api into core without running into circular import issues. Let me propose another approach:

Define protocols for the different Profile types in `core`, like:
```
  @runtime_checkable
  class LanguageModelProfile(Protocol):
      """Base marker protocol for model profiles"""
     pass

  @runtime_checkable
  class ThinkingTokenBudgetProfile(Protocol):
      """Protocol for Anthropic/Gemini reasoning models"""
      thinking_token_budget: Optional[int]

  @runtime_checkable
  class ReasoningEffortProfile(Protocol):
      """Protocol for reasoning models"""
      reasoning_effort: Optional[ReasoningEffort]

  @runtime_checkable
  class GPT5Profile(Protocol):
      """Protocol for GPT-5 models"""
      reasoning_effort: Optional[ReasoningEffortGPT5]
      verbosity: Optional[Verbosity]
 ```
 and have the profiles implement these protocols, with each implementing `LanguageModelProfile` and one of the protocols with behavior. When configuring the model parameters, set the `model_profile_class` to the interface type you can now import in core (or create an enum and say that the model supports a given `ProfileType`). 

```
  class ProfileType(Enum):
      OPENAI_REASONING = auto()
      OPENAI_GPT5 = auto()
      THINKING_TOKEN_BUDGET = auto()
```

Instead the duplicative `post_init` logic to each `ModelConfig` class, just add one `validate_profile` function that encapsulates all of the logic and call that from our existing `validate_models` validator. 
 ```
   def validate_profile(model_name: str, params: CompletionModelParameters, profile: LanguageModelProfile) -> None:
      if params.supported_profile_type == ProfileType.OPENAI_GPT5:
          if not isinstance(profile, GPT5Profile):
              raise ConfigurationError(f"Model {model_name} requires GPT5Profile")
      elif params.supported_profile_type == ProfileType.OPENAI_REASONING:
          if not isinstance(profile, ReasoningEffortProfile):
              raise ConfigurationError(f"Model {model_name} requires ReasoningProfile")
...
```

bcallender (commented):
Apologies for the back and forth on this -- I don't mean to be picky on this but If we can make this as clean as possible, it will be a lot easier to extend it as we add additional providers and models. I also have an eye towards externalizing the `model_catalog` in json or yaml form, so I want to make sure things are understandable if you're just looking at the configuration data in a separate file.

src/fenic/api/session/config.py
```diff
@@ -482,6 +490,34 @@ class Profile(BaseModel):
             default=None, description="The reasoning effort level for the profile"
         )
 
+    class GPT5ModelProfile(BaseModel):
```
rohitrastogi: Are we sure we aren't complicating the surface area too much here? I agree in an ideal world certain configurations are mapped to certain models, but we also don't want to devolve to per-model configs.  I also find the naming unclear because gpt5 is a reasoning model, but I wouldn't be able to use the reasoningmodelprofile for it.

For simplicity's sake in the api, we could just have one languagemodel profile for open ai and just fail at runtime during sessionconfig creation if the setting is incompatible with the model. This might make it easier to support new models in the future so we don't have to make breaking changes for new models.

YoungVor: I agree, its worth more complexity in our code to reduce the surface area and complexity for the user

YoungVor:
Had a in person conversation with @bcallender and @rohitrastogi , we decided to keep the class API surface area simple for the user, and add explicit checking logic in the validation step.  PTAL, the param validation is at least all in the same place so it shouldn't be hard to maintain as providers (OpenAI...) change their params

src/fenic/core/_inference/model_catalog.py
YoungVor: Spoke in person, this thread is out of date

src/fenic/api/session/config.py
```diff
@@ -462,9 +470,11 @@ class Profile(BaseModel):
         the reasoning effort can be customized.
 
         Attributes:
-            reasoning_effort: If configuring a reasoning model, provide a reasoning effort. OpenAI has separate o-series reasoning models,
-                for which thinking cannot be disabled. If an o-series model is specified, but no `reasoning_effort` is provided,
-                the `reasoning_effort` will be set to `low`.
+            reasoning_effort: Provide a reasoning effort. Only for gpt5 and o-series models.
+                If reasoning effort is not provided, the `reasoning_effort` will be set to `low` (for o-series models) or `minimal` (for gpt5 models).
```
bcallender, line 473: docstring missing verbosity?

src/fenic/core/_inference/model_catalog.py
```diff
@@ -38,8 +38,11 @@ class CompletionModelParameters:
         context_window_length: Maximum number of tokens in the context window
         max_output_tokens: Maximum number of tokens the model can generate in a single request.
         max_temperature: Maximum temperature for the model.
+        supports_parameter_profiles: Whether the model supports parameter profiles.
```
bcallender: nit: why not just `supports_profiles`? I don't think we call them parameter profiles anywhere else. not saying it's wrong, just wondering if you had a specific reason for it.

tests/conftest.py
```diff
@@ -259,7 +259,20 @@ def configure_language_model(model_provider: ModelProvider, model_name: str) ->
...
     # these limits are purposely low so we don't consume our entire project limit while running multiple tests in multiple branches
     if model_provider == ModelProvider.OPENAI:
-        if model_parameters.supports_reasoning:
+        if model_parameters.supports_reasoning and model_parameters.supports_verbosity:
+            language_model = OpenAILanguageModel(
+                model_name=model_name,
+                rpm=500,
+                tpm=100_000,
+                profiles={
+                    "minimal": OpenAILanguageModel.Profile(reasoning_effort="minimal", verbosity="low"),
```
bcallender, line 268: nit: this is more of a philosophical nit, but based on my understanding i think we should advocate for keeping verbosity as low as possible -- it's okay for the model to reason more, but for a dataframe usecase, i feel like we always want the final output result to be as small as possible while still fulfilling the task.

pyproject.toml
```diff
@@ -27,7 +27,7 @@ dependencies = [
   "polars>=1.20.0,<=1.30.0",
   "tiktoken>=0.9.0",
   "lancedb>=0.22.0",
-  "openai>=1.82.0",
+  "openai>=1.99.9",
```
bcallender: could you run `uv sync` and commit the lock change? i feel like this would generate one.

bcallender (approved):
this looks good, just a couple of nits

tests/conftest.py
YoungVor: ah, good point.  I can keep it low regardless for all the test profiles

pyproject.toml
YoungVor: will do!

src/fenic/core/_inference/model_catalog.py
YoungVor: no particular reason except it sounded right to me in that instant ;). I'll put it back to supports_profiles to keep it consistent

src/fenic/api/session/config.py
YoungVor: good catch, thanks!

PR #108, opened 2025-07-25
feat: add serde registry functions for logical expressions

bcallender:
## Add Protocol Buffer Serde Functions for existing Logical Expressions

This change registers all of the existing LogicalExprs with serializer/deserializer methods. It uses `@singledispatch` to route between all of the different types, so we can organize the serde functions like we do the expressions in `_logical_plan` to preserve clarity. 

### Testing
Added an extensive test of all existing `LogicalExpr` subtypes in `test_expression_serde.py`

PR #107, opened 2025-07-25
refactor: protobuf plan serde framework

bcallender:
## 🚀 Add Protocol Buffer Serialization System for Logical Plans and Expressions

  This PR introduces a comprehensive Protocol Buffer-based serialization/deserialization system for Fenic's logical plans, expressions, and data types,
  replacing the legacy cloudpickle implementation.

  ### Overview

  The new system provides efficient, cross-language compatible serialization through:

  - **Centralized Context System**: All serde operations go through `SerdeContext` with built-in error handling and automatic field path tracking for
  precise debugging
  - **Type-Safe Dispatch**: Uses Python's `@singledispatch` to automatically route serialization based on object types (70+ expression types supported)
  - **Modular Architecture**: Clean separation with dedicated modules for expressions, plans, and data types matching the codebase structure
  - **Compression Support**: Integrated zstandard compression reduces serialized size by ~60-80%

  ### Key Improvements Over Cloudpickle

  - **Cross-Language Compatibility**: Protocol Buffers enable serialization across different languages/runtimes (Python, Go, Rust)
  - **Structured Error Handling**: Automatic field path tracking provides precise error locations (e.g., `"root.filter.predicate.left.name"`)
  - **Performance**: More efficient binary format with smaller serialized size (see benchmarks below)
  - **Type Safety**: Explicit schema definitions prevent version compatibility issues
  - **Introspection**: Protobuf schemas are self-documenting and versionable

  ### Architecture

  ```python
  # High-level API
  plan_bytes = ProtoSerde.serialize(logical_plan)
  logical_plan = ProtoSerde.deserialize(plan_bytes, session_state)

  # Context-based operations with automatic error handling
  context = create_serde_context()
  expr_proto = context.serialize_logical_expr(SerdeContext.EXPR, expression)

  What's Included

  Protocol Buffer Definitions (protos/logical_plan/v1/)

  - expressions.proto: 70+ expression types (Column, Literal, Semantic operations, Aggregates, etc.)
  - plans.proto: All logical plan nodes (Source, Transform, Sink operations)
  - datatypes.proto: Fenic data types including specialized types (Embedding, Transcript, Markdown, etc.)
  - complex_types.proto: Complex structures (ScalarValue, ResolvedClassDefinition, NumpyArray, etc.)
  - enums.proto: Shared enumerations (Operator, JoinType, ChunkLengthFunction, etc.)

  Core Implementation (src/fenic/core/_serde/)

  - proto_serde.py: Main API with compression support
[...]

PR #136, opened 2025-08-08
refactor: remove `type[BaseModel]` from expr constructors to make serde easier

bcallender:
# Add jsonschema dependency and refactor structured output handling

This PR refactors how structured output formats are handled in semantic operators by:

1. Adding `jsonschema` as a dependency for schema validation
2. Creating a new `ResolvedResponseFormat` class that encapsulates JSON schema handling
3. Replacing direct Pydantic model usage with the resolved format in all semantic operators
4. Moving schema validation and parsing logic from utility functions into the `ResolvedResponseFormat` class
5. Adding schema normalization as required for provider-specific requirements (OpenAI, etc.)
6. Replaces Pydantic validation of output data with jsonschema based validation. 

This change makes serialization/deserialization easier by eliminating the need to serialize/deserialize the `type[BaseModel]` class information at the expression layer -- we can much more easily serialize the json schema, fenic struct type, and the prompt_schema_definition. 

Tested with Anthropic, OpenAI and Google Completion Models


bcallender:
* **#107** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/107?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#108](https://github.com/typedef-ai/fenic/pull/108) <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/108?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#140](https://github.com/typedef-ai/fenic/pull/140) <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/140?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#136** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/136?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/136?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tests/_backends/local/functions/test_semantic_map.py
```diff
@@ -262,7 +262,7 @@ def test_semantic_map_schema_validation_mismatch_error(local_session):
     ))
 
     # This should raise ValidationError - PersonInfo example doesn't match ProductSummary schema
-    with pytest.raises(ValidationError, match="Expected `semantic.map` example output to be an instance of"):
```
YoungVor, line 265: Why is this one ValidationError instead of InvalidExample like the string example below?

YoungVor, line 265: Why is this one ValidationError instead of InvalidExample like the string example below?

src/fenic/core/_logical_plan/resolved_types.py
```diff
@@ -20,3 +35,238 @@ class ResolvedModelAlias:
...
+    This class wraps a JSON schema dictionary to make it clear that this is
+    the resolved format used for model client communication, as opposed to
+    the original Pydantic model type.
+
+    Attributes:
+        schema: The JSON schema dictionary.
+        struct_type: The StructType of the model.
+            Only generated as required. This is only needed if the Operator returns the struct type itself (e.g. semantic.map, semantic.extract).
+            In cases like semantic.classify, the struct type is not returned, only the class labels.
+        prompt_schema_definition: The description of the schema that will be used in the prompt. Only generated if struct_type is generated.
```
YoungVor, line 59: What does it mean for the struct_type to be generated?  Does it get returned any time we provide a struct to their API, or is it provider specific?

src/fenic/core/_logical_plan/resolved_types.py
```diff
@@ -20,3 +35,238 @@ class ResolvedModelAlias:
...
+
+        - Ensures additionalProperties: false on every object (incl. nested, $defs, items)
+        - Strips volatile metadata keys (title, $id, $schema) for stable comparison
+        """
+        def deep_copy(obj: Any) -> Any:
+            return json.loads(json.dumps(obj))
+
+        def strip_metadata(s: Dict[str, Any]) -> None:
+            for k in ("title", "$id", "$schema"):
+                if k in s:
```
YoungVor: do you have to add a recursive step here too (like with additional_props)?  I think theycan appear in nested fields (allOf, nested, etc)

src/fenic/core/_logical_plan/resolved_types.py
```diff
@@ -20,3 +35,238 @@ class ResolvedModelAlias:
...
+        return {
+            "type": "json_schema",
+            "json_schema": {
+                "name": name,
+                "schema": self._normalized_schema_for_openai(),
+                "strict": True,
+            },
+        }
+
+    # Internal: stricter normalization for OpenAI parse API. This is required -- previously
```
YoungVor: is OpenAI the only provider where we need to worry about the schema parse API?  Iirc for anthropic we're just baking it into the prompt but not sure about google

src/fenic/core/_logical_plan/resolved_types.py
```diff
@@ -20,3 +35,238 @@ class ResolvedModelAlias:
...
+    # we were passing the pydantic model directly to the SDK, where all of this was handled internally.
+    # Now we are passing the schema directly to the SDK, so we need to normalize it to the OpenAI parse API expectations.
+    def _normalized_schema_for_openai(self) -> Dict[str, Any]:
+        """Return normalized schema with OpenAI parse expectations.
+
+        - additionalProperties: false for all objects
```
YoungVor: Could we just use the openai normalized schema everywhere, rather than having two concepts of normalized schema?

src/fenic/core/_logical_plan/expressions/semantic.py
```diff
@@ -151,17 +155,16 @@ class SemanticExtractExpr(ValidatedDynamicSignature, SemanticExpr):
     def __init__(
         self,
         expr: LogicalExpr,
-        schema: type[BaseModel],
```
YoungVor, line 154: What is the naming convention for schema, vs response_format, vs type, etc?

src/fenic/_backends/local/semantic_operators/analyze_sentiment.py
```diff
@@ -139,7 +142,7 @@ def __init__(
                 inference_config=InferenceConfiguration(
                     max_output_tokens=MAX_TOKENS_DETERMINISTIC_OUTPUT_SIZE,
                     temperature=temperature,
-                    response_format=SENTIMENT_ANALYSIS_MODEL,
+                    response_format=ResolvedResponseFormat.from_pydantic_model(SENTIMENT_ANALYSIS_MODEL, generate_struct_type=False),
```
YoungVor: nit: should we define the responseFormat  for SENTIMENT_ANALYSIS_MODEL and other pre-defined structures in our library?

YoungVor (commented):
Some minor comments and questions, but overall looks great!

src/fenic/core/_utils/schema.py
```diff
@@ -97,10 +97,6 @@ def convert_pydantic_type_to_custom_struct_type(
 
     fields = []
     for field_name, field_info in model.model_fields.items():
-        if field_info.description is None:
```
YoungVor, line 100: why don't we want to enforce descriptions?

bcallender: we are enforcing descriptions for user provided models, check the validation methods above this and how they are called in the API layer to validate the model provided by the user

src/fenic/core/_logical_plan/expressions/semantic.py
bcallender: we're generally moving towards response format -- the jinja pr changed it to response_format for all user facing operations.

src/fenic/core/_logical_plan/resolved_types.py
bcallender: The struct type has nothing to do with the model provider. The struct type is required so `fenic` knows what the `DataType` is for the column that gets returned if you're doing something like `semantic.map` or `semantic.extract` and are returning the result as a struct.

src/fenic/core/_logical_plan/resolved_types.py
bcallender: let me see about simplifying this. openai is very strict and the other providers don't really seem to care very much, but there does seem to be some duplication.

src/fenic/core/_logical_plan/resolved_types.py
YoungVor: Ah, makes sense.

src/fenic/core/_logical_plan/resolved_types.py
```diff
@@ -20,3 +40,169 @@ class ResolvedModelAlias:
...
+            Validated dictionary representation of the model, or None if validation fails
+        """
+        try:
+            if json_resp is None:
+                return None
+            if isinstance(json_resp, str):
+                json_resp = json.loads(json_resp)
+            self.validate_structured_response(json_resp)
+            # Apply defaults from schema to ensure missing optionals become nulls and shapes are consistent.
+            # Required for openai/google -- if for some reason, NONE of the responses have the optional field filled,
```
YoungVor: Interesting, so polars will respect the struct field if its 'None', but not if its an Optional with no value?

Why isn't this a problem for anthropic?  Can't it be the case where it also never fills the optional field for any of the rows?

src/fenic/_backends/local/semantic_operators/analyze_sentiment.py
bcallender: done

src/fenic/core/_logical_plan/resolved_types.py
bcallender: This comment is slightly out of date I think -- this is needed for all providers now that we are passing the `strict` json schema to each of them. The standard json schema created by a pydantic type has a default `None` value for optional fields, so when Anthropic was still taking the standard schema, it would generate `None` for those fields, and we wouldn't have to backfill the defaults.

tests/_backends/local/functions/test_semantic_map.py
bcallender: it should be `InvalidExampleCollectionError`. This only passed the test because `InvalidExampleCollectionError` is a subclass of `ValidatonError`, good catch

src/fenic/core/_logical_plan/resolved_types.py
bcallender: this has been entirely reworked. now we generate a `strict` json schema and use that for all providers.

src/fenic/_inference/google/gemini_native_chat_completions_client.py
```diff
@@ -396,3 +394,36 @@ async def make_single_request(
                 return FatalException(e)
         except Exception as e:  # noqa: BLE001 – catch-all mapped to Fatal
             return TransientException(e)
+
+    def _prepare_schema(self, response_format: ResolvedResponseFormat) -> dict[str, Any]:
+        def remove_additional_properties(result: Any) -> Any:
```
YoungVor, line 410: nit: consider adding a comment giving context on the fields google strips away, maybe with a link to the source - and call out the drawback that this has to run on every call

src/fenic/core/_utils/json_schema_utils.py
```diff
@@ -0,0 +1,119 @@
...
+    for key in ("anyOf", "oneOf"):
+        alts = node.get(key)
+        if isinstance(alts, list):
+            for alt in alts:
+                if isinstance(alt, dict) and alt.get("type") != "null":
+                    return alt
+    return node
+
+def to_strict_json_schema(schema: Dict[str, Any]) -> Dict[str, Any]:
+    """Return a strict JSON Schema suitable for OpenAI/Google.
```
YoungVor: nit: can we call out that google has additional refinement steps, and that OpenAI is the most strict,

YoungVor (approved):
I like that we're mostly using the same schema resolution process everywhere now for everyone (except google).

Pending a few nits on comments, looks good to me!

src/fenic/_inference/google/gemini_native_chat_completions_client.py
bcallender: it's quite odd, their documentation says that it does support this field, but if you send it it will error back 400 every time. i added some comments.

PR #138, opened 2025-08-08
chore: Contributor documentation for adding models and params

YoungVor:
### TL;DR

Added documentation for contributing model support to the project.

### What changed?

- Added a new section "Adding model support" to the CONTRIBUTING.md file with detailed instructions for:
  - Adding language or embedding models to the model catalog
  - Adding model parameters to the Profile
  - Adding model-specific support for parameters
- Updated the note about API keys to only mention OpenAI (removed Anthropic)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -167,7 +167,7 @@ uv sync --extra=cloud
 uv run pytest -m cloud tests
 ```
 
-> ⚠️ Note: All tests require a valid OpenAI/Anthropic API key set in the environment variables.
+> ⚠️ Note: All tests require a valid OpenAI API key set in the environment variables.
 
 ---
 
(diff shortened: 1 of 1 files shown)


YoungVor:
* **#138** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/138?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/138?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

CONTRIBUTING.md
```diff
@@ -196,6 +196,46 @@ To run the demo notebooks:
 
 ---
 
+## Adding model support
```
rohitrastogi, line 199: Can we add a bit more context on how inference works in fenic? I think it'd make sense to add a readme in the /_inference directory and link to it here.

Specifically it would be good to mention:
- inference currently runs in a map_batches Polars udf
- we have LanguageModel/EmbeddingModel classes that each own a ModelClient that the udfs use
- ModelClients work though a producer/consumer mechanism - Polars threads produce requests, consumer threads submit them asynchronously in an event loop to maximize thoughput without spawning too many threads
- ModelClients have RateLimitStrategies to make sure we don't overwhelm inference providers and exceed rate limits

It's also worth mentioning how each provider has its own set of embedding/language model configs set in the sessionconfig where we specify the model name, rate limits, and any special per model/provider functionality.

Lastly, I'd describe the nuances of token accounting and how an integration needs to fit in there. Would be useful to link some related PRs for past model integration to the guide.

YoungVor: Definitely agree, there is a lot of detail around adding provider/model support and it deserves its own readme.  That also means we can add more detail.

I'll take a second pass and add this detail as well

bcallender:
Closing this as part of a PR backlog prune. The branch has gone stale against current `main`, and we do not expect to merge it in this form. If the underlying idea is still useful, we should revive it as a fresh, focused issue or PR against current `main`. Thanks for the work here.

PR #144, opened 2025-08-13
docs: Update example notebooks

cpard:
I'm updating the notebooks in our examples to add some CTAs, links to other examples from within the notebooks and most importantly to make the notebooks possible to directly run inside Google colab.

I'm also updating the README.md to add a link to open the notebook in Google colab.


cpard:
I've added only the Hello World! so I can get some feedback on how it looks when you interact with it, I'll then update all the notebooks before merging.

cpard:
> I've added only the Hello World! so I can get some feedback on how it looks when you interact with it, I'll then update all the notebooks before merging. 

All examples are updated now.

examples/podcast_summarization/podcast_summarization.ipynb
rohitrastogi, line 1: why is this diff so large?

rohitrastogi: we definitely don't want to have the transcript be an in memory dict. please just add the file to our public s3 (or even google drive) and use wget in the notebook

cpard: I removed the data from the notebooks and I'm doing a wget to get the files during initialization.

rohitrastogi (approved):

PR #148, opened 2025-08-18
fix: only use regex search for the mcp server example

german-typedef:
### TL;DR

Improved search functionality in the docs-server with better regex handling, structured logging, and enhanced error handling.

### What changed?

This uses the changes proposed by Brandon for our own MCP server. We'll use just the regex.

Also, I incorrectly assumed we didn't need the models for the server, that is not correct as store_learnings requires the models. I've added them back to the configuration and README.md.

I've tested also retrieving the information from the learnings (using the full string as well as the terms), this wasn't working properly as store_learnings was failing due to the lack of model configuration.

### DIFF
--- a/examples/mcp/docs-server/server.py
+++ b/examples/mcp/docs-server/server.py
@@ -5,21 +5,41 @@ It is used to search the Fenic codebase and provide documentation for the Fenic 
 import datetime
 import logging
 import os
-import re
 import threading
 import uuid
-from typing import Callable, List, Literal
+from typing import List, Literal
 
-from fastmcp import FastMCP
+import structlog
+from fastmcp import Context, FastMCP
+from fastmcp.exceptions import ToolError, ValidationError
+from search import FenicAPIDocQuerySearch
 from utils.schemas import get_learnings_schema
 from utils.tree_operations import build_tree, tree_to_string
+from utils.validation import validate_and_sanitize_regex
 
 import fenic as fc
 
-logging.basicConfig(level=logging.DEBUG)
-logger = logging.getLogger(__name__)
-logger.setLevel(logging.DEBUG)
+structlog.configure(
+    processors=[
+        structlog.stdlib.add_logger_name,
+        structlog.stdlib.add_log_level,
+        structlog.processors.TimeStamper(fmt="iso"),
+        structlog.processors.StackInfoRenderer(),
+        structlog.processors.format_exc_info,
+        structlog.processors.JSONRenderer(),
+    ],
+    logger_factory=structlog.stdlib.LoggerFactory(),
+    wrapper_class=structlog.stdlib.BoundLogger,
+    cache_logger_on_first_use=True,
+)
 
+# When running this locally we don't want to use stdout, as this will used
+# by the MCP server to communicate back and forth with the client.
+logging.basicConfig(
+    format="%(message)s",
+    level=logging.INFO,
+)
+logger = structlog.get_logger(__name__)
 
 class FenicSession:
     """Singleton class to manage Fenic session."""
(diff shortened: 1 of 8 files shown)


german-typedef:
* **#148** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/148?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/148?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

examples/mcp/docs-server/utils/validation.py
```diff
@@ -0,0 +1,118 @@
...
+    """
+    if len(pattern) >= 2 and pattern.startswith("/") and pattern.rfind("/") > 0:
+        last = pattern.rfind("/")
+        core = pattern[1:last]
+        flags = set(pattern[last + 1 :].lower())
+        return core, flags
+    return pattern, set()
+
+
+def validate_and_sanitize_regex(user_query: str) -> str:
```
bcallender, line 40: so I only added all of this complexity to the deployed version because we don't trust the user inputs -- I'm not sure if we need anything beyond just doing `re.compile` for the example, to keep the attention on the stuff we're building out in `search`. But I suppose it doesn't hurt to have it here too.

examples/mcp/docs-server/server.py
```diff
@@ -53,141 +73,46 @@ def _create_session(cls):
...
-        df = df.filter(
-            (fc.col("is_public")) &
-            (~fc.col("qualified_name").contains("._"))
-        )
-
-        if cls._is_regex(query):
-            logger.debug(f"Searching API docs with regex: {query}")
-            search_df = cls._search_api_docs_regex(df, query)
+        if not os.environ.get("OPENAI_API_KEY") and not os.environ.get("GEMINI_API_KEY"):
+            logger.warn("No API keys found. Set OPENAI_API_KEY and/or GEMINI_API_KEY environment variables.")
```
bcallender: we generally recommend using `GOOGLE_API_KEY` for `GoogleDeveloperLanguageModel` now, though both will work.

bcallender: it might be worth using a google embedding model, or openai language model, so that someone can run the example with only 1 API Key.

examples/mcp/docs-server/server.py
```diff
@@ -227,87 +152,84 @@ def search(query: str, max_results: int = 30) -> str:
...
+
+        query_id = uuid.uuid4()
+        logger.info(
+            f"Performing search for query: {sanitized_query} ({query_id})",
+            query_id=query_id,
+            query=sanitized_query,
+            endpoint="/search",
+        )
         session = FenicSession().get_session()
-        api_doc_query_search = FenicAPIDocQuerySearch()
```
bcallender, line 244: i removed all of this code that queries the learnings df in the deployed version since it is going to be read-only for now. You should probably add it back here, so the learnings are still searched.

examples/mcp/docs-server/utils/validation.py
german-typedef: that's what I was thinking, not required but I think as an example is better to be a bit paranoid.

examples/mcp/docs-server/server.py
german-typedef: I like the idea of using one API key, let me do the change, I'll probably lean towards Google.

examples/mcp/docs-server/server.py
german-typedef: You are right, yep, I was too eager here. Adding it back.

german-typedef: Updated in the next iteration.

examples/mcp/docs-server/server.py
german-typedef: Went the GOOGLE_API_KEY route.

cpard:
I'll defer to @bcallender on this PR, this is pretty much moving his implementation of the functions to the example. Regardless of that, I think more tests are needed to ensure that whatever happens in the future can be tested appropriately for regressions and edge cases.

PR #149, opened 2025-08-19
feat: add support for async udfs (for tool calling)

rohitrastogi:
### Summary

This PR introduces **async user-defined functions (UDFs)** to Fenic, enabling IO-bound operations—like API calls, database queries, or MCP tool calls—within DataFrame transformations while preserving familiar DataFrame semantics.

### Key Features

* **Configurable Concurrency:** Control maximum concurrent executions via `max_concurrency`.
* **Automatic Retries:** Built-in retry logic with exponential backoff for transient failures.
* **Timeout Support:** Per-item timeout configuration to prevent hanging operations.
* **Ordered Results:** Maintains input order despite asynchronous execution.
* **Resource Management:** Automatic task cancellation on errors and bounded memory usage.
* **Type Safety:** Runtime type checking with clear error messages.

### Usage Example

```python
import fenic as fc
from fenic.core.types import IntegerType
import aiohttp

@fc.async_udf(
    return_type=IntegerType,
    max_concurrency=10,
    timeout_seconds=5,
    num_retries=2
)
async def fetch_score(user_id: int) -> int:
    async with aiohttp.ClientSession() as session:
        async with session.get(f"https://api.example.com/score/{user_id}") as resp:
            data = await resp.json()
            return data["score"]

# Apply to DataFrame
df = df.select(
    fc.col("user_id"),
    fetch_score(fc.col("user_id")).alias("score")
)
```

### Implementation Details

**Architecture**

* `EventLoopManager`: Singleton managing a shared event loop across all async operations.
* `AsyncUDFExpr`: New expression type in the logical plan.
* `AsyncUDFSyncStream`: Core execution engine with bounded concurrency and ordered streaming.
* Context Manager: Automatic resource lifecycle management.

**Key Design Decisions**

1. **Unified Event Loop:** Single background thread/loop shared across `ModelClient` and async UDFs.
2. **Smart Buffer Management:** Separate limits for pending tasks (`max_pending`) and buffered results (`max_buffer_size`).
3. **Intelligent Waiting:** Wait for any task normally, wait for next result when buffer is full to prevent deadlocks.
4. **Individual Failure Handling:** Failed items return `None` instead of failing the entire batch.

### Testing

Comprehensive test suite covering:

* Basic async operations
* Concurrency limits
* Retry logic
* Timeout handling
* Struct return types
* Multiple async UDFs in a single query
* Handling of `None` results

### Performance Considerations

* Default limits prevent resource exhaustion (`1000 max pending tasks`, `50k max buffer`).
* Exponential backoff with jitter for retries.
* Lazy task scheduling minimizes memory usage.

### Breaking Changes

* None. This is a new feature and does not affect existing functionality.


tests/_backends/local/functions/test_async_udf.py
```diff
@@ -0,0 +1,471 @@
...
+
+import pydantic
+import pytest
+
+import fenic as fc
+from fenic.core.error import ExecutionError, ValidationError
+from fenic.core.types import IntegerType, JsonType, StringType, StructField, StructType
+
+
+class TestAsyncUDF:
```
german-typedef, line 15: nit: Question about the style here of using a class to group the test cases, any advantages / disadvantages vs just exposing the test functions?

rohitrastogi: We should group the test cases in a class so that we don't have to worry about test names colliding across features. When they are grouped in a class, you only have to worry about unique tests names within a class.

The pytest output is a bit nicer too when the tests are grouped, but other than that, no real difference.

german-typedef: Makes sense! I like the idea, will do for subsequent tests and opportunistically update some of the existing ones.

src/fenic/_backends/local/async_udf_stream.py
```diff
@@ -0,0 +1,167 @@
...
+        self.num_retries = num_retries
+        self.timeout = timeout
+
+        self.max_buffer_size = min(HARD_BUFFER_LIMIT, DEFAULT_BUFFER_MULTIPLIER * max_concurrency)
+        self.max_pending = min(HARD_PENDING_LIMIT, DEFAULT_PENDING_MULTIPLIER * max_concurrency)
+
+        # Make dictionaries and state members for external access
+        self._pending: Dict[int, asyncio.Task] = {}
+        self._results: Dict[int, Any] = {}
+        self._next_to_yield: int = 0
```
german-typedef: nit: might be clearer (at least for me) if this was named: _next_idx_to_yield.

src/fenic/_backends/local/async_udf_stream.py
```diff
@@ -0,0 +1,167 @@
...
+        """Execute a single async call with retries, timeout, and concurrency control."""
+        async with self.semaphore:
+            last_err = None
+            for attempt in range(self.num_retries + 1):
+                try:
+                    return await asyncio.wait_for(self.fn(item), timeout=self.timeout)
+                except Exception as e:
+                    last_err = e
+                    error_msg = str(e) or f"{type(e).__name__} occurred"
+                    logger.info(f"AsyncUDFStream: {error_msg} (attempt {attempt+1}/{self.num_retries+1})")
```
german-typedef: nit: I would log this as an error instead of info

tests/_backends/local/functions/test_async_udf.py
```diff
@@ -0,0 +1,471 @@
...
+        """Test that async UDF respects concurrency limits."""
+        call_times = []
+
+        @fc.async_udf(
+            return_type=IntegerType,
+            max_concurrency=2,  # Only 2 concurrent calls
+            timeout_seconds=5,
+            num_retries=0
+        )
+        async def track_concurrency(x: int) -> int:
```
YoungVor, line 64: Nice trick for measuring concurrency!

tests/_backends/local/functions/test_async_udf.py
```diff
@@ -0,0 +1,471 @@
...
+            track_concurrency(fc.col("x")).alias("doubled")
+        ).to_pylist()
+        elapsed = time.time() - start_time
+
+        # Verify results
+        expected = [{"x": i, "doubled": i * 2} for i in range(1, 6)]
+        assert result == expected
+
+        # Verify timing - should take at least 0.5 seconds (3 batches * 0.2s)
+        assert elapsed >= 0.5, f"Expected at least 0.5s, got {elapsed:.2f}s"
```
YoungVor, line 91: nit: assert less than 1s?

src/fenic/api/functions/builtin.py
```diff
@@ -313,6 +315,97 @@ def _udf_wrapper(*cols: ColumnOrName) -> Column:
...
+                    }
+        ```
+
+    Note:
+        - Individual failures return None instead of raising exceptions
+    """
+
+    def _create_async_udf(func: Callable[..., Awaitable[Any]]) -> Callable:
+        if not inspect.iscoroutinefunction(func):
+            raise ValidationError(
```
YoungVor, line 392: Nit: would be good to have a unit test trying to add a non-async function

src/fenic/api/functions/builtin.py
```diff
@@ -313,6 +315,97 @@ def _udf_wrapper(*cols: ColumnOrName) -> Column:
...
+        async def slow_add(x: int, y: int) -> int:
+            await asyncio.sleep(1)
+            return x + y
+
+        df = df.select(slow_add(fc.col("x"), fc.col("y")).alias("slow_sum"))
+        ```
+
+    Example: API call with retries
+        ```python
+        @async_udf(
```
YoungVor, line 366: Nit: include an example where they call async_udf directly, not using the decorator?

german-typedef (approved):

src/fenic/_backends/local/async_utils.py
```diff
@@ -0,0 +1,147 @@
...
+import asyncio
+import logging
+import threading
+from contextlib import contextmanager
+from typing import Generator, Optional
+
+logger = logging.getLogger(__name__)
+
+
+class EventLoopManager:
```
YoungVor, line 12: Its awesome having a singleton managing the loop for all async clients

src/fenic/_backends/local/async_utils.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+async def _cancel_event_loop_tasks(loop: asyncio.AbstractEventLoop):
+    """Cancels all pending tasks in the given asyncio event loop, except the current task.
+
+    Args:
+        loop: The event loop to cancel tasks for.
+    """
+    asyncio.set_event_loop(loop)
+    tasks = [t for t in asyncio.all_tasks(loop) if t is not asyncio.current_task(loop)]
```
YoungVor, line 144: Thinking about cases where a developer can run into trouble - can we enforce that asyncio tasks on the loop are managed by the EventLoopManager?  (And not through     
     loop = asyncio.get_running_loop()
    task = loop.create_task(my_coroutine())

We could check the all_tasks every once in a while and throw InternalError on count mismatch.. wdyt?

src/fenic/_backends/local/transpiler/expr_converter.py
```diff
@@ -364,6 +370,44 @@ def _convert_udf_expr(self, logical: UDFExpr) -> pl.Expr:
...
+                    timeout=logical.timeout_seconds,
+                    num_retries=logical.num_retries
+                )
+
+                results = []
+                for result in async_udf.call(items):
+                    if isinstance(result, Exception):
+                        # Log the exception before returning None
+                        error_msg = str(result) or f"{type(result).__name__} occurred"
+                        logger.warning(f"Async UDF execution failed: {error_msg}")
```
YoungVor: Nit: Can we add row name, or index, or something telling the user which row failed in the warning?

src/fenic/_backends/local/async_udf_stream.py
```diff
@@ -0,0 +1,167 @@
+import asyncio
+import logging
+import random
+from typing import Any, AsyncGenerator, Awaitable, Callable, Dict, Iterable
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_BUFFER_MULTIPLIER = 10
+HARD_BUFFER_LIMIT = 50_000
+DEFAULT_PENDING_MULTIPLIER = 3
```
YoungVor, line 10: What's the context behind the multipliers?  If the user specifies a concurrency of 5, we'll actually allow 15 tasks to be running in parallel?

Iiuc this is because there maybe items queued in the _pending dict that are already done and waiting for the next stage, which means in practice, on average we're running less than max_concurrency tasks, but I'm curious how we choose this number

src/fenic/_backends/local/async_udf_stream.py
```diff
@@ -0,0 +1,167 @@
...
+            last_err = None
+            for attempt in range(self.num_retries + 1):
+                try:
+                    return await asyncio.wait_for(self.fn(item), timeout=self.timeout)
+                except Exception as e:
+                    last_err = e
+                    error_msg = str(e) or f"{type(e).__name__} occurred"
+                    logger.info(f"AsyncUDFStream: {error_msg} (attempt {attempt+1}/{self.num_retries+1})")
+
+                    if attempt < self.num_retries:
```
YoungVor, line 151: An idea for future - (not necessary for this PR)- can we introduce 'retry count' to Operator and QueryMetrics? 

For now, consider logging total retries upon async_udf completion

src/fenic/_inference/model_client.py
```diff
@@ -396,8 +398,7 @@ async def _enqueue_request(self, queue_item: QueueItem[RequestT]):
         """
         await self.request_queue.put(queue_item)
 
-
-
+    # TODO(rohitrastogi): We should stream the requests to the model client and pipe results back from the background thread to the main thread to avoid unnecessary memory usage.
```
YoungVor, line 401: is there a ticket for this?

src/fenic/_backends/local/async_udf_stream.py
```diff
@@ -0,0 +1,167 @@
...
+    """
+    Async UDF execution engine with bounded concurrency, retries, and ordered results.
+
+    High-level intuition:
+
+    - Goal: Run an async function on many input items while:
+        * Limiting concurrency (max_pending tasks in-flight)
+        * Handling retries and timeouts per item
+        * Yielding results in input order
+        * Enforcing bounded memory usage (max_buffer_size)
```
YoungVor, line 24: (probably out of scope for this PR):  We should think about how to enforce memory bounding based on the memory used by each row, and the memory available to fenic.  One option is to just expose per row memory usage limit to the user API and modify the buffer size accordingly.. But its hard to determine dynamically how much memory is being used by a row at runtime.

The exponential backoff you have might help - if we raise an exception well before an OOM, we could reduce the max_buffer_size, or something...

src/fenic/api/functions/builtin.py
rohitrastogi: Already exists

src/fenic/_backends/local/async_utils.py
rohitrastogi: I think we can enforce the invariant in code review.

src/fenic/_backends/local/async_udf_stream.py
rohitrastogi: every retry is already logged. we can definitely add it to metrics in a followup if folks ask for it

src/fenic/_backends/local/async_udf_stream.py
rohitrastogi: out of scope for this pr. in the future, we can make the constants configurable.

src/fenic/_inference/model_client.py
rohitrastogi: yes

src/fenic/_backends/local/async_udf_stream.py
rohitrastogi: the point of the multipliers is actually to make sure the pending/result buffers don't inhibit concurrency. the reason there's an upper bound is in case the user has such a big max concurrency that it defeats the purpose of processing the data in a streaming manner

src/fenic/_backends/local/async_udf_stream.py
rohitrastogi: done

tests/_backends/local/functions/test_async_udf.py
rohitrastogi: done

src/fenic/_backends/local/transpiler/expr_converter.py
rohitrastogi: updated the logging

src/fenic/api/functions/builtin.py
rohitrastogi: done

src/fenic/_backends/local/async_udf_stream.py
rohitrastogi: I don't think it should be error. I think it should be a warning because we insert none for rows that error.

bcallender (approved):
serde seems fine to me, makes sense to approach it the same way as UDFExpr.

PR #151, opened 2025-08-21
feat: support loading directory content into a dataframe

german-typedef:
### TL;DR

Added a new `DocLoader` class to enable loading files from directories with filtering capabilities and proper error handling.
This implements the logic for traversing the file system.

### What changed?

- Created a new `DocLoader` class that provides functionality to traverse folders and load document files (markdown/json).
- Added support for glob pattern inclusion and regex pattern exclusion when selecting files
- Added a `docs()` method to the reader interface
- Added new error classes for file loading operations
- Added a new DocSource logical plan and DocSourcExec physical plan to support reading from a folder
- Added the serde support for the DocSource
- Added test cases for the DocLoader functionality, session.reader.docs, also serde test cases for DocSource.


german-typedef:
* **#151** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/151?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/151?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/fenic/api/io/reader.py
```diff
@@ -227,3 +228,37 @@ def _read_file(
         from fenic.api.dataframe import DataFrame
 
         return DataFrame._from_logical_plan(logical_node, self._session_state)
+
+    def from_folder(
```
rohitrastogi: the existing csv/parquet interfaces don't have the 'from' prefix.
Can we call this docs() or dir(), instead?

src/fenic/api/io/reader.py
```diff
@@ -227,3 +228,37 @@ def _read_file(
...
+            self,
+            path: str,
+            data_type: DataType,
+            include: Optional[str] = None,
+            exclude: Optional[str] = None,
+            recursive: bool = False,
+    ) -> DataFrame:
+        """Load a DataFrame from a folder.
```
rohitrastogi: I think this docstring is a bit underspecified. Can you add that each file in the directory corresponds to one row in the dataframe? It's also useful to mention what the output schema of the dataframe will be.

src/fenic/_backends/local/utils/folder_loader.py
```diff
@@ -0,0 +1,290 @@
...
+            column_fields=[
+                ColumnField(name="file_path", data_type=StringType),
+                ColumnField(name="started_at", data_type=StringType),
+                ColumnField(name="error", data_type=StringType),
+                ColumnField(name="content", data_type=StringType),
+            ]
+        )
+    
+    @staticmethod
+    def validate_path(
```
rohitrastogi: prefer this method either validates the path OR creates the fingerprint, but not both

src/fenic/_backends/local/utils/folder_loader.py
```diff
@@ -0,0 +1,290 @@
...
+
+        Returns:
+            DataFrame: A dataframe containing the files in the folder.
+        """
+        individual_dfs: List[pl.DataFrame] = []
+
+        if files:
+            logger.debug(f"Processing {len(files)} files")
+            for file_path in files:
+                file_df = FolderLoader._process_single_file(file_path, is_s3_path)
```
rohitrastogi: do we want to parallelize this using a threadpool? We can use a similar technique to the asyncudfstream where we only append to the dataframe as we read so we don't have to have the whole contents of the dataframe in memory twice at once

roughly:
- create an interator over the files, read them in batches, once the batch is complete, append to the dataframe (this way we only have maximum result_set + batch_size in memory instead of 2 * resultset

rohitrastogi (commented):
Some initial directional feedback.

src/fenic/_backends/local/utils/folder_loader.py
```diff
@@ -0,0 +1,290 @@
...
+    ) -> str:
+        """Validate the path and returns a hash calculated from the files in the path."""
+        is_s3_path = is_s3_location(path)
+        files = FolderLoader._enumerate_files(path, is_s3_path, include_pattern, exclude_pattern, recursive)
+        return hash_list_of_strings(files)
+    
+    @staticmethod
+    def verify_supported_data_type(data_type: DataType):
+        """Verify that the data type is supported."""
+        if data_type not in [JsonType, MarkdownType, TranscriptType]:
```
rohitrastogi: we haven't implemented the transcripttype yet, so best to exclude

src/fenic/_backends/local/utils/folder_loader.py
```diff
@@ -0,0 +1,290 @@
...
+
+        Returns:
+            Schema: The schema for the data type
+        """
+        return Schema(
+            column_fields=[
+                ColumnField(name="file_path", data_type=StringType),
+                ColumnField(name="started_at", data_type=StringType),
+                ColumnField(name="error", data_type=StringType),
+                ColumnField(name="content", data_type=StringType),
```
rohitrastogi: should this always be stringtype? I was under the impression that this should be JsonType/MarkdownType? If not, do you expect users to cast after loading the files in?

Should we be looking for particular file formats (.md and .json)?

src/fenic/_backends/local/physical_plan/source.py
```diff
@@ -98,3 +100,41 @@ def _build_lineage(
         source_operator = self._build_source_operator_lineage(materialize_df)
         leaf_nodes.append(source_operator)
         return source_operator, materialize_df
+
+
+class FolderSourceExec(PhysicalPlan):
```
YoungVor: Suggestion - its really more of a DocSource then a FolderSource.. DocSource would be more idiomatic with the way we handle FileSource - they both use paths and globs to determine the list of input files.  The only difference between the two is one inserts in a tabular fashion after checking schema, and DocSource inserts row by row after checking they're all consistently markdown or json

src/fenic/api/io/reader.py
german-typedef: moved it to be "docs".

src/fenic/api/io/reader.py
german-typedef: Updated the doc string with the corresponding note and the schema definition.

src/fenic/_backends/local/utils/folder_loader.py
german-typedef: Updated the code to use the ThreadPoolExecutor and batches of files, it calls df.extend(other_df) to append new batch data to the dataframe.

src/fenic/_backends/local/utils/folder_loader.py
german-typedef: Excluded.

src/fenic/_backends/local/utils/folder_loader.py
german-typedef: Updated, the folder_loader will return the basic schema that doesn't include the casted column.
reader.docs will have a column casted to the corresponding datatypes.
We look at the extension as well.

src/fenic/_backends/local/physical_plan/source.py
german-typedef: Moved it to be "DocSource"

src/fenic/_backends/local/utils/folder_loader.py
german-typedef: Agreed, I've split the method.

src/fenic/_backends/local/physical_plan/source.py
```diff
@@ -98,3 +100,45 @@ def _build_lineage(
         source_operator = self._build_source_operator_lineage(materialize_df)
         leaf_nodes.append(source_operator)
         return source_operator, materialize_df
+
+
+class DocSourceExec(PhysicalPlan):
+    def __init__(
+            self,
+            path: str,
```
YoungVor: Do we still need include, or can you just process the glob in the path argument?  Also, can we accept a list of path globs?  (FileSource has a list "paths")

src/fenic/_backends/local/physical_plan/source.py
```diff
@@ -98,3 +100,45 @@ def _build_lineage(
...
+            data_type: DataType,
+            session_state: LocalSessionState,
+            include: Optional[str] = None,
+            exclude: Optional[str] = None,
```
YoungVor: Should we add exclude and recursive to FileSource?

src/fenic/_backends/local/utils/folder_loader.py
```diff
@@ -0,0 +1,350 @@
+from __future__ import annotations
```
YoungVor: nit: should rename the file to doc_loader

src/fenic/_backends/local/utils/folder_loader.py
```diff
@@ -0,0 +1,350 @@
...
+            schema={
+                "file_path": pl.Utf8,
+                "started_at": pl.Utf8,
+                "error": pl.Utf8,
+                "content": pl.Utf8,
+            }
+        )
+
+    @staticmethod
+    def _enumerate_files_s3(
```
YoungVor: thank you for adding the s3 consideration!

if you rebase after https://github.com/typedef-ai/fenic/pull/157, can you do the same for hf?

src/fenic/_backends/local/utils/folder_loader.py
```diff
@@ -0,0 +1,350 @@
...
+    def calculate_files_hash(
+        files: List[str],
+    ) -> str:
+        """Calculate a hash of a list of files."""
+        return hash_list_of_strings(files)
+    
+    @staticmethod
+    def verify_supported_data_type(data_type: DataType):
+        """Verify that the data type is supported."""
+        if data_type not in [JsonType, MarkdownType]:
```
YoungVor: IMO its better to check this at the API layer

src/fenic/api/io/reader.py
```diff
@@ -227,3 +229,78 @@ def _read_file(
         from fenic.api.dataframe import DataFrame
 
         return DataFrame._from_logical_plan(logical_node, self._session_state)
+
+    def docs(
```
YoungVor, line 266: We could consider a separate api reader.md(paths) and reader.json(paths), to be consistent with reader.csv vs reader.parquet.  But I could imagine many different formats (html, pdf, images, etc) being accepted in future, so I don't feel strongly

src/fenic/api/io/reader.py
```diff
@@ -227,3 +229,78 @@ def _read_file(
...
+            self,
+            path: str,
+            data_type: DataType,
```
YoungVor: Can we accept a Literal[JsonType, Markdown], or Literal["json", "markdown"], and enforce it that way?

tests/_backends/local/io/test_folder_loader.py
```diff
@@ -0,0 +1,153 @@
+"""Tests for the FolderLoader class."""
```
YoungVor: DocLoader

YoungVor (commented):
Finished my pass.  Its looking great!  The majority of my feedback is around the api - its worth examining any difference between the way session.read.parquet() and session.read.docs() work.. (can we do a similar list of path globs instead of the path + include, and do we want the same recursive/exclude ability in parquet/csv(), etc)

src/fenic/api/io/reader.py
```diff
@@ -227,3 +229,78 @@ def _read_file(
...
+            include: Optional[str] = None,
+            exclude: Optional[str] = None,
+            recursive: bool = False,
+            max_workers: Optional[int] = None,
```
rohitrastogi: I don't think we should expose this parameter to the user. The user should be oblivious to execution level decisions.

src/fenic/api/io/reader.py
rohitrastogi: I think we should do Union[JsonType, MarkdownType]

src/fenic/api/io/reader.py
rohitrastogi: I think a single docs() method parameterized by the datatype makes most sense - really what we're capturing in this method is that the directory is the source.

src/fenic/api/io/reader.py
```diff
@@ -227,3 +229,78 @@ def _read_file(
...
+            ValidationError: If any file does not have a `.md` or `.json` depending on the data_type.
+            UnsupportedFileTypeError: If the data_type is not supported.
+
+        Notes:
+            - Each row in the dataframe corresponds to a file in the path.
+            - The dataframe has the following columns:
+                - file_path: The path to the file.
+                - started_at: The time the file was started to be loaded.
+                - error: The error message if the file failed to be loaded.
+                - content: The content of the file.
```
rohitrastogi: Why include both content and casted content? If the user is specifying the datatype, I think they will expect the content to be in the type they specified.

src/fenic/_backends/local/utils/folder_loader.py
```diff
@@ -0,0 +1,350 @@
...
+            for i in range(0, len(files), batch_size):
+                batch = files[i:i + batch_size]
+                futures.append(executor.submit(DocFolderLoader._process_batch, batch, is_s3_path))
+
+            for future in as_completed(futures):
+                batch_df = future.result()
+                if resulting_df is None:
+                    resulting_df = batch_df
+                else:
+                    resulting_df.extend(batch_df)
```
rohitrastogi: Extend is pretty inefficient here. See the docs: https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.extend.html

We should be using vstack to avoid allocating and then calling rechunk() at the very end.

protos/logical_plan/v1/plans.proto
```diff
@@ -70,6 +71,16 @@ message TableSource {
...
 }
 
+message DocSource {
+  string path = 1;
+  DataType data_type = 2;
+  optional string include = 3;
+  optional string exclude = 4;
+  optional bool recursive = 5;
+  optional string path_hash = 6;
+  optional int32 max_workers = 7;
```
rohitrastogi: Shouldn't be part of the logical plan. This dictates HOW the plan is being executed, so if you want to store it, it should be in the physical plan.

src/fenic/_backends/local/physical_plan/source.py
rohitrastogi: lets keep this pr focused and consider tackling that later.

src/fenic/_backends/local/physical_plan/source.py
rohitrastogi: + 1 to accepting a list of path globs.

tests/markdown_test01.txt
rohitrastogi, line 1: does this need to be checked in?

src/fenic/_backends/local/physical_plan/source.py
german-typedef: +1.

src/fenic/_backends/local/utils/folder_loader.py
```diff
@@ -0,0 +1,350 @@
...
+    StringType,
+)
+
+logger = logging.getLogger(__name__)
+class DocFolderLoader:
+    """A class that encapsulates folder traversal and async/multi-threaded file processing.
+
+    This class provides functionality to:
+    1. Traverse folders with glob patterns and regex exclusions
+    2. Process files asynchronously
```
rohitrastogi: nit: this isn't really true - maybe in parallel

src/fenic/_backends/local/utils/folder_loader.py
rohitrastogi: if this is enumerating the files, maybe we should rename it to list_files instead of validate_path

src/fenic/_backends/utils/catalog_utils.py
```diff
@@ -146,20 +146,31 @@ def normalize_object_name(name: str) -> str:
...
                 logical_plan._table_name, session_state)
+        elif isinstance(logical_plan, DocSource):
+            clone_logical_plan = DocSource.from_session_state(
+                logical_plan._path,
+                logical_plan._data_type,
+                session_state,
+                logical_plan._include,
+                logical_plan._exclude,
+                logical_plan._recursive)
+            if clone_logical_plan._path_hash != logical_plan._path_hash:
```
rohitrastogi: After thinking about this more, I don't think we need to do any view invalidation for the docs sources since the schema of the dataframe isn't a function of the data anymore.

src/fenic/api/io/reader.py
```diff
@@ -227,3 +229,78 @@ def _read_file(
...
+
+        Raises:
+            ValidationError: If any file does not have a `.md` or `.json` depending on the data_type.
+            UnsupportedFileTypeError: If the data_type is not supported.
+
+        Notes:
+            - Each row in the dataframe corresponds to a file in the path.
+            - The dataframe has the following columns:
+                - file_path: The path to the file.
+                - started_at: The time the file was started to be loaded.
```
rohitrastogi: what value does this add at a per row level?

src/fenic/_backends/local/utils/folder_loader.py
rohitrastogi: agreed, we can just check this once and never have to check it again.

src/fenic/_backends/local/utils/folder_loader.py
german-typedef: Updated to do a path validation.

src/fenic/_backends/local/physical_plan/source.py
german-typedef: Updated to accept the path of globs.

src/fenic/_backends/local/utils/folder_loader.py
german-typedef: Added HF.

src/fenic/_backends/local/utils/folder_loader.py
german-typedef: Done.
The check is done in the API layer.

tests/_backends/local/io/test_folder_loader.py
german-typedef: Updated and renamed the file to reflect that is testing the doc loader.

src/fenic/api/io/reader.py
german-typedef: Removed.

src/fenic/api/io/reader.py
german-typedef: Thought originally of having some flexibility for the user, we can go with just the casted content. Updated accordingly.

src/fenic/_backends/local/utils/folder_loader.py
german-typedef: I went the iterator route, I think it works nicely with the ThreadPoolExecutor map functionality.

tests/markdown_test01.txt
german-typedef: Removed in favor of a simpler text that is available in the conftest.

src/fenic/_backends/local/utils/folder_loader.py
german-typedef: Updated.

src/fenic/_backends/utils/catalog_utils.py
german-typedef: Removed view invalidation.

src/fenic/api/io/reader.py
german-typedef: Removed.

src/fenic/_backends/local/utils/folder_loader.py
german-typedef: done.

protos/logical_plan/v1/plans.proto
german-typedef: removed unnecessary fields.

src/fenic/api/io/reader.py
```diff
@@ -260,3 +262,74 @@ def _read_file(
         from fenic.api.dataframe import DataFrame
 
         return DataFrame._from_logical_plan(logical_node, self._session_state)
+
+    def docs(
+            self,
+            paths: Union[str, list[str]],
+            data_type: Union[MarkdownType, JsonType],
+            exclude: Optional[str] = None,
+            recursive: bool = False,
```
YoungVor, line 271: Awesome, the api looks super clean!

Can we create a ticket for adding support for exclude and recursive in csv/parquet, so we don't forget?

german-typedef: done!
I've created a couple of tickets, one for the recursive / exclude for csv / parquet and one to implement hf & s3 for docs.

src/fenic/_backends/local/utils/doc_loader.py
```diff
@@ -0,0 +1,295 @@
...
+            paths: list[str],
+            valid_file_extension: str,
+            exclude_pattern: Optional[str],
+            recursive: bool = False,
+    ) -> List[str]:
+        """Enumerate files in a local fs folder based on include/exclude patterns."""
+        all_files = []
+        for path_pattern in paths:
+            split_path_pattern = path_pattern.split("*")
+            base_path = Path(split_path_pattern[0])
```
YoungVor, line 239: spoke in person but adding for paper-trail - we should handle path to an individual file case here and in the path validate code

src/fenic/api/io/reader.py
```diff
@@ -260,3 +262,74 @@ def _read_file(
...
+    ) -> DataFrame:
+        r"""Load a DataFrame from a folder of documents (markdown or json).
+
+        Args:
+            paths: Glob pattern (or list of glob patterns) to the folder(s) to load.
+            data_type: Data type that will be used to cast the content of the files.
+                       One of MarkdownType or JsonType.
+            exclude: A regex pattern to exclude files.
+                     If it is not provided no files will be excluded.
+            recursive: Whether to recursively load files from the folder.
```
YoungVor, line 281: Does it make sense to say that 'recursive' requires you to use ** glob pattern?

My first thought was that if they pass a path to a directory, you would automatically look for all matching files in that directory recursively

(So passing '/base/directory' with recursive==True would result in getting all matching files in  /base/directory/, /base/directory/foo, /base/directory/foo/bar, etc)

src/fenic/api/io/reader.py
```diff
@@ -260,3 +262,74 @@ def _read_file(
...
+                - casted_content: The content of the file casted to the data_type.
+
+        Example: Read a folder of markdown files
+            ```python
+            df = session.read.docs("data/docs", data_type=MarkdownType, include="*.md", recursive=True)
+            ```
+
+        Example: Read a folder of markdown files excluding some files.
+            ```python
+            df = session.read.docs("data/docs", data_type=MarkdownType, include="*.md", exclude=r"\.bak.md$", recursive=True)
```
YoungVor: out of date, you need the ** now for recursive?

tests/_backends/local/io/test_doc_loader.py
```diff
@@ -0,0 +1,113 @@
...
+    def test_enumerate_files_basic(self, temp_dir_with_test_files):
+        """Enumerate all the files in a folder."""
+        files = DocFolderLoader._enumerate_files(
+            self.get_globbed_path(temp_dir_with_test_files, "*.md"),
+            valid_file_extension=".md")
+        assert len(files) == 2
+        self._assert_file_in_results(files, temp_dir_with_test_files, "file1.md")
+        self._assert_file_in_results(files, temp_dir_with_test_files, "file2.md")
+
+    def test_enumerate_files_include_pattern_recursively(self, temp_dir_with_test_files):
```
YoungVor, line 29: good to include path to single file case

src/fenic/_backends/local/utils/doc_loader.py
german-typedef: Done.

german-typedef: Plus added a test case for that.

src/fenic/api/io/reader.py
german-typedef: I've updated the notes for docstring to reflect the glob behavior.

tests/_backends/local/io/test_doc_loader.py
german-typedef: Added.

src/fenic/api/io/reader.py
german-typedef: I've updated the examples to reflect the use of ** in conjunction with recursive = True.

src/fenic/core/_logical_plan/plans/source.py
```diff
@@ -180,3 +181,63 @@ def with_children(self, children: List[LogicalPlan], session_state: Optional[Bas
...
+    def with_children(self, children: List[LogicalPlan], session_state: Optional[BaseSessionState] = None) -> LogicalPlan:
+        if len(children) != 0:
+            raise InternalError(f"DocSource must have no children, got {len(children)}")
+        result = DocSource.from_schema(
+            self._paths, self._valid_file_extension, self._exclude, self._recursive, self._schema)
+        result.set_cache_info(self.cache_info)
+        return result
+
+    def _eq_specific(self, other: DocSource) -> bool:
+        return (self._paths == other._paths and
```
YoungVor, line 241: need to check exclude too

tests/_backends/local/io/test_doc_loader.py
```diff
@@ -0,0 +1,113 @@
+"""Tests for the FolderLoader class."""
```
YoungVor: nit: DocFilterLoader

src/fenic/core/_serde/proto/plans/source.py
```diff
@@ -125,3 +127,38 @@ def _deserialize_table_source(
...
         schema=schema,
     )
+
+# =============================================================================
+# DocSource
+# =============================================================================
+
+
+@_serialize_logical_plan_helper.register
+def _serialize_folder_source(
```
YoungVor: _serialize_doc_source

YoungVor: I'm trying to remember how the registration happens.. I guess it doesn't strictly matter what you call these functions because they're all dispatches to the serialize helper, so the unit testing still worked

uv.lock
```diff
@@ -11,9 +11,9 @@ resolution-markers = [
 name = "annotated-types"
 version = "0.7.0"
 source = { registry = "https://pypi.org/simple" }
-sdist = { url = "https://files.pythonhosted.org/packages/ee/67/531ea369ba64dcff5ec9c3402f9f51bf748cec26dde048a2f973a4eea7f5/annotated_types-0.7.0.tar.gz", hash = "sha256:aff07c09a53a08bc8cfccb9c85b05f1aa9a2a6f23728d790723543408344ce89", size = 16081 }
```
YoungVor, line 14: Probably dont need to update the uv.lock, since there's no pyproject change.  Unless you had to update a library version?

src/fenic/api/io/reader.py
```diff
@@ -260,3 +262,73 @@ def _read_file(
...
+
+        Returns:
+            DataFrame: A dataframe with all the files in the folder.
+
+        Raises:
+            ValidationError: If any file does not have a `.md` or `.json` depending on the data_type.
+            UnsupportedFileTypeError: If the data_type is not supported.
+
+        Notes:
+            - Each row in the dataframe corresponds to a file in the path.
```
YoungVor: nit: paths

src/fenic/api/io/reader.py
```diff
@@ -260,3 +262,73 @@ def _read_file(
...
+            - The dataframe has the following columns:
+                - file_path: The path to the file.
+                - error: The error message if the file failed to be loaded.
+                - casted_content: The content of the file casted to the data_type.
```
YoungVor: nit: could just call it 'content', its type is implicit in the schema.  Up to you tho

src/fenic/api/io/reader.py
YoungVor: thank you german!

src/fenic/_backends/local/utils/io_utils.py
```diff
@@ -137,3 +145,24 @@ def _configure_duckdb_conn(db_conn: duckdb.DuckDBPyConnection) -> duckdb.DuckDBP
     """Common configuration for duckdb connections."""
     db_conn.execute("set arrow_large_buffer_size=true;")
     return db_conn
+
+def get_path_scheme(path: str) -> PathScheme:
```
YoungVor, line 147: should this also check for any schema that isn't s3, hf, or file://, and throw?  We should catch that in the API, but if that fails we don't want to treat unsupported types as file://

src/fenic/_backends/local/utils/doc_loader.py
```diff
@@ -0,0 +1,299 @@
...
+            paths: list[str],
+            valid_file_extension: str,
+            exclude_pattern: Optional[str],
+            recursive: bool,
+    ) -> List[str]:
+        """Enumerate files in a HuggingFace dataset based on include/exclude patterns."""
+        raise NotImplementedError("HF file enumeration is not implemented yet.")
+
+    @staticmethod
+    def _enumerate_files_local_fs(
```
YoungVor, line 229: As a future change, it would probably make sense to use this enumerate function for csv and parquet, right?  We could simply call this first and pass the list to duckdb.  (definitely not needed now)

src/fenic/_backends/local/utils/doc_loader.py
```diff
@@ -0,0 +1,299 @@
...
+        # Filter out directories and apply exclude pattern
+        matching_files = []
+        for file_str in all_files:
+            file_path = Path(file_str)
+            if not file_path.is_file():
+                continue
+            if exclude_regex and exclude_regex.search(file_str):
+                continue
+            if not file_str.endswith(valid_file_extension):
+                raise ValueError(f"File is not supported: {file_str}, only files with the extension {valid_file_extension} are supported")
```
YoungVor: FileLoaderError?

also a nit:  its not that we only support {valid_file_extension}, its that this plan can only load files with extension {valid_file_extension}

src/fenic/_backends/local/utils/doc_loader.py
```diff
@@ -0,0 +1,299 @@
...
+            else:
+                raise ValueError(f"Path does not exist: {base_path}")
+
+        # Compile exclude regex if provided
+        exclude_regex = None
+        if exclude_pattern:
+            try:
+                exclude_regex = re.compile(exclude_pattern)
+            except re.error as e:
+                raise ValueError(f"Invalid exclude pattern regex: {exclude_pattern}") from e
```
YoungVor: I think this one should be ValidationError

src/fenic/_backends/local/utils/doc_loader.py
```diff
@@ -0,0 +1,299 @@
...
+            ]
+        )
+    
+    @staticmethod
+    def validate_paths(
+        paths: list[str],
+    ):
+        """Checks that the path is valid, and returns a list of files that match the include and exclude patterns."""
+        if not get_path_scheme(paths[0]) == PathScheme.LOCALFS:
+            raise ValueError("S3 paths are not supported yet.")
```
YoungVor: Could we have an UnsupportedError, similar to UnsupportedTypeError?

src/fenic/_backends/local/utils/doc_loader.py
```diff
@@ -0,0 +1,299 @@
...
+    @staticmethod
+    def _list_paths_local_fs(
+        paths: list[str],
+    ) -> list[Path]:
+        base_paths = []
+        for path_pattern in paths:
+            # Extract the base directory (everything before the first wildcard)
+            base_path = Path(path_pattern.split("*")[0]).parent
+            if not base_path.exists():
+                raise ValueError(f"path does not exist: {base_path}")
```
YoungVor: ValidationError

src/fenic/_backends/local/utils/doc_loader.py
```diff
@@ -0,0 +1,299 @@
...
+        Returns:
+            DataFrame: A dataframe containing the files in the folder.
+
+        Notes:
+            - The exclude pattern is a regex pattern to exclude files.
+            - The recursive flag indicates whether to recursively load files from the paths.
+            - The dataframe will be the union of all the files in the folders, plus additional columns for file metadata.
+        """
+        if not paths:
+            raise ValueError("No paths provided")
```
YoungVor: ValidationError

src/fenic/_backends/local/utils/doc_loader.py
YoungVor: also, should update the message - this could be hf:// too.  (Assuming get_path_scheme will raise a different error if the path is non file, s3, or hf)

YoungVor (commented):
A few more suggestions, but other than that, it looks great!  Thank you German, this is  a really powerful feature!

src/fenic/_backends/local/utils/doc_loader.py
YoungVor: actually, we dont need a new error, can just use NotImplementedError

src/fenic/_backends/local/utils/doc_loader.py
```diff
@@ -0,0 +1,299 @@
...
+
+        Args:
+            files: List of file paths to process
+            max_workers: Number of worker threads
+
+        Returns:
+            DataFrame: A dataframe containing the files in the folder.
+        """
+        with ThreadPoolExecutor(max_workers=max_workers) as executor:
+            results = executor.map(DocFolderLoader._process_single_file, files)
```
rohitrastogi: we can use as_completed, since order doesn't matter. This should reduce the amount of memory required - executor.map returns results in order so if data is out of ordered, results are buffered.

```
results_generator = (future.result() for future in as_completed(futures))
```
and you can pass this directly to pl.Dataframe

src/fenic/core/_logical_plan/plans/source.py
german-typedef: Added. Also found a thing that I needed in the deserialization to make sure that if "excluded" is not set, then it should be None.

tests/_backends/local/io/test_doc_loader.py
german-typedef: Updated.

src/fenic/core/_serde/proto/plans/source.py
german-typedef: Yep, it doesn't matter as they are dispatchers on the type. Makes sense to have them being correct.

german-typedef: Updated.

src/fenic/api/io/reader.py
german-typedef: Done.

src/fenic/_backends/local/utils/doc_loader.py
german-typedef: Updated the message.

src/fenic/_backends/local/utils/doc_loader.py
german-typedef: Makes sense, I updated a few of the errors.

src/fenic/_backends/local/utils/doc_loader.py
german-typedef: Went with NotImplementedError since that is what we have right now (we haven't implemented those).

src/fenic/_backends/local/utils/doc_loader.py
german-typedef: Done.

src/fenic/_backends/local/utils/doc_loader.py
german-typedef: Done.

src/fenic/_backends/local/utils/doc_loader.py
german-typedef: Updated accordingly.

src/fenic/api/io/reader.py
german-typedef: I prefer the explicitness fo the casted_content.

src/fenic/_backends/local/utils/doc_loader.py
german-typedef: discussed offline. Since this is done at the duckdb level, we might not want to use the same enumerate function. However, we should support "exclude". I've created a ticket for us to do so.

src/fenic/_backends/local/utils/io_utils.py
german-typedef: I think no scheme and "file" can be treated as LOCALFS, everything else should fail?
Let's talk about the change, in other places we explicitly check for s3/hf to call the right provider and then everything else we assume it is the file system.

src/fenic/api/io/reader.py
rohitrastogi: I don't think the user needs to know that the data was casted.

src/fenic/_backends/local/utils/doc_loader.py
```diff
@@ -0,0 +1,301 @@
...
+            ]
+        )
+    
+    @staticmethod
+    def validate_paths(
+        paths: list[str],
+    ):
+        """Checks that the path is valid, and returns a list of files that match the include and exclude patterns."""
+        if not get_path_scheme(paths[0]) == PathScheme.LOCALFS:
+            raise NotImplementedError("Neither s3 or hf paths are not supported yet.")
```
rohitrastogi: double negative: raise NotImplementedError("S3 and HF paths are not supported yet.")

src/fenic/_backends/local/utils/io_utils.py
```diff
@@ -137,3 +145,24 @@ def _configure_duckdb_conn(db_conn: duckdb.DuckDBPyConnection) -> duckdb.DuckDBP
...
+    scheme = urlparse(path).scheme
+    if scheme == "s3":
+        return PathScheme.S3
+    elif scheme == "hf":
+        return PathScheme.HF
+    else:
+        return PathScheme.LOCALFS
+
+
+def hash_list_of_strings(string_list: List[str]) -> str:
```
rohitrastogi: we don't need this anymore, right?

src/fenic/api/io/reader.py
```diff
@@ -260,3 +262,73 @@ def _read_file(
...
         return DataFrame._from_logical_plan(logical_node, self._session_state)
+
+    def docs(
+            self,
+            paths: Union[str, list[str]],
+            data_type: Union[MarkdownType, JsonType],
+            exclude: Optional[str] = None,
+            recursive: bool = False,
+    ) -> DataFrame:
+        r"""Load a DataFrame from a list of paths of documents (markdown or json).
```
rohitrastogi, line 273: Can we mention that each document is its own row. That isn't clear here.

rohitrastogi (approved):

src/fenic/_backends/local/utils/doc_loader.py
german-typedef: Updated.

src/fenic/_backends/local/utils/io_utils.py
german-typedef: Removed.

src/fenic/api/io/reader.py
german-typedef: Done.

YoungVor (approved):

src/fenic/api/io/reader.py
german-typedef: Got it. I've updated the name of the column to just content.

PR #150, opened 2025-08-19
feat: declarative tool creation

bcallender:
# Add support for declarative tools

This PR adds support for creating, listing, and dropping tools in the catalog. Tools are parameterized queries that can be executed with different parameter values.

Key features:

- Add `tool_param()` function to create parameter placeholders in queries
- Add catalog methods for tool management: `create_tool()`, `list_tools()`, `drop_tool()`
- Implement parameter binding with type validation
- Add system tables for storing tool metadata
- Add `fastmcp` integration for exposing tools as MCP endpoints
- Add `fenic-serve` script which bundles an executable into the package when installed via `pip` or built using `uv`. This allows users to spawn MCP servers in the command line by passing arguments to determine the tools the MCP server will serve to your agent. 

The implementation includes:

- Parameter binding with type validation and default values
- Pydantic model generation for tool parameters
- Support for allowed values constraints on parameters
- MCP server generation for exposing tools as API endpoints

PR #153, opened 2025-08-21
feat: automated mcp server generation

bcallender:
feat: automated mcp server generation

allo semantic query to find all primitive columns


bcallender:
* **#153** <a href="https://app.graphite.com/github/pr/typedef-ai/fenic/153?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/typedef-ai/fenic/153?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#150** <a href="https://app.graphite.com/github/pr/typedef-ai/fenic/150?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#163](https://github.com/typedef-ai/fenic/pull/163) <a href="https://app.graphite.com/github/pr/typedef-ai/fenic/163?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#168** <a href="https://app.graphite.com/github/pr/typedef-ai/fenic/168?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

yoni1887:
this is quite nifty!

src/fenic/api/tools.py
```diff
@@ -0,0 +1,762 @@
...
+    BooleanType,
+    DoubleType,
+    FloatType,
+    IntegerType,
+    StringType,
+)
+
+
+@dataclass
+class DatasetSpec:
```
rohitrastogi, line 54: why not repurpose tables or views for this to simplify user concepts and api surface?
we can add alter ddl statements that allow for addition of descriptions if they weren't added at table creation (same goes for column descriptions or column specific annotations - pks, fks)

src/fenic/api/tools.py
```diff
@@ -0,0 +1,762 @@
...
+
+    - name: short identifier used in SQL (e.g., {orders}) and displayed to the model
+    - description: brief description of dataset contents/semantics for tooldocs
+    - df: the DataFrame object
+    """
+    name: str
+    description: str
+    df: DataFrame
+
+class ReadFuncParams(BaseModel):
```
rohitrastogi, line 65: Is it possible to collapse this into the SQL tool? This is just a SQL query generator after all.

You could use sqlglot to verify that the query meets your constraints (here or elsewhere)

If you are going to take the query generator approach, why allow projections? It's just another vector for the model to mess up.

src/fenic/api/tools.py
```diff
@@ -0,0 +1,762 @@
...
+        result_limit=result_limit,
+    )
+
+"""
+Replace single search generator with two split generators:
+- auto_generate_search_summary_tool
+- auto_generate_search_content_tool
+"""
+
+def auto_generate_search_summary_tool(
```
rohitrastogi, line 154: what is the rationale for having a summary tool that searches all datasets in bulk instead of just having the model call it for each registered dataset as it thinks its needed?

Is the idea that this functions similarly to the grep tool claude code has access to which searches over all files in a dir?

src/fenic/api/tools.py
```diff
@@ -0,0 +1,762 @@
...
+    enriched_description = f"{tool_description}\n\n{schema_description}"
+    return DynamicTool(
+        name=tool_name,
+        description=enriched_description,
+        func=search_rows,
+        result_limit=result_limit,
+    )
+
+
+def auto_generate_schema_tool(
```
rohitrastogi, line 277: Why not simply re-purpose the existing describe_table catalog api and use it in batch?

src/fenic/api/tools.py
bcallender: Yeah, @cpard was proposing that we would have higher level tools that correspond to the `Grep` tool, with the ability to either return all of the results or just the filenames the results were found in.

src/fenic/api/tools.py
bcallender: that just returns a`Schema` , is there a difference between getting the `Schema` from the dataframe vs reading it from the catalog? we'd still have to do the same transformation to normalize the formatting.

bcallender: Unless you mean the pretty print capability? I was using that initially before switching to generating a more structured schema in this simplified JSON form

src/fenic/api/tools.py
bcallender: i went back and moved descriptions of tables and views to natively live in the catalog in #160, let me know if this is what you had in mind.

PR #157, opened 2025-08-25
feat: Add connector for reading huggingface scheme

YoungVor:
### TL;DR

Added support for HuggingFace datasets in Fenic's I/O operations (reads only).

### What changed?

- Added support for the `hf://` scheme to read and write data from HuggingFace datasets
- Implemented HuggingFace token authentication via environment variables (`HF_TOKEN`) or AWS Secrets Manager
- Updated DuckDB query building to handle HuggingFace credentials
- Added support for mixed source queries (e.g., reading from both S3 and HuggingFace)
- Updated documentation to reflect the new supported storage scheme
- Added comprehensive tests for public and private HuggingFace datasets

### How to test?

1. Unit tests 
 - Token reading and DuckDB query setup
 - Failing to authenticate Token
 - Reading CSV/Parquet, file lists, globs, hash revision
 - Mixing HuggingFace source with s3/local

2. Not covered by Unit Tests: Test reading from private HF datasets

-  Set your HuggingFace token:
   ```
   export HF_TOKEN=your_token_here
   ```

-  Read from a HuggingFace dataset:
   ```python
   import fenic as fn
   session = fn.Session.get_or_create()
   df = session.read.csv("hf://datasets/your-username/your-dataset/file.csv")
   ```

3. Not tested: using AWS secret manager for HuggingFace token

### Why make this change?

This change expands Fenic's data source capabilities to include HuggingFace datasets, which are widely used in the ML community. It allows users to seamlessly read from and write to HuggingFace datasets using the same familiar Fenic API, enabling more flexible data workflows that can combine data from multiple sources (local, S3, and now HuggingFace).

### DIFF
--- a/src/fenic/_backends/local/utils/io_utils.py
+++ b/src/fenic/_backends/local/utils/io_utils.py
@@ -1,6 +1,7 @@
+import logging
 import os
 from pathlib import Path
-from typing import List, Literal, Tuple
+from typing import List, Literal, Optional, Tuple
 from urllib.parse import urlparse
 
 import boto3
@@ -10,9 +11,10 @@ from botocore.credentials import ReadOnlyCredentials
 
 from fenic.core.error import ConfigurationError, ValidationError
 
+logger = logging.getLogger(__name__)
 
 def does_path_exist(path: str, s3_session:boto3.session.Session) -> bool:
-    """Check if a s3 or local path exists."""
+    """Check if a s3, hf, or local path exists."""
     scheme = urlparse(path).scheme
     if scheme == "s3":
         _, _ = _fetch_and_validate_s3_credentials(s3_session)
@@ -27,17 +29,32 @@ def does_path_exist(path: str, s3_session:boto3.session.Session) -> bool:
             if e.response["Error"]["Code"] == "404":
                 return False
             raise
+    elif scheme == "hf":
+        # For HF paths, we assume they exist if we have a token
+        hf_token = _fetch_hf_token(s3_session)
+        if not hf_token:
+            logger.info("HuggingFace token not found. Set HF_TOKEN environment variable or store in AWS Secrets Manager as 'huggingface-token'.")
+        return True  # We'll let DuckDB handle the actual path validation
     elif scheme == "file" or scheme == "":
         return os.path.exists(path)
     else:
-        raise ValidationError(f"Unsupported file type: {scheme} for path: {path}.  Please use s3 scheme ('s3://') or local scheme ('file://' or no prefix).")
+        raise ValidationError(f"Unsupported file type: {scheme} for path: {path}.  Please use s3 scheme ('s3://'), hf scheme ('hf://'), or local scheme ('file://' or no prefix).")
 
 
 def query_files(query: str, paths: List[str], s3_session: boto3.session.Session) -> pl.DataFrame:
-    """Execute a DuckDB query with s3 credentials."""
+    """Execute a DuckDB query with s3 and/or hf credentials."""
     duckdb_conn = _configure_duckdb_conn(duckdb.connect())
-    if any(urlparse(path).scheme == "s3" for path in paths):
-        query = _build_query_with_s3_creds(query, s3_session)
+    
+    has_s3_paths = any(urlparse(path).scheme == "s3" for path in paths)
+    has_hf_paths = any(urlparse(path).scheme == "hf" for path in paths)
+    
+    if has_s3_paths or has_hf_paths:
+        query = _build_query_with_httpfs_extensions(query)
+        if has_s3_paths:
+            query = _build_query_with_s3_creds(query, s3_session)
+        if has_hf_paths:
+            query = _build_query_with_hf_creds(query)
+    
     arrow_result = duckdb_conn.execute(query).arrow()
     return pl.from_arrow(arrow_result)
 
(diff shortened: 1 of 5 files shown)


YoungVor:
* **#157** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/157?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/157?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/fenic/_backends/local/utils/io_utils.py
```diff
@@ -76,6 +99,23 @@ def _fetch_and_validate_s3_credentials(s3_session:boto3.session.Session) -> Tupl
...
+def _fetch_hf_token(s3_session: boto3.session.Session) -> Optional[str]:
+    """Get HuggingFace token from environment variable or AWS Secrets Manager."""
+    # First check environment variable
+    hf_token = os.environ.get("HF_TOKEN")
+    if hf_token:
+        return hf_token
+    
+    # If not in environment, try AWS Secrets Manager
+    # we don't let user pass the region, but user can specify AWS_DEFAULT_REGION in the environment
+    try:
```
bcallender: hmm, not sure this will be helpful to users -- they would need to have the exact name we expect set as the name of the secret. i would suggest that this be taken out and eventually replaced with a system in which the user can provide a function  they define that returns a valid token (HF, S3 Access Token, JWT, etc), that we can use to authenticate to the source.

bcallender (commented):
this is going to be super helpful! just a quick first pass around the auth mechanism itself.

src/fenic/_backends/local/utils/io_utils.py
rohitrastogi: yep, i think we should just support env variable only for now

YoungVor: Makes sense.  Easily done!  I'll remove the support for using aws secrets manager

src/fenic/_backends/local/utils/io_utils.py
```diff
@@ -89,6 +115,40 @@ def _build_query_with_s3_creds(query: str, s3_session: boto3.session.Session) ->
     return query
 
 
+def _build_query_with_hf_creds(query: str, s3_session: boto3.session.Session) -> str:
+    """Helper method to add HuggingFace credentials to a DuckDB query."""
+    hf_token = _fetch_hf_token(s3_session)
+    if not hf_token:
+        logger.info("HuggingFace token not found. Set HF_TOKEN environment variable or store in AWS Secrets Manager as 'huggingface-token'.")
```
rohitrastogi: shouldn't this error if no hugging face api key is provided (ideally we actually do this during the planning phase with both s3/hf)?
also, this log line is stale since we arent using aws secrets manager

src/fenic/_backends/local/utils/io_utils.py
```diff
@@ -27,17 +29,32 @@ def does_path_exist(path: str, s3_session:boto3.session.Session) -> bool:
...
 def query_files(query: str, paths: List[str], s3_session: boto3.session.Session) -> pl.DataFrame:
-    """Execute a DuckDB query with s3 credentials."""
+    """Execute a DuckDB query with s3 and/or hf credentials."""
     duckdb_conn = _configure_duckdb_conn(duckdb.connect())
-    if any(urlparse(path).scheme == "s3" for path in paths):
+    
+    has_s3_paths = any(urlparse(path).scheme == "s3" for path in paths)
+    has_hf_paths = any(urlparse(path).scheme == "hf" for path in paths)
+    
+    if has_s3_paths and has_hf_paths:
```
rohitrastogi: instead of having three different distinct functions, I think this would be much cleaner using a builder pattern where you iteratively build the query string.

tests/_backends/local/io/test_reader.py
```diff
@@ -799,6 +804,151 @@ def test_read_queries_with_no_aws_credentials(local_session_config, temp_dir):
 
     session.stop()
 
+# =============================================================================
```
rohitrastogi, line 808: We don't need to check these in (it's just bloat). I think we can trust that DuckDB's hf connector works.

There's also no guarantee that these datasets won't change (which could break tests) unless you attach a revision/version number to the path.

src/fenic/_backends/local/utils/io_utils.py
YoungVor: Good point, this should be an error.

Correct, I believe this will happen during planning, on the infer schema step

src/fenic/_backends/local/utils/io_utils.py
YoungVor: I thought about that too, happy to change it

tests/_backends/local/io/test_reader.py
YoungVor: I did create the datasets myself though, so we could have some guarantee they wouldn't go anywhere.  It would be nice to catch in our tests if duckdb connector behavior ever changed, but I guess that's super unlikely, so I can remove these tests and the param

YoungVor: spoke in person, we decided to keep these tests as parameterized, and using datasets in the typedef org

src/fenic/_backends/local/utils/io_utils.py
rohitrastogi: it'd be nice to at least abstract out some of the shared bits

YoungVor: done!

PR #159, opened 2025-08-26
feat: Validate provider API keys during session configure

YoungVor:
### TL;DR

Added API key validation for model providers during session initialization to catch bad API keys early.

### What changed?

- Implemented `validate_api_key()` method for all model clients (OpenAI, Anthropic, Google, Cohere)
- Added asynchronous validation of provider API keys during session creation
- Created a validation mechanism that runs once per provider to avoid redundant validation
- Added tests to verify behavior with invalid API keys for each provider

### How to test?

1. Create a session with an invalid API key for each provider.  Make sure it fails the validate step with an authentication-related failure
2. Run the new tests, with and without anthropic, cohere, and google

### Why make this change?

Previously, API key validation only happened when a model was actually used, which could lead to delayed errors during execution. This change provides immediate feedback during session initialization, helping users identify configuration issues early.

### DIFF
--- a/src/fenic/_backends/local/model_registry.py
+++ b/src/fenic/_backends/local/model_registry.py
@@ -1,7 +1,9 @@
+import asyncio
 import logging
 from dataclasses import dataclass
-from typing import Optional
+from typing import Optional, Union
 
+from fenic._backends.local.async_utils import EventLoopManager
 from fenic._inference import (
     EmbeddingModel,
     LanguageModel,
@@ -56,12 +58,20 @@ class SessionModelRegistry:
 
         Args:
             config (ResolvedSemanticConfig): Configuration containing model settings and defaults.
+
+        Notes:
+            We only need to validate an API key once for each provider.  We prefer to validate a provider with
+            a language model over an embedding model, since the ping for an LLM is cheaper than
+            the embedding models (returning a single token calling list models, vs embedding a small string)
         """
+        validate_providers: set[str] = set()
         if config.language_models:
             language_model_config = config.language_models
             models: dict[str, LanguageModel] = {}
             for alias, model_config in language_model_config.model_configs.items():
-                models[alias] = self._initialize_language_model(model_config)
+                model = self._initialize_language_model(model_config)
+                models[alias] = model
+                validate_providers.add(model.client.model_provider)
             self.language_model_registry = LanguageModelRegistry(
                 models=models,
                 default_model=models[language_model_config.default_model],
@@ -71,11 +81,20 @@ class SessionModelRegistry:
             embedding_model_config = config.embedding_models
             models: dict[str, EmbeddingModel] = {}
             for alias, model_config in embedding_model_config.model_configs.items():
-                models[alias] = self._initialize_embedding_model(model_config)
+                model = self._initialize_embedding_model(model_config)
+                models[alias] = model
+                validate_providers.add(model.client.model_provider)
             self.embedding_model_registry = EmbeddingModelRegistry(
                 models=models,
                 default_model=models[embedding_model_config.default_model],
             )
+        if len(validate_providers) > 0:
+            with EventLoopManager().loop_context() as loop:
+                future = asyncio.run_coroutine_threadsafe(
+                    _validate_provider_api_keys(validate_providers),
+                    loop,
+                )
+                future.result()
 
     def get_language_model_metrics(self) -> LMMetrics:
         """Get aggregated metrics for all language models.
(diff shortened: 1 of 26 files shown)


YoungVor:
* **#159** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/159?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/159?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/fenic/_backends/local/model_registry.py
```diff
@@ -56,12 +57,21 @@ def __init__(self, config: ResolvedSemanticConfig):
...
         """
+        validate_provider_key_with_client: dict[str, Union[LanguageModel, EmbeddingModel]] = {}
         if config.language_models:
             language_model_config = config.language_models
             models: dict[str, LanguageModel] = {}
             for alias, model_config in language_model_config.model_configs.items():
-                models[alias] = self._initialize_language_model(model_config)
+                model = self._initialize_language_model(model_config)
+                models[alias] = model
+                if model.client.model_provider not in validate_provider_key_with_client:
```
rohitrastogi: you can simplify this by just overwriting in case the same model provider is used twice. all we care is that we capture one languagemodel/embedding model per model provider (doesn't matter which one)

src/fenic/_backends/local/model_registry.py
```diff
@@ -263,6 +285,7 @@ def _initialize_language_model(self, model_config: ResolvedModelConfig) -> Langu
                         AnthropicBatchCompletionsClient,
                     )
                 except ImportError as err:
+                    breakpoint()
```
rohitrastogi: please remove this

src/fenic/_backends/local/model_registry.py
```diff
@@ -71,11 +81,21 @@ def __init__(self, config: ResolvedSemanticConfig):
...
+                model = self._initialize_embedding_model(model_config)
+                models[alias] = model
+                if model.client.model_provider not in validate_provider_key_with_client:
+                    validate_provider_key_with_client[model.client.model_provider] = model
             self.embedding_model_registry = EmbeddingModelRegistry(
                 models=models,
                 default_model=models[embedding_model_config.default_model],
             )
+        if len(validate_provider_key_with_client) > 0:
+            from fenic._inference.model_client import ModelClientManager
```
rohitrastogi: here and elsewhere, please don't use local imports.
Also, this will need to be rebased to use the EventLoopManager

src/fenic/_inference/anthropic/anthropic_batch_chat_completions_client.py
```diff
@@ -400,3 +403,13 @@ def convert_messages(self, messages: LMRequestMessages) -> tuple[TextBlockParam,
         )
         message_params.append(MessageParam(content=[user_prompt], role="user"))
         return system_prompt, message_params
+
+    async def validate_api_key(self):
+        """Validate the Anthropic API key by making a minimal API call."""
+        _ = await self._client.completions.create(
```
rohitrastogi: anthropic has a list api too: https://github.com/anthropics/anthropic-sdk-python/blob/0e9e50124532ec6f110c36726f8267048111053d/src/anthropic/resources/models.py#L220

I'm sure gemini/cohere do as well.

rohitrastogi: You'll be duplicating this logic across embedding/language models, so I wonder if the validate/ping endpoint should be part of the model provider concept and not the model client.

you're deduping modelclients by model provider and then calling methods on the model client for deduped model providers - seems more direct to just call the method on the model provider itself

src/fenic/_backends/local/model_registry.py
YoungVor: I like that, definitely cleaner

src/fenic/_backends/local/model_registry.py
YoungVor: good point, will rebase and use the EventLoopManager

src/fenic/_inference/anthropic/anthropic_batch_chat_completions_client.py
YoungVor: hmm, that's a good point.  It would be nice not to tie it to  a specific model, (which may vary in cost), and its nice not to actually create an embedding or generate tokens

YoungVor:
Moved to here: https://github.com/typedef-ai/fenic/pull/164

PR #160, opened 2025-08-27
feat: allow users to add descriptions to views and tables

bcallender:
### TL;DR

Added support for table and view descriptions in the catalog system.

### What changed?

- Implemented description storage and retrieval in the local catalog system
- Added new methods to set and get descriptions for tables and views
- Created `DatasetMetadata` class to return schema and description together
- Added system table schema updates to store descriptions in the catalog

### How to test?

```
# Create a table with a description
df = session.create_dataframe({"a": [1, 2, 3]})
df.write.save_as_table("my_table")
session.catalog.set_table_description("my_table", "This is my table")

# Create a view with a description
df.write.save_as_view("my_view", description="This is my view")

# Get metadata including description
table_meta = session.catalog.get_table_metadata("my_table")
print(table_meta.description)  # "This is my table"

view_meta = session.catalog.get_view_metadata("my_view")
print(view_meta.description)  # "This is my view"

# Update a description
session.catalog.set_table_description("my_table", "Updated description")
```

### Why make this change?

This change enhances the catalog system by allowing users to add human-readable descriptions to tables and views. These descriptions provide context and documentation for data assets, which will be used in an upcoming PR to generate MCP tools against saved tables.

### DIFF
--- a/src/fenic/_backends/local/catalog.py
+++ b/src/fenic/_backends/local/catalog.py
@@ -296,8 +296,9 @@ class LocalCatalog(BaseCatalog):
                 self.get_current_catalog(),
                 self.get_current_database())
             _verify_table_catalog(table_identifier)
+            cursor = self.db_conn.cursor()
             try:
-                self.system_tables.set_table_description(table_identifier.db, table_identifier.table, description)
+                self.system_tables.set_table_description(cursor, table_identifier.db, table_identifier.table, description)
             except Exception as e:
                 raise CatalogError(
                     f"Failed to set description for table: `{table_identifier.db}.{table_identifier.table}`"
@@ -309,8 +310,9 @@ class LocalCatalog(BaseCatalog):
                 self.get_current_catalog(),
                 self.get_current_database())
             _verify_table_catalog(view_identifier)
+            cursor = self.db_conn.cursor()
             try:
-                self.system_tables.set_view_description(view_identifier.db, view_identifier.table, description)
+                self.system_tables.set_view_description(cursor, view_identifier.db, view_identifier.table, description)
             except Exception as e:
                 raise CatalogError(
                     f"Failed to set description for view: `{view_identifier.db}.{view_identifier.table}`"
@@ -323,7 +325,8 @@ class LocalCatalog(BaseCatalog):
                 self.get_current_database())
             _verify_table_catalog(table_identifier)
             schema = self.describe_table(table_name)
-            description = self.system_tables.get_table_description(table_identifier.db, table_identifier.table)
+            cursor = self.db_conn.cursor()
+            description = self.system_tables.get_table_description(cursor, table_identifier.db, table_identifier.table)
             return TableMetadata(schema=schema, description=description)
 
     def get_view_metadata(self, view_name: str) -> ViewMetadata:
@@ -334,7 +337,8 @@ class LocalCatalog(BaseCatalog):
             _verify_table_catalog(view_identifier)
             logical_plan = self.describe_view(view_name)
             schema = logical_plan.schema()
-            description = self.system_tables.get_view_description(view_identifier.db, view_identifier.table)
+            cursor = self.db_conn.cursor()
+            description = self.system_tables.get_view_description(cursor, view_identifier.db, view_identifier.table)
             return ViewMetadata(schema=schema, description=description)
 
     def describe_table(self, table_name: str) -> Schema:
--- a/src/fenic/_backends/local/catalog.py
+++ b/src/fenic/_backends/local/catalog.py
@@ -15,7 +15,7 @@ from fenic._backends.utils.catalog_utils import (
     TableIdentifier,
     compare_object_names,
 )
-from fenic.core._interfaces.catalog import BaseCatalog, TableMetadata, ViewMetadata
+from fenic.core._interfaces.catalog import BaseCatalog
 from fenic.core._logical_plan.plans.base import LogicalPlan
 from fenic.core._utils.misc import generate_unique_arrow_view_name
 from fenic.core._utils.schema import convert_custom_schema_to_polars_schema
@@ -29,6 +29,7 @@ from fenic.core.error import (
 )
 from fenic.core.metrics import QueryMetrics
 from fenic.core.types import (
+    DatasetMetadata,
     Schema,
 )
 
@@ -304,20 +305,6 @@ class LocalCatalog(BaseCatalog):
                     f"Failed to set description for table: `{table_identifier.db}.{table_identifier.table}`"
                 ) from e
 
-    def set_view_description(self, view_name: str, description: Optional[str]) -> None:
-        with self.lock:
-            view_identifier = TableIdentifier.from_string(view_name).enrich(
-                self.get_current_catalog(),
-                self.get_current_database())
-            _verify_table_catalog(view_identifier)
-            cursor = self.db_conn.cursor()
-            try:
-                self.system_tables.set_view_description(cursor, view_identifier.db, view_identifier.table, description)
-            except Exception as e:
-                raise CatalogError(
-                    f"Failed to set description for view: `{view_identifier.db}.{view_identifier.table}`"
-                ) from e
-
     def get_table_description(self, table_name: str) -> Optional[str]:
         with self.lock:
             table_identifier = TableIdentifier.from_string(table_name).enrich(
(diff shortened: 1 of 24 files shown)


bcallender:
* **#160** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/160?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/160?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/fenic/api/catalog.py
```diff
@@ -510,6 +510,20 @@ def list_views(self) -> List[str]:
...
+
+    def get_view_metadata(self, view_name: str) -> ViewMetadata:
+        """Return schema and description for a view in one call."""
+        return self.catalog.get_view_metadata(view_name)
+
+    # Description mutators
+    def set_table_description(self, table_name: str, description: str | None) -> None:
+        """Set or clear the description for a table."""
+        self.catalog.set_table_description(table_name, description)
+
```
rohitrastogi, line 616: do we need set_view_description?

src/fenic/core/_interfaces/catalog.py
```diff
@@ -1,14 +1,27 @@
...
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, List, Optional
 
 from fenic.core.types import Schema
 
 if TYPE_CHECKING:
     from fenic.core._logical_plan.plans.base import LogicalPlan
 
 
+@dataclass(frozen=True)
```
rohitrastogi: can we add docstrings for these?

src/fenic/api/catalog.py
```diff
@@ -510,6 +510,20 @@ def list_views(self) -> List[str]:
         """
         return self.catalog.list_views()
 
+    # Metadata convenience
+    def get_table_metadata(self, table_name: str) -> TableMetadata:
```
rohitrastogi: can we just change the api of describe table to return a tablemetadata instead of adding new methods?

src/fenic/_backends/local/system_table_client.py
```diff
@@ -385,10 +472,15 @@ def _initialize_views_metadata(self, cursor: duckdb.DuckDBPyConnection) -> None:
                     view_name TEXT NOT NULL,
                     view_blob TEXT NOT NULL,
                     creation_time TIMESTAMP NOT NULL,
+                    description TEXT,
                     PRIMARY KEY (database_name, view_name)
                 );
             """
             )
+            # Ensure description column exists for dev-time schema evolution
```
rohitrastogi: is this to migrate users with existing views?
not sure we need to worry about that for now.

src/fenic/api/io/writer.py
```diff
@@ -36,6 +36,7 @@ def save_as_table(
         self,
         table_name: str,
         mode: Literal["error", "append", "overwrite", "ignore"] = "error",
+        description: str | None = None,
```
rohitrastogi: Wondering if we should separate the concerns of writing data from writing metadata to keep things explicit. I know this API does the write thing when the user has a table and the description is None (doesn't blindly overwrite), but I feel that that behavior can be confusing.

src/fenic/api/catalog.py
bcallender: ~yeah, i only didn't do that because I couldn't also change `describe_view` as it returns the `LogicalPlan` instead of the schema, i can change it as long as we're okay with some slight asymmetry.~

I misread this, thought it was in `system_table_client` where we do have a `describe_view` but the thought was the same

bcallender: or i can technically just remove the concept of `ViewMetadata`, we're not using it yet

src/fenic/api/io/writer.py
bcallender: Yeah I also struggled with that. I figured that the convenience outweighed the confusion factor (and added some safeguards like you pointed out), but I've been staring at this for a week and a half, so I'm less able to really tell if something is confusing for someone looking at it for the first time. 

I can remove description from here (it would mean a lot less code, we wouldn't need to plumb it through `TableSink` and only allow the user to set a description using `set_table_description`?

src/fenic/_backends/local/system_table_client.py
bcallender: this was mostly for me, i had a bunch of different test apps in varying states of their original schema. can remove.

src/fenic/api/catalog.py
bcallender: There isn't a technical reason why you couldn't autogenerate tools for views in addition to table, so I figured tables and views should have the same underlying metadata available.

src/fenic/api/catalog.py
bcallender: done

src/fenic/api/io/writer.py
bcallender: done, as discussed over slack, `save_as_table` no longer accepts a description

src/fenic/core/types/schema.py
```diff
@@ -120,3 +121,15 @@ def column_names(self) -> List[str]:
...
+@dataclass(frozen=True, config=ConfigDict(arbitrary_types_allowed=True))
+class DatasetMetadata:
+    """Metadata for a dataset (table or view).
+
+    Attributes:
+        schema: The schema of the dataset.
+        description: The natural language description of the dataset's contents.
+    """
+    schema: Schema
+    description: Optional[str]
```
YoungVor, line 135: thinking about the future and how we might eventually want a description for every field..

do we eventually want to bake 'description' into the ColumnField itself?  In which case, we could go the whole way and just bake the table description into the Schema class, eliminating the need for a separate Metadata class.  Though that gets tricky in schema equivalency checks, in some cases you may want to use description in equals checks and not in others.

I guess the other option is to have a class structure that can map to a nested schema, and include that in the metadata object?

src/fenic/api/catalog.py
```diff
@@ -510,6 +531,45 @@ def list_views(self) -> List[str]:
...
+    def set_view_description(self, view_name: str, description: str) -> None:
+        """Set the description for a view.
+
+        Args:
+            view_name (str): Fully qualified or relative view name to set the description for.
+            description (str): The description to set for the view.
+
+        Raises:
+            ViewNotFoundError: If the view doesn't exist.
+            ValidationError: If the description is empty.
```
YoungVor, line 608: what if they want to unset the table description?

src/fenic/_backends/local/system_table_client.py
```diff
@@ -238,6 +300,79 @@ def get_view(
...
+                FROM "{SYSTEM_SCHEMA_NAME}"."{VIEWS_METADATA_TABLE}"
+                WHERE database_name = ? AND view_name = ?
+            """,
+                (database_name, view_name),
+            ).fetchone()
+            # trunk-ignore-end(bandit/B608)
+            if result is None:
+                return None
+            view_blob = base64.b64decode(result[1])
+            schema = LogicalPlanSerde.deserialize(view_blob).schema()
```
YoungVor, line 335: Ah, before there was no way to query the schema of a table/df before, without materializing the df?  Nice addition :)

src/fenic/_backends/local/catalog.py
```diff
@@ -283,22 +284,53 @@ def list_views(self) -> List[str]:
...
-            if maybe_schema is None:
+            if maybe_table_metadata is None:
                 raise TableNotFoundError(table_identifier.table, table_identifier.db)
-            return maybe_schema
+            return maybe_table_metadata
 
-    def describe_view(self, view_name: str) -> LogicalPlan:
-        """Get the schema of the specified view."""
+    def get_view_plan(self, view_name: str) -> LogicalPlan:
+        """Get the schema and description of the specified view."""
```
YoungVor: I think the comments are out of sync - this should go under describe_view, and get_view_plan needs a new comment

src/fenic/_backends/local/catalog.py
```diff
@@ -283,22 +284,53 @@ def list_views(self) -> List[str]:
...
+            cursor = self.db_conn.cursor()
+            try:
+                self.system_tables.set_table_description(cursor, table_identifier.db, table_identifier.table, description)
+            except Exception as e:
+                raise CatalogError(
+                    f"Failed to set description for table: `{table_identifier.db}.{table_identifier.table}`"
+                ) from e
+
+    def get_table_description(self, table_name: str) -> Optional[str]:
+        with self.lock:
```
YoungVor, line 313: nit: good to include a comment for consistency

YoungVor (commented):
did a pass and looks good, but Rohit should do a pass too

src/fenic/_backends/local/system_table_client.py
rohitrastogi: describe table did this

bcallender: There wasn't an easy way for `views`, tables always had `describe_table` to get the schema.

src/fenic/core/types/schema.py
rohitrastogi: If we do this, we're going to need a separate semantic view/semantic layer concept which will need to be designed from scratch. 

Here's how snowflake does it: https://docs.snowflake.com/en/user-guide/views-semantic/overview

bcallender: Yeah, I thought about baking the description into the schema but like you pointed out it makes schema equivalence quite difficult -- I think it makes sense for Schema to be a purely structural table description of the columns and datatypes, while `semantic` metadata is stored separately, as part of making a view or something.

src/fenic/api/catalog.py
bcallender: made it optional

src/fenic/_backends/local/catalog.py
bcallender: fixed

src/fenic/_backends/local/catalog.py
bcallender: added

src/fenic/_backends/local/system_table_client.py
```diff
@@ -105,16 +118,59 @@ def get_schema(self, cursor: duckdb.DuckDBPyConnection, database_name: str, tabl
...
+        """Retrieve a table's description, if present."""
+        try:
+            # trunk-ignore-begin(bandit/B608): Query built from constants; parameters are bound.
+            result = cursor.execute(
+                f"""
+                SELECT description
+                FROM "{SYSTEM_SCHEMA_NAME}"."{SCHEMA_METADATA_TABLE}"
+                WHERE database_name = ? AND table_name = ?
+            """,
+                (normalize_object_name(database_name), normalize_object_name(table_name)),
```
rohitrastogi: I don't think you need to call normalize since the caller normalized it already.

bcallender: removed

PR #161, opened 2025-08-27
feat: local metrics table

YoungVor:
### TL;DR

Added a metrics tracking system that automatically records query execution statistics in a read-only `metrics` table.

### What changed?

- Created a new read-only `metrics` table that tracks query execution statistics
- Added concept of read-only database in the catalog: `fenic_system`
- Added ability to System_Table_client to insert metrics and aggregate from last session
- Added ability for session to manage metrics table through catalog api
- Enhanced `QueryMetrics` with session tracking and timestamp information
- Added session usage summary that displays at the end of a session

### Example Output on session.close():

Session Usage Summary:
App Name: test_app
Session ID: 268f2b29-096a-4a90-bbd2-abb02e9a382b
Total queries executed: 4
Total execution time: 15.22ms
Total rows processed: 10
Total language model cost: $0.000000
Total embedding model cost: $0.000000
Total cost: $0.000000

### How to test?

1. Run queries and verify metrics are being collected:
2. Verify read-only protection works:
3. Test session usage summary by stopping a session:

### Why make this change?

This change enables automatic tracking of query execution metrics, providing users with insights into their session's performance and cost. The metrics table stores information about execution time, row counts, token usage, and estimated costs, which helps users understand and optimize their queries. The session summary provides a convenient overview of resource usage when a session ends.

### DIFF
--- a/src/fenic/_backends/local/catalog.py
+++ b/src/fenic/_backends/local/catalog.py
@@ -1,11 +1,12 @@
 import logging
 import threading
-from typing import List
+from typing import Dict, List
 
 import duckdb
 import polars as pl
 
 from fenic._backends.local.system_table_client import (
+    READ_ONLY_SYSTEM_SCHEMA_NAME,
     SYSTEM_SCHEMA_NAME,
     SystemTableClient,
 )
@@ -26,6 +27,7 @@ from fenic.core.error import (
     TableAlreadyExistsError,
     TableNotFoundError,
 )
+from fenic.core.metrics import QueryMetrics
 from fenic.core.types import (
     Schema,
 )
@@ -56,7 +58,7 @@ class DuckDBTransaction:
         return self.connection
 
     def __exit__(self, exc_type, exc_val, exc_tb):
-        """Commit transaction if no exceptions, otherwise rollback and log error."""
+        """Commit transaction if no exceptions, otherwise rollback and l error."""
         if exc_type is None:
             # No exception occurred, commit the transaction
             logger.debug("Committing DuckDB transaction")
@@ -168,6 +170,10 @@ class LocalCatalog(BaseCatalog):
         with self.lock:
             db_identifier = DBIdentifier.from_string(database_name).enrich(self.get_current_catalog())
             _verify_db_catalog(db_identifier)
+            if db_identifier.is_db_name_equal(READ_ONLY_SYSTEM_SCHEMA_NAME):
+                raise CatalogError(
+                    f"Cannot drop read-only system database '{READ_ONLY_SYSTEM_SCHEMA_NAME}'"
+                )
             if db_identifier.is_db_name_equal(self.get_current_database()):
                 raise CatalogError(
                     f"Cannot drop the current database '{database_name}'. Switch to another database first."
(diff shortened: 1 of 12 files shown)


YoungVor:
* **#161** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/161?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/161?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/fenic/_backends/local/catalog.py
```diff
@@ -292,6 +304,33 @@ def describe_table(self, table_name: str) -> Schema:
                 self.get_current_database())
             _verify_table_catalog(table_identifier)
 
+            # Special handling for metrics table
```
rohitrastogi: what do you think of treating user-facing system tables exactly like user tables (even writing their schemas to the schema system table). This way there's no difference in the read codepath between the table types.

Basically the catalog would store three types of tables:
- user tables
- user facing system tables (metrics, traces -planned)
- hidden system tables (schemas, views, tools-planned, prompts-planned)

user facing system tables are exposed just like user tables.

YoungVor: yes, I think that makes sense, and it should be easy to take a SystemsTableClient in the metrics client file

src/fenic/_backends/local/catalog.py
```diff
@@ -39,10 +39,22 @@
...
+    "metrics",
+]
 DEFAULT_CATALOG_NAME = "typedef_default"
 DEFAULT_DATABASE_NAME = "typedef_default"
 
 
+def _is_read_only_table(table_identifier: TableIdentifier) -> bool:
+    """Check if a table is read-only and should be protected from modification."""
+    # Only protect read-only tables in the default database
+    if not table_identifier.is_db_name_equal(DEFAULT_DATABASE_NAME):
```
rohitrastogi: i dont think the user facing system tables should be in the default database as they will collide (metrics is not an unreasonable table name)

we should namespace them in a separate db (possibly 'fenic_system'). "hidden" fenic tables can remain in "__fenic_system"

src/fenic/_backends/local/catalog.py
```diff
@@ -434,13 +450,18 @@ def write_df_to_table(self, df: pl.DataFrame, table_name: str, schema: Schema):
                 self.get_current_catalog(),
                 self.get_current_database())
             _verify_table_catalog(table_identifier)
+
+            if _is_read_only_table(table_identifier):
+                raise CatalogError(
+                    f"Cannot write to read-only table '{table_identifier.table}'"
```
rohitrastogi: after we move to a different db, it would be useful to log the entire table identifier

src/fenic/_backends/local/metrics_client.py
```diff
@@ -0,0 +1,212 @@
...
+    IntegerType,
+    StringType,
+)
+
+logger = logging.getLogger(__name__)
+
+METRICS_TABLE_NAME = "metrics"
+
+
+class MetricsClient:
```
rohitrastogi: Why do we need an entirely separate MetricsClient? Isn't this a system table - why not put this in the system table client?

src/fenic/_backends/local/catalog.py
YoungVor: that makes sense, I'll add the system view

src/fenic/_backends/local/catalog.py
YoungVor: agreed

src/fenic/_backends/local/metrics_client.py
YoungVor: makes sense.  moved to Systems Table Client

src/fenic/_backends/local/session_state.py
```diff
@@ -49,6 +52,8 @@ def __init__(
         self._model_registry = self._configure_models(config.semantic)
         self.intermediate_df_client = TempDFDBClient(self.app_name)
         self.s3_session = boto3.Session()
+        self.system_table_client = SystemTableClient(self.duckdb_conn)
```
rohitrastogi: I prefer that the system table client is an implementation detail of the catalog implementation - I think all crud on catalog related objects should go through the catalog implementation (which can of course forward requests on to the system table client)

YoungVor: makes sense!  Added pass-through api to catalog and removed system_table_client from session

PR #162, opened 2025-08-28
fix: use rust regex to validate regular expression used for rlike, ilike, like

german-typedef:
### TL;DR

Use rust's regex crate to validate a regular expression passed to rlike, ilike and like. The validation provides a consistent experience with the execution layer.

### What changed?

- Modified `RLikeExpr`, `LikeExpr`, and `ILikeExpr` in the protocol buffer to accept a `LogicalExpr` for the pattern instead of a string
- Added regex validation functionality through a new Rust module
- Updated the column API methods (`rlike`, `like`, `ilike`) to accept either string literals or column expressions
- Modified the serialization/deserialization logic to handle the new pattern type
- Added validation for regex patterns to catch invalid expressions early
- Added the test cases that were passing with python regex validation but were failing during execution


german-typedef:
* **#162** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/162?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/162?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

protos/logical_plan/v1/expressions.proto
```diff
@@ -327,17 +327,17 @@ message ContainsAnyExpr {
 
 message RLikeExpr {
   LogicalExpr expr = 1;
-  string pattern = 2;
```
bcallender, line 330: We haven't released proto serde yet so I don't think we need to worry about breaking the API until then, it can stay 2.

If we did need to do this after release though, it would be best practice to leave `reserved 2` after deleting the field so the tag doesn't get accidentally reused

src/fenic/api/column.py
```diff
@@ -485,14 +484,17 @@ def rlike(self, other: str) -> Column:
...
+        return Column._from_logical_expr(RLikeExpr(self._logical_expr, other_expr))
 
-    def like(self, other: str) -> Column:
+    def like(self, other: Union[str, Column]) -> Column:
         r"""Check if the column matches a SQL LIKE pattern.
 
         This method creates a boolean expression that checks if each value in the column
-        matches the specified SQL LIKE pattern. The pattern must be a literal string
-        and cannot be a column expression.
+        matches the specified SQL LIKE pattern. The pattern must be a string or a LiteralExpr with StringType.
```
bcallender: LiteralExpr is an internal component, to the user we should reference the API `.lit()` function

bcallender: But really, the idea is that they can fill them in with `.tool_param()`

src/fenic/api/column.py
```diff
@@ -460,15 +460,14 @@ def ends_with(self, other: Union[str, Column]) -> Column:
...
         r"""Check if the column matches a regular expression pattern.
 
         This method creates a boolean expression that checks if each value in the column
-        matches the specified regular expression pattern. The pattern must be a literal string
-        and cannot be a column expression.
+        matches the specified regular expression pattern.
 
         Args:
-            other (str): The regular expression pattern to match against
+            other (Union[str, Column]): The regular expression pattern to match against (can be a string or a column expression)
```
bcallender: Wait, can it be an arbitrary column expression or only a `LiteralExpr`? Because later on it seems like it is required to only be one

src/fenic/core/_logical_plan/expressions/text.py
```diff
@@ -408,22 +412,22 @@ class RLikeExpr(ValidatedSignature, LogicalExpr):
         pattern: The regular expression pattern to match against
 
     Raises:
-        TypeError: If the input expression is not a string column
+        TypeError: If the pattern is not a LiteralExpr with StringType
```
bcallender: Yeah this is why I was confused, the docs seem contradictory. I think `rlike` can technically work with any expr that resolves to `StringType` but `ilike` _requires_ a LiteralExpr because we need to convert it to regex for Polars `contains` to work

src/fenic/core/_logical_plan/expressions/text.py
```diff
@@ -434,7 +438,7 @@ def children(self) -> List[LogicalExpr]:
         return [self.expr]
 
     def __str__(self) -> str:
-        return f"{self.function_name}({self.expr}, {self.pattern})"
```
bcallender: This assumes it's a `LiteralExpr`, no? It can just be `self.pattern` and let the expression handle itself

src/fenic/core/_logical_plan/expressions/text.py
```diff
@@ -457,16 +461,18 @@ class LikeExpr(ValidatedSignature, LogicalExpr):
...
         self.raw_pattern = pattern
-        self.pattern = self._convert_to_regex(pattern)
+        self.pattern = pattern
 
-        # Validate the converted pattern
-        try:
-            re.compile(self.pattern)
-        except Exception as e:
-            raise ValidationError(f"Invalid LIKE pattern: {self.raw_pattern}") from e
+        if isinstance(pattern, LiteralExpr) and pattern.data_type == StringType:
```
bcallender, line 466: `ilike` only works with a  `LiteralExpr`, so if anything this one should throw an error if it's not.

bcallender (commented):
quick first pass

protos/logical_plan/v1/expressions.proto
german-typedef: Good point.
Left them at 2.

src/fenic/api/column.py
german-typedef: I've updated the comment.

src/fenic/core/_logical_plan/expressions/text.py
german-typedef: Modified the doc strings for the expressions changed:
rlike can be an arbitrary expression, if it is a literal, then we'll validate the regular expression.
For like and ilike it must be a literal expression with a string type.

src/fenic/core/_logical_plan/expressions/text.py
german-typedef: done.

src/fenic/core/_logical_plan/expressions/text.py
german-typedef: Updated the doc strings for like and ilike as both require a literal expression.

src/fenic/core/_logical_plan/expressions/text.py
german-typedef: done.

src/fenic/core/_logical_plan/expressions/text.py
german-typedef: updated.

src/fenic/core/_logical_plan/expressions/text.py
rohitrastogi: We can do the ilike/like conversion itself as polars expressions so we can just use expr.contains(strict=False) if we don't get string/LiteralExpr patterns.

german-typedef: Updated.

src/fenic/_backends/local/transpiler/expr_converter.py
```diff
@@ -1253,3 +1263,23 @@ def adapted_udf(row):
...
+        pattern_expr
+        # Escape each meta character by prefixing with a backslash
+        .str.replace_all(meta_class, r"\\\1")
+        # Translate SQL wildcards to regex
+        .str.replace_all("%", ".*", literal=True)
+        .str.replace_all("_", ".",  literal=True)
+    )
+
+    if case_insensitive:
+        return regex_expr.map_elements(lambda s: f"(?i)^{s}$")
```
bcallender: i'm not sure that the `^` and `$` are needed to make it case-insensitive. either way, it would probably be faster to do this as `regex_expr = (pl.lit("(?i)") + regex_expr)`

src/fenic/api/column.py
```diff
@@ -517,14 +521,18 @@ def like(self, other: str) -> Column:
...
 
-    def ilike(self, other: str) -> Column:
+    def ilike(self, other: Union[str, Column]) -> Column:
         r"""Check if the column matches a SQL LIKE pattern (case-insensitive).
 
         This method creates a boolean expression that checks if each value in the column
-        matches the specified SQL LIKE pattern, ignoring case. The pattern must be a literal string
-        and cannot be a column expression.
+        matches the specified SQL LIKE pattern, ignoring case. The pattern must be either a string
+        or a call to .lit() with a string literal.
```
bcallender: this and `like` can just be the same as `rlike` now that we can create the regex at execution time.

src/fenic/_backends/local/transpiler/expr_converter.py
```diff
@@ -690,11 +690,21 @@ def _convert_contains_any_expr(self, logical: ContainsAnyExpr) -> pl.Expr:
 
 
     # Group like/regex expressions together
-    def _handle_regex_like_expr(self, logical, pattern_field="pattern", literal: bool = False) -> pl.Expr:
+    def _handle_regex_like_expr(self, logical, literal: bool = False) -> pl.Expr:
```
bcallender: nit: if it's `regex_like` then `literal` will always be `False` i think, so it can be removed from here

bcallender (commented):
just a few nits -- wanted to get this pass in for the morning

src/fenic/_backends/local/transpiler/expr_converter.py
german-typedef: Thanks! not being an expert, I went with the Claude suggestion, but I like the simpler version.
Updated.

src/fenic/api/column.py
german-typedef: Agreed. I've updated the comment.

src/fenic/_backends/local/transpiler/expr_converter.py
german-typedef: Removed. Yep not needed.

german-typedef:
### Merge activity

* **Aug 29, 8:54 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/typedef-ai/fenic/162).
* **Aug 29, 8:55 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/typedef-ai/fenic/162) rebased this pull request as part of a merge.
* **Aug 29, 9:00 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/typedef-ai/fenic/162) couldn't merge this PR because it failed for an unknown reason.
* **Aug 29, 9:26 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/typedef-ai/fenic/162).
* **Aug 29, 9:30 PM UTC**: [Graphite](https://app.graphite.dev/github/pr/typedef-ai/fenic/162) couldn't merge this PR because it failed for an unknown reason.

rohitrastogi (approved):

PR #164, opened 2025-08-28
feat: validate provider keys

YoungVor:
## TL;DR

Added API key validation for model providers during session initialization to catch bad API keys early.

## What changed?

* Added common provider functionality with a ModelProviderClass structure, for the ModelClient class
* Implemented cheap validate_api_key for each provider
* Added asynchronous validation of provider API keys during session creation
* Added tests to verify behavior with invalid API keys for each provider
* Switched Cohere support to use async instead of sync client
* Fixed pytest suite for google and cohere embeddings

## How to test?

* Create a session with an invalid API key for each provider.  Make sure it fails the validate step with an authentication-related failure
* Run the new tests, with and without anthropic, cohere, and google extras
* Run subset of embedding and semantic tests with all providers

## Why make this change?

Previously, API key validation only happened when a model was actually used, which could lead to delayed errors during execution. This change provides immediate feedback during session initialization, helping users identify configuration issues early.


src/fenic/_inference/anthropic/anthropic_provider.py
```diff
@@ -0,0 +1,43 @@
...
+    
+    async def validate_api_key(self) -> None:
+        """Validate Anthropic API key by making a minimal completion request."""
+        client = self.get_client()
+        _ = await client.models.list()
+        logger.debug("Anthropic API key validation successful")
+
+
+# Create singleton instance
+anthropic_provider = AnthropicModelProvider()
```
bcallender: (keeping comment here for records, discussed in person) 
I'm confused as to why these are singletons -- they don't really have any state to maintain that would benefit from being a singleton. It would be one thing if we were storing the created client and reusing it, but `get_client` will just recreate the client each time anyway.

src/fenic/_backends/local/model_registry.py
```diff
@@ -71,11 +77,21 @@ def __init__(self, config: ResolvedSemanticConfig):
...
             for alias, model_config in embedding_model_config.model_configs.items():
-                models[alias] = self._initialize_embedding_model(model_config)
+                model = self._initialize_embedding_model(model_config)
+                models[alias] = model
+                validate_providers.add(model.client.model_provider_class)
             self.embedding_model_registry = EmbeddingModelRegistry(
                 models=models,
                 default_model=models[embedding_model_config.default_model],
             )
+        for provider in validate_providers:
```
bcallender: (also discussed in person, leaving for transparency)
This defeats the purpose of having the validation function be async if we are blocking on them sequentially.

YoungVor: Thanks for the catch, went back to doing them in parallel

src/fenic/_inference/anthropic/anthropic_provider.py
YoungVor: Agreed, changing them to instances of the provider class, since we aren't keeping state anyway

src/fenic/_inference/anthropic/anthropic_provider.py
```diff
@@ -0,0 +1,31 @@
...
```
bcallender: we don't need the singleton anymore

src/fenic/core/_inference/model_provider.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+    
+    @property
+    @abstractmethod
+    def name(self) -> str:
+        """Return the string identifier for this provider."""
+        pass
+    
+    @abstractmethod
+    def get_client():
```
bcallender: nit: since a new client is being created each time, i would say this would more accurately be `create_client`.

src/fenic/_inference/google/google_provider.py
```diff
@@ -0,0 +1,55 @@
...
+    @property
+    def name(self) -> str:
+        return "google-developer"
+    
+    def get_client(self):
+        # Native gen-ai client. 
+        if "GEMINI_API_KEY" in os.environ:
+            return genai.Client(api_key=os.environ["GEMINI_API_KEY"])
+        else:
+            return genai.Client()
```
bcallender, line 45: its odd that this returns a synchronous client while all of the others return an async one -- i would think that either _all_ of them would return async clients, or the `ModelProviderClass` would expose methods for getting either a sync or async client.

tests/api/test_session.py
```diff
@@ -504,4 +506,108 @@ def test_model_profile_validation():
             semantic=SemanticConfig(
                 language_models={"o3": OpenAILanguageModel(model_name="o3", rpm=100, tpm=1000, profiles={"fast": GoogleDeveloperLanguageModel.Profile(thinking_token_budget=0)})}
             )
-        )
\ No newline at end of file
+        )
+
+def test_session_config_with_invalid_api_keys(monkeypatch):
```
bcallender, line 511: these are nice, thanks for figuring this out.

src/fenic/_backends/local/model_registry.py
```diff
@@ -263,6 +279,7 @@ def _initialize_language_model(self, model_config: ResolvedModelConfig) -> Langu
                         AnthropicBatchCompletionsClient,
                     )
                 except ImportError as err:
+                    breakpoint()
```
german-typedef: Please remove.

src/fenic/_backends/local/model_registry.py
```diff
@@ -302,3 +319,19 @@ def _initialize_language_model(self, model_config: ResolvedModelConfig) -> Langu
...
+    except Exception as e:
+        for task in tasks:
+            if not task.done():
+                task.cancel()
+        await asyncio.gather(*tasks, return_exceptions=True)
+        raise ConfigurationError(f"Error during API key validation: {e}") from e
+    finally:
+        for task in tasks:
+            if not task.done(): 
+                task.cancel()
```
german-typedef: nit: prefer we end the file with '\n', when I see this I always think there is something wrong :).

src/fenic/_inference/anthropic/anthropic_provider.py
german-typedef: see my comment about the EOF

src/fenic/_inference/common_openai/openai_provider.py
```diff
@@ -0,0 +1,29 @@
...
+        """Get OpenAI client instance."""
+        client = AsyncOpenAI()
+        return client
+    
+    async def validate_api_key(self) -> None:
+        """Validate OpenAI API key by listing models."""
+        client = self.get_client()
+        _ = await client.models.list()
+        logger.debug("OpenAI API key validation successful")
+
```
german-typedef: nit: extra '\n', I thought Trunk will flag something like that.

src/fenic/_inference/google/gemini_batch_embeddings_client.py
```diff
@@ -134,4 +129,4 @@ def reset_metrics(self):
         self._metrics = RMMetrics()
 
     def get_metrics(self) -> RMMetrics:
-        return self._metrics
+        return self._metrics
```
german-typedef: nit: EOF

src/fenic/_inference/openai/openai_batch_chat_completions_client.py
```diff
@@ -128,4 +124,4 @@ def _get_max_output_tokens(self, request: FenicCompletionsRequest) -> int:
 
         # Get profile-specific reasoning effort
         profile_config = self._profile_manager.get_profile_by_name(request.model_profile)
-        return base_tokens + profile_config.expected_additional_reasoning_tokens
+        return base_tokens + profile_config.expected_additional_reasoning_tokens
```
german-typedef: nit: EOF

src/fenic/core/_inference/model_provider.py
YoungVor: good point, will change

src/fenic/_backends/local/model_registry.py
YoungVor: thanks, good catch

src/fenic/_backends/local/model_registry.py
YoungVor: sure!  Done.

src/fenic/_inference/google/google_provider.py
YoungVor: Agreed.  I'll make it consistent one way or another

src/fenic/_inference/common_openai/openai_provider.py
YoungVor: yeah, it would be nice if it catches anything other than a single '\n'!  Maybe @scottames knows if we could update the linter trunk uses

PR #171, opened 2025-08-31
feat: support openrouter as a model provider

bcallender:
# feat: support openrouter as a model provider

This PR adds support for OpenRouter as a model provider, allowing users to access a wide range of LLMs through a single API.

Key changes:
- Added `OpenRouterModelProvider` class that dynamically loads available models from OpenRouter API
  - Added system for `model_catalog` to be able to add functions at runtime to lazy load the available models from providers like OpenRouter only if an `OpenRouterLanguageModel` was provisioned.
- Created `OpenRouterBatchChatCompletionsClient` for handling chat completions to OpenRouter -- there ended up being too much distinct logic specific to OpenRouter (like metrics and handling structured outputs) to reuse the OpenAI Core 
- Added configs/profiles for OpenRouter settings like provider routing preferences, still more to add there
- Implemented `AdaptiveBackoffRateLimitStrategy` to handle rate-limiting for OpenRouter -- starts off with a very high RPM and will gradually ramp it down if 429s are encountered. If OpenRouter provides headers for the rate limiting that provide a `x-ratelimit-limit` and `x-ratelimit-reset`, we ramp down the rpm bucket to that new limit and wait until the limit resets before resuming request transmissions. 
- Added configuration options for the `OpenRouterLanguageModel` that align with the OpenRouter API -- users can specify fallback models and a variety of provider filtering/sorting configurations for provider routing. 
- Updated model registry to initialize OpenRouter models

The implementation follows the same patterns as other providers, with OpenRouter-specific features like provider routing and reasoning parameters.


bcallender:
* **#212** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/212?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#171** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/171?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/171?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#199** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/199?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/fenic/_inference/rate_limit_strategy.py
```diff
@@ -102,6 +102,31 @@ def context_tokens_per_minute(self) -> int:
...
+    - backoff zeros request capacity to briefly pause on provider 429s
+    - context_tokens_per_minute returns a very large sentinel value
+    """
+
+    def __init__(self, rpm: int = 50_000_000):
+        super().__init__(rpm=rpm)
+
+    def backoff(self, curr_time: float) -> int:
+        # Minimal backoff: drop request bucket capacity to 0 to yield scheduling
+        self.requests_bucket._set_capacity(0, curr_time)
```
rohitrastogi: driveby: does this do anything if we're not ever updating the requests_bucket?

YoungVor: thinking about this as well- do we want exponential backoff in this case?  We could just use the requests_bucket and ignore the token bucket

src/fenic/_inference/openrouter/openrouter_provider.py
```diff
@@ -0,0 +1,176 @@
...
+                raise RuntimeError(
+                    f"OpenRouter models request failed: {resp.status_code} {resp.text}"
+                )
+            payload = resp.json() or {}
+            models = payload.get("data") or []
+            loaded = 0
+            for model in models:
+                params = self._translate_model(model)
+                if not params:
+                    continue
```
YoungVor: do we want to raise if we fail to translate the model params?

pyproject.toml
```diff
@@ -70,6 +70,7 @@ cloud = [
   "fenic-cloud>=0.1.3",
 ]
 mcp = ["fastmcp>=2.11.3"]
+openrouter = ["instructor"]
```
YoungVor: version?

src/fenic/api/session/config.py
```diff
@@ -628,6 +632,31 @@ class Profile(BaseModel):
...
+        Pass-through of selected OpenRouter chat completion parameters for parity:
+        - models: backup models for routing overrides
+        - provider_sort: provider routing preference (price, throughput, latency)
+        - reasoning_effort / reasoning_max_tokens: reasoning config (OpenRouter fields)
+        """
+        model_config = ConfigDict(extra='forbid')
+
+        reasoning_effort: Optional[Literal["high","medium","low"]] = Field(default=None, description="OpenAI-style reasoning effort")
+        reasoning_max_tokens: Optional[int] = Field(default=None, gt=0, description="Non-OpenAI-style reasoning effort (max tokens)")
+        models: Optional[list[str]] = Field(default=None, description="Alternate models for routing overrides", max_length=3)
```
YoungVor: Curious what the strategy is for choosing a different model?  Like, choose a different model on some number of failures?

bcallender: https://openrouter.ai/docs/features/model-routing#the-models-parameter

```
The models parameter lets you automatically try other models if the primary model’s providers are down, rate-limited, or refuse to reply due to content moderation.
...
If the model you selected returns an error, OpenRouter will try to use the fallback model instead. If the fallback model is down or returns an error, OpenRouter will return that error.

By default, any error can trigger the use of a fallback model, including context length validation errors, moderation flags for filtered models, rate-limiting, and downtime.
```

pyproject.toml
bcallender: removed it entirely, solved the issue a different way.

src/fenic/_inference/rate_limit_strategy.py
bcallender: I've refactored this entire system now, adding a new `AdaptiveBackoffRateLimitStrategy` that starts with a high rpm and ramps it down multiplicatively upon encontering rate limits. If a rate limit also provides an `x-ratelimit-reset` header that tells us when the rate limit will reset, we pause all requests (as much as we can, b/c of race conditions, in-flight requests etc) by returning `False` from `check_and_consume_rate_limit` until the reset time is reached.

src/fenic/_inference/openrouter/openrouter_provider.py
bcallender: I think warning is fine here -- we don't want to block all openrouter models just because one has a parameter we're not expecting.

src/fenic/_inference/openrouter/openrouter_profile_manager.py
```diff
@@ -0,0 +1,126 @@
...
+        ] = None,
+        default_profile_name: Optional[str] = None,
+    ):
+        self._model_parameters = model_parameters
+        super().__init__(
+            profile_configurations=profile_configurations,
+            default_profile_name=default_profile_name,
+        )
+
+    def _process_profile(
```
YoungVor, line 97: do we need to worry about default reasoning settings in the default_profile path (get_default_profile)?

src/fenic/_inference/openrouter/openrouter_provider.py
YoungVor: Forcing the user to tweak their config until they have the model list they expect is imo a valid behavior.  What's the experience if many or all of the models fail to load?

src/fenic/_inference/request_utils.py
```diff
@@ -6,12 +6,44 @@
 from fenic._inference.types import FenicCompletionsRequest
 
 
+def parse_openrouter_rate_limit_headers(
```
YoungVor, line 9: woudl it make sense to have this in openai_utils?

bcallender: this is only relevant to openrouter, at least for now. I think the headers for openai are different.

src/fenic/_inference/openrouter/openrouter_batch_chat_completions_client.py
```diff
@@ -0,0 +1,253 @@
...
+            self._metrics.num_output_tokens += total_output_tokens
+            self._metrics.num_requests += 1
+
+            # Cost from OpenRouter usage, fallback to catalog if not provided
+            model_extra = usage.model_extra
+            cost_value = model_extra.get("cost")
+            if isinstance(cost_value, (int, float)):
+                self._metrics.cost += float(cost_value)
+            else:
+                self._metrics.cost += model_catalog.calculate_completion_model_cost(
```
YoungVor, line 198: did you find we need this fallback?  openrouter seems to say cost will always be there if usage accounting is enabled

src/fenic/_inference/openrouter/openrouter_batch_chat_completions_client.py
```diff
@@ -0,0 +1,253 @@
...
+                    common_params[RESPONSE_FORMAT] = {
+                        "type": "json_schema",
+                        "json_schema": {
+                            "name": "fenic_response",
+                            "schema": request.structured_output.strict_schema,
+                            "strict": True,
+                        },
+                    }
+                elif TOOLS in self._model_parameters.supported_parameters:
+                    common_params[TOOLS] = [
```
YoungVor, line 129: were the tools field a openrouter specific thing, while openai protocol just uses structured outputs?  Curious why we needed it here, and do we want to use the tools response structure for openai too?

src/fenic/_inference/openrouter/openrouter_profile_manager.py
```diff
@@ -0,0 +1,126 @@
...
+from fenic.core._inference.model_catalog import CompletionModelParameters
+from fenic.core._resolved_session_config import (
+    ResolvedOpenRouterModelProfile,
+    ResolvedOpenRouterProviderRouting,
+)
+
+
+@dataclass
+class OpenRouterCompletionProfileConfiguration(BaseProfileConfiguration):
+    """Extends OpenAI profile configuration with OpenRouter extras."""
```
YoungVor: comment out of date?  Or should it extend OpenAICompletionProfileConfiguration?  Extends Base

src/fenic/api/session/config.py
```diff
@@ -620,13 +637,92 @@ class Profile(BaseModel):
...
+class OpenRouterLanguageModel(BaseModel):
+    """Configuration for OpenRouter language models.
+
+    This class defines the configuration settings for OpenRouter language models,
+    including model selection and rate limiting parameters. When fetching available models from OpenRouter, results
+    will be filtered to only include models from providers that are not in the user’s ignored providers list and are either
+    in the user’s allowed providers list (if configured) or from any provider (if no allowed providers are specified).
+    """
+
+    model_name: str = Field(...,description="The name of the OpenRouter model to use, typically `{provider}/{model_name}`")
```
YoungVor: do we want to enforce the {provider}/{model_name} syntax?  Or have a separate provider_name field?

src/fenic/api/session/config.py
```diff
@@ -620,13 +637,92 @@ class Profile(BaseModel):
...
+        max_completion_price: Optional[float] = Field(
+            default=None, description="Maximum completion price per 1M tokens."
+        )
+
+    class Profile(BaseModel):
+        """Profile configurations for OpenRouter language models.
+
+        Pass-through of selected OpenRouter chat completion parameters for parity:
+        - models: backup models for routing overrides
+        - provider: provider routing preferences
```
YoungVor: maybe something like routing include/exclude preferences?

src/fenic/api/session/config.py
```diff
@@ -620,13 +637,92 @@ class Profile(BaseModel):
...
+            default=None, description="Maximum prompt price per 1M tokens."
+        )
+        max_completion_price: Optional[float] = Field(
+            default=None, description="Maximum completion price per 1M tokens."
+        )
+
+    class Profile(BaseModel):
+        """Profile configurations for OpenRouter language models.
+
+        Pass-through of selected OpenRouter chat completion parameters for parity:
```
YoungVor: user won't care about parity - if reasoning_effort and reasoning_max_tokens take effect for some models available through openrouter, we can say that

src/fenic/_inference/rate_limit_strategy.py
```diff
@@ -102,6 +108,84 @@ def context_tokens_per_minute(self) -> int:
...
+        self._on_cooldown = False
+        self._lock = threading.Lock()
+
+    def register_rate_limit_hint(
+        self, rpm_hint: int | None, retry_at_epoch_seconds: float | None
+    ) -> None:
+        """Register provider-processed hints: rpm limit and absolute retry time.
+
+        Args:
+            rpm_hint: Max requests per minute allowed by provider (if known)
```
YoungVor, line 163: how will this play with alternative route providers that might have different rpms?

src/fenic/_inference/rate_limit_strategy.py
```diff
@@ -102,6 +108,84 @@ def context_tokens_per_minute(self) -> int:
...
+    - Optionally accepts provider hints via register_openrouter_rate_limit to clamp rpm.
+    """
+
+    def __init__(
+        self, rpm: int = 10_000, min_rpm: int = 50, backoff_multiplier: float = 0.75
+    ):
+        super().__init__(rpm=rpm)
+        self._min_rpm = max(1, min_rpm)
+        self._backoff_multiplier = (
+            backoff_multiplier if 0 < backoff_multiplier < 1 else 0.75
```
YoungVor: would prefer to throw internal error if backoff_multiplier is out of expected range.

src/fenic/_inference/rate_limit_strategy.py
```diff
@@ -102,6 +108,84 @@ def context_tokens_per_minute(self) -> int:
...
+        )
+        self._rpm_hint: int | None = None
+        self._cooldown_until: float = 0.0
+        self._on_cooldown = False
+        self._lock = threading.Lock()
+
+    def register_rate_limit_hint(
```
YoungVor, line 157: just curious - is there any documentation/insight from OpenRouter on how often they provide rate limit suggestions?

src/fenic/api/session/config.py
```diff
@@ -1093,14 +1266,20 @@ def resolve_model(model: ModelConfig) -> ResolvedModelConfig:
...
                     default_profile=model.default_profile,
                 )
-            elif isinstance(model, (GoogleDeveloperEmbeddingModel, GoogleVertexEmbeddingModel)):
-                resolved_profiles = {
-                    profile_name: ResolvedGoogleModelProfile(
-                        embedding_dimensionality=profile.output_dimensionality,
-                        embedding_task_type=profile.task_type,
-                    ) for
-                    profile_name, profile in model.profiles.items()
-                } if model.profiles else None
```
YoungVor, line 1103: curious, there was a lot of space formatting changes in this pr.  Did the linter change?

src/fenic/api/session/config.py
```diff
@@ -620,13 +637,92 @@ class Profile(BaseModel):
...
+        thinking_token_budget: Optional[int] = Field(default=None, description="The thinking budget in tokens for the profile", ge=1024)
+
+
+class OpenRouterLanguageModel(BaseModel):
+    """Configuration for OpenRouter language models.
+
+    This class defines the configuration settings for OpenRouter language models,
+    including model selection and rate limiting parameters. When fetching available models from OpenRouter, results
+    will be filtered to only include models from providers that are not in the user’s ignored providers list and are either
+    in the user’s allowed providers list (if configured) or from any provider (if no allowed providers are specified).
```
YoungVor, line 654: Would be good to include a few examples, one with just a single provider/model and options, one with some routing preferences

src/fenic/_inference/openrouter/openrouter_provider.py
```diff
@@ -0,0 +1,185 @@
...
+            supports_profiles=True,
+            supports_reasoning=supports_reasoning,
+            supports_custom_temperature=supports_custom_temperature,
+            supported_parameters=supported_params,
+        )
+
+    def _load_models_once(self):
+        if self._models_loaded:
+            return
+        with self._models_lock:
```
YoungVor: do we need the _models_lock lock in the provider if we also have the _loader_mutex when we call registered loaders?

src/fenic/core/_inference/model_catalog.py
```diff
@@ -1,8 +1,12 @@
+import logging
```
YoungVor, line 1: can there be some unit tests in test_model_catalog for dynamically loading models?

src/fenic/_inference/rate_limit_strategy.py
```diff
@@ -102,6 +108,84 @@ def context_tokens_per_minute(self) -> int:
...
+class AdaptiveBackoffRateLimitStrategy(RateLimitStrategy):
+    """Simple RPM limiter that multiplicatively reduces RPM on backoff.
+
+    - Uses a single requests bucket; no token accounting.
+    - On backoff(), reduce rpm by multiplier and zero capacity to remove burst.
+    - Optionally accepts provider hints via register_openrouter_rate_limit to clamp rpm.
+    """
+
+    def __init__(
+        self, rpm: int = 10_000, min_rpm: int = 50, backoff_multiplier: float = 0.75
```
YoungVor: are there any scenarios we want to increase rpm again, long after backoff multiplier has reduced the ceiling?

src/fenic/_inference/openrouter/openrouter_batch_chat_completions_client.py
bcallender: Tools are a general thing for all models that support them. OpenAI models support structured outputs using `json_schema`, so we have no reason to use tools for those models to get structured output out of them. We would only use tools for models that don't support structured output in the response format, like some of the models on OpenRouter and all of the Anthropic models.

src/fenic/_inference/openrouter/openrouter_profile_manager.py
bcallender: fixed

src/fenic/api/session/config.py
bcallender: updated language to be more detailed

src/fenic/_inference/openrouter/openrouter_provider.py
bcallender: I added a `raise` if we indeed can't translate any models at all, that would probably be a blocker on working with OpenRouter at all. We really shouldn't ever fail here unless openrouter changes its API on us though.

src/fenic/_inference/openrouter/openrouter_profile_manager.py
bcallender: yep -- added the default reasoning there too.

src/fenic/api/session/config.py
bcallender: when i was trying to add an import using the quick fix menu in cursor, I accidentally enabled the `black` linter, and didn't realize until i had already made some other changes with the new formatting. i've tried to go back and change the formatting back, but there's a lot of small places that are evading me

src/fenic/api/session/config.py
bcallender: done

src/fenic/_inference/rate_limit_strategy.py
bcallender: nothing useful: `https://openrouter.ai/docs/api-reference/limits#rate-limits-and-credits-remaining`

src/fenic/core/_inference/model_catalog.py
bcallender: yes, done.

src/fenic/_inference/rate_limit_strategy.py
bcallender: currently this only supports hints that are provided by `OpenRouter` itself, when it is gating access to a model for some reason (it's free, or there is some other limitation (mistral-small-3b is just inexplicably limited to 100rpm with no warning on the model page). If the _provider_ is rate limiting us, OpenRouter will just try a different provider and then if none work (I think) return an error that we would handle in the client itself. these hints in headers are only from OpenRouter (I think), but none of this is documented, so  ¯\_(ツ)_/¯

I don't think we can account for each model provider's rate limiting, at least not for now.

src/fenic/_inference/rate_limit_strategy.py
bcallender: fair enough, done!

src/fenic/core/_inference/model_catalog.py
bcallender: yep! done.

src/fenic/_inference/rate_limit_strategy.py
bcallender: added logic to allow the strategy to additively increase RPM unless it was clamped by an explicit rpm hint.

src/fenic/_inference/openrouter/openrouter_provider.py
bcallender: probably not. i removed it.

src/fenic/_inference/openrouter/openrouter_batch_chat_completions_client.py
bcallender: i never saw it used, to be honest. i don't think it hurts to leave it here though -- if that cost value is `None` somehow it would break all of the downstream logic.

src/fenic/api/session/config.py
YoungVor: ah, makes sense.  No worries

src/fenic/_inference/openrouter/openrouter_batch_chat_completions_client.py
YoungVor: agreed, leaving in the fallback shouldn't hurt and protects us from future api changes

src/fenic/_inference/openrouter/openrouter_batch_chat_completions_client.py
```diff
@@ -0,0 +1,253 @@
...
+        except RateLimitError as e:
+            if isinstance(self.rate_limit_strategy, AdaptiveBackoffRateLimitStrategy):
+                rpm_hint, retry_at_s = parse_openrouter_rate_limit_headers(
+                    e.response.headers
+                )
+                self.rate_limit_strategy.register_rate_limit_hint(rpm_hint, retry_at_s)
+            return TransientException(e)
+        except (APITimeoutError, APIConnectionError) as e:
+            return TransientException(e)
+        # encountered when the response is not valid JSON. can sometimes be fixed with a retry
```
YoungVor, line 225: wouldn't openrouter try the fallback provider on the server end before returning this error?

src/fenic/_inference/openrouter/openrouter_provider.py
```diff
@@ -0,0 +1,191 @@
...
+            if isinstance(model_id, str):
+                # Register into global catalog so standard lookups succeed
+                model_catalog.add_model(ModelProvider.OPENROUTER, model_id, params)
+                added_models.append(model_id)
+        if not added_models:
+            raise SessionError("Failed to process and load OpenRouter models")
+        self._models_loaded = True
+        logger.info(f"OpenRouter model cache initialized with {len(added_models)} models")
+        if untranslated_models:
+            logger.info(f"Failed to process and load OpenRouter models: {untranslated_models}")
```
YoungVor: Lets make it a Warn

tests/_inference/test_model_catalog.py
```diff
@@ -61,4 +168,71 @@ def test_all_embedding_models_have_valid_parameters(models: Enum, provider: Mode
...
-        assert params.output_dimensions, f"Missing output_dimensions for {provider} embedding model: {model_name}"
\ No newline at end of file
+        assert params.output_dimensions, f"Missing output_dimensions for {provider} embedding model: {model_name}"
+
+def test_openrouter_provider_loads_models(mock_openrouter_models):
+    """Test that the OpenRouter provider can fetch the models from the OpenRouter API."""
+    catalog = model_catalog
+    assert len(catalog._get_supported_completions_models_by_provider(ModelProvider.OPENROUTER)) == 5
+
+def test_openrouter_provider_loads_openai_models_correctly(mock_openrouter_models):
```
YoungVor, line 194: nice, this is a good test to add

tests/_inference/test_rate_limit_strategy.py
```diff
@@ -0,0 +1,76 @@
+import typing as _t
```
YoungVor, line 1: thank you for adding the ratelimit unit tests!

src/fenic/_inference/rate_limit_strategy.py
```diff
@@ -107,6 +110,139 @@ def context_tokens_per_minute(self) -> int:
...
+        # Cap upward growth; default to the starting rpm
+        self._max_rpm = max(rpm, self._min_rpm) if max_rpm is None else max(max_rpm, self._min_rpm)
+        # Additive increase controls (only when no provider hint is present)
+        self._additive_increment = max(1, additive_increment)
+        self._increase_after_successes = max(1, increase_after_successes)
+        self._consecutive_successes = 0
+        self._rpm_hint: int | None = None
+        self._cooldown_until: float = 0.0
+        self._on_cooldown = False
+        self._lock = threading.Lock()
```
YoungVor: can you just use the mutex in base RateLimitStrategy here?

YoungVor: it doesn't seem to be used currently anyway

tests/conftest.py
```diff
@@ -150,6 +151,13 @@ def pytest_addoption(parser):
         default=None,
         help="If set, will run reader tests that read from HuggingFace.",
     )
+    parser.addoption(
```
YoungVor: where is this used?

src/fenic/_inference/openrouter/openrouter_batch_chat_completions_client.py
bcallender: no. this is a 'successful' request in openrouter's eyes, we passed an input and received some tokens back, the error is happening when the openai library attempts to parse the json into the Pydantic object.

src/fenic/_inference/rate_limit_strategy.py
bcallender: yeah we totally can, will fix.

tests/conftest.py
bcallender: ah it was used previously, but replaced it with a pytest mark instead, will remove

src/fenic/_inference/rate_limit_strategy.py
```diff
@@ -107,6 +110,139 @@ def context_tokens_per_minute(self) -> int:
...
+                self._cooldown_until = float(retry_at_epoch_seconds)
+                if not self._on_cooldown:
+                    self._on_cooldown = True
+                    logger.info(
+                        f"Provider is throttling requests -- pausing for {retry_at_epoch_seconds - time.time():.2f}s before resuming at the provider specified limit of {rpm_hint} requests per minute"
+                    )
+
+    def backoff(self, curr_time: float) -> int:
+        """Backoff the request rate limit bucket."""
+        with self._lock:
```
YoungVor: no cooldown usage in backoff?

src/fenic/_inference/rate_limit_strategy.py
```diff
@@ -107,6 +110,139 @@ def context_tokens_per_minute(self) -> int:
...
+                and retry_at_epoch_seconds > 0
+            ):
+                self._cooldown_until = float(retry_at_epoch_seconds)
+                if not self._on_cooldown:
+                    self._on_cooldown = True
+                    logger.info(
+                        f"Provider is throttling requests -- pausing for {retry_at_epoch_seconds - time.time():.2f}s before resuming at the provider specified limit of {rpm_hint} requests per minute"
+                    )
+
+    def backoff(self, curr_time: float) -> int:
```
YoungVor, line 186: no cooldown usage in backoff?

src/fenic/_inference/rate_limit_strategy.py
bcallender: we don't need to use it in the backoff -- the backoff will just reduce the RPM if a hint wasn't provided

we only need to use it in `check_and_consume_rate_limit` to ensure we don't continue while we're still in cooldown, and in  `register_rate_limit_hint` where the cooldown would be established if openrouter returned an `x-ratelimit-reset` header.

PR #176, opened 2025-09-02
fix: add retry to OpenAI for 404 errors when no processing was done

german-typedef:
### TL;DR

Added a TransientException if we get a 404 and no processing from OpenAI.
Fenic's batch processing should be able to handle the exception and perform the retry.

### What changed?

- The error handling for both openai chat and embeddings now handle the NotFoundError. if we have the headers and the headers indicate no work was done on OpenAI, we'll retry the operation.

### DIFF
--- a/src/fenic/_inference/common_openai/openai_chat_completions_core.py
+++ b/src/fenic/_inference/common_openai/openai_chat_completions_core.py
@@ -7,6 +7,7 @@ from openai import (
     APIConnectionError,
     APITimeoutError,
     AsyncOpenAI,
+    NotFoundError,
     OpenAIError,
     RateLimitError,
 )
@@ -177,6 +178,18 @@ class OpenAIChatCompletionsCore:
 
         except (RateLimitError, APITimeoutError, APIConnectionError) as e:
             return TransientException(e)
+
+        except NotFoundError as e:
+            # During our CI tests, where we run a larger set of tests, we've seen an intermittent 404 error
+            # that is not coming from OpenAI.
+            # Usually 404 errors should be considered as fatal, as there is no use in retrying them,
+            # in this case some of the requests have succeeded, so we're marking this as transient and as such
+            # the request will be retried.
+            if e.response.headers.get("openai-processing-ms") == "0":
+                logger.error("404 with zero processing time → likely routing glitch - Marking this as transient.")
+                return TransientException(e)
+            else:
+                return FatalException(e)
 
         except OpenAIError as e:
             return FatalException(e)
--- a/src/fenic/_inference/common_openai/openai_embeddings_core.py
+++ b/src/fenic/_inference/common_openai/openai_embeddings_core.py
@@ -8,6 +8,7 @@ from openai import (
     APIConnectionError,
     APITimeoutError,
     AsyncOpenAI,
+    NotFoundError,
     OpenAIError,
     RateLimitError,
 )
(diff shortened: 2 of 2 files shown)


german-typedef:
* **#176** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/176?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/176?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/fenic/_inference/common_openai/openai_provider.py
```diff
@@ -1,13 +1,33 @@
...
 from openai import AsyncOpenAI, OpenAI
 
+from fenic._inference.model_client import TransientException
 from fenic.core._inference.model_provider import ModelProviderClass
 
 logger = logging.getLogger(__name__)
 
+class TransientErrorHandlerAsyncTransport(httpx.AsyncHTTPTransport):
+    """
+    Handle transient 404 errors from OpenAI.
```
YoungVor: maybe worth adding a sentence describing the issue we saw - that it happens rarely but enough to break CI, and that the error seemed to come from cloudflare without any openAI payload

src/fenic/_inference/common_openai/openai_provider.py
```diff
@@ -22,10 +42,20 @@ def create_client(self):
...
         """Validate OpenAI API key by listing models."""
         client = self.create_aio_client()
         _ = await client.models.list()
         logger.debug("OpenAI API key validation successful")
+
+    @classmethod
+    def _create_http_async_client(cls):
+        return httpx.AsyncClient(
+            http2=False,
+            limits=httpx.Limits(max_keepalive_connections=10, max_connections=20, keepalive_expiry=5.0),
```
YoungVor: are these the same limits as default?

YoungVor (approved):
lgtm, pending a small nit on the comment and the limits default

src/fenic/_inference/common_openai/openai_provider.py
rohitrastogi: The SDK will return NotFoundError which has the httpx.Response. Instead of doing the check here, we can do it in openai_chat_completions_core.py and openai_embeddings_core.py where se the other transient exceptions in case of 429s. This will consolidate all of the retry logic into one location.

german-typedef: I'm moved the code to the chat and embeddings files to handle: NotFoundError.

rohitrastogi (approved):

src/fenic/_inference/common_openai/openai_chat_completions_core.py
```diff
@@ -178,6 +179,18 @@ async def make_single_request(
         except (RateLimitError, APITimeoutError, APIConnectionError) as e:
             return TransientException(e)
 
+        except NotFoundError as e:
+            # During our CI tests, where we run a larger set of tests, we've seen an intermittent 404 error
+            # that is not coming from OpenAI.
+            # Usually 404 errors should be considered as fatal, as there is no use in retrying them,
+            # in this case some of the requests have succeeded, so we're marking this as transient and as such
+            # the request will be retried.
+            if e.response.headers.get("openai-processing-ms") == "0":
```
rohitrastogi, line 188: what happens if the key does not exist in get()? Will it error?

rohitrastogi: Nvmd looked at httpx docs, it returns None, which is safe.

PR #177, opened 2025-09-03
fix: write one metrics row per query not per operator to local fenic_system.query_metrics system table

rohitrastogi:
* Moved `insert_query_metrics` out of `PhysicalPlan.execute()` (which was causing metrics to be written **per operator**).
* Now `insert_query_metrics` is called in  local `execution` , ensuring metrics are written **once per query** instead of once per operator.
* Added logging for both `execution_id` at query start and `session_id` at session start.
* Rename `fenic_system.metrics` system table to `fenic_system.query_metrics` for clarity.

### DIFF
--- a/src/fenic/api/catalog.py
+++ b/src/fenic/api/catalog.py
@@ -42,31 +42,33 @@ class Catalog:
 
     Query metrics are recorded for each session and stored locally
     in `fenic_system.query_metrics`. Metrics can be loaded into a DataFrame
-    for analysis, and a summary is printed when a session ends.
+    for analysis.
 
-    Example: Basic Metrics Table queries
+    Example:
         ```python
-        # Load all metrics for the current session
+        # Load all metrics for the current application
         metrics_df = session.table("fenic_system.query_metrics")
 
-        # Show the 10 most recent queries
-        session.sql(\"\"\"
+        # Show the 10 most recent queries in the application
+        recent_queries = session.sql(\"\"\"
             SELECT *
             FROM {df}
             ORDER BY CAST(end_ts AS TIMESTAMP) DESC
             LIMIT 10
         \"\"\", df=metrics_df)
+        recent_queries.show()
 
-        # Get metrics for queries in a specific session with non-zero LM costs
-        session.sql(\"\"\"
+        # Find query metrics for a specific session with non-zero LM costs
+        specific_session_queries = session.sql(\"\"\"
             SELECT *
             FROM {df}
             WHERE session_id = '9e7e256f-fad9-4cd9-844e-399d795aaea0'
                 AND total_lm_cost > 0
             ORDER BY CAST(end_ts AS TIMESTAMP) ASC
         \"\"\", df=metrics_df)
+        specific_session_queries.show()
 
-        # Total LM costs and requests between a specific time window
+        # Aggregate total LM costs and requests between a specific time window
         metrics_window = session.sql(\"\"\"
             SELECT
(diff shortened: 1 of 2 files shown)


src/fenic/_backends/local/execution.py
```diff
@@ -262,3 +209,72 @@ def _build_read_parquet_query(
...
+                    f"Existing schema: {existing_schema} "
+                    f"Your DataFrame schema: {plan_schema} "
+                    f"To fix this: "
+                    f"1) Use mode='overwrite' to replace the table with your DataFrame's schema, "
+                    f"2) Modify your DataFrame to match the existing table's schema, "
+                    f"3) Use a different table name."
+                )
+        return True
+
+    def _should_write_file(
```
YoungVor, line 249: nice, definitely a bit cleaner with the helper methods

tests/_backends/local/catalog/test_metrics_table.py
```diff
@@ -3,74 +3,75 @@
...
+    initial_metrics = local_session.table("fenic_system.query_metrics")
     initial_count = initial_metrics.count()
-    
+
     # Execute some operations
     sample_df.show()  # This should add 1 metric row
     sample_df.count()  # This should add another metric row
     sample_df.collect()  # This should add another metric row
-    
+    sample_df.select("name").filter(col("name") == "Alice").show()  # This should add another metric row
```
YoungVor: good call, just a slightly more complicated plan would have caught this

src/fenic/api/catalog.py
```diff
@@ -13,71 +13,66 @@
...
+        session.catalog.create_table(
+            "my_table",
+            Schema([ColumnField("id", IntegerType)])
+        )  # → True
+        ```
 
-    Fenic keeps track of system metrics and costs per session in your app.  It persists
-    in your local database, and you can load it into a dataframe for analysis.
-    A summary of your session's metrics will print once you stop your session.
+    ### Metrics Table (Local Sessions Only)
```
YoungVor, line 41: good call on local sessions only

src/fenic/api/catalog.py
```diff
@@ -13,71 +13,66 @@
...
+    Example:
         ```python
-        # Get the metrics for a session
-        metrics_df = session.table("fenic_system.metrics")
+        # Load all metrics for the current session
+        metrics_df = session.table("fenic_system.query_metrics")
 
-        # Show metrics from the last 10 queries
+        # Show the 10 most recent queries
         metrics_df.order_by(fc.col("index").desc()).limit(10).show()
```
YoungVor: remove the index

src/fenic/api/catalog.py
```diff
@@ -13,71 +13,66 @@
...
             fc.sum("total_lm_requests").alias("session_lm_requests")
         ).show()
         ```
 
-    Example: Query metrics by timestamp
-        You can use the index column to order the metrics, or you can use the start_ts and end_ts columns.
-        Currently, fenic does not natively support timestamp types so these are strings. To query by timestamps,
-        CAST them to timestamp in a SQL query.
+    ### Querying by Timestamp
+    The `index` column provides ordering, while `start_ts` and `end_ts` can
```
YoungVor: remove the index

YoungVor (approved):
looks good.  thanks rohit!

src/fenic/api/catalog.py
rohitrastogi: Done

src/fenic/api/catalog.py
rohitrastogi: Done

PR #189, opened 2025-09-04
add no-API-key playground with MockLLM, canned responses, and README

Hanzzalaaa:
…und:

- playground.py with MockLLM for semantic_extract and semantic_join
- canned_responses.json with sample mock outputs
- README.md with quickstart guide and usage instructions

This helps new users try fenic workflows locally without API setup.

### DIFF
--- a/examples/playground/README.md
+++ b/examples/playground/README.md
@@ -1,0 +1,26 @@
+# Fenic Playground (No API Key)
+
+This playground lets you try out **fenic-style workflows locally without any API key**.  
+It uses a lightweight `MockLLM` that returns canned responses so you can explore core ideas like:
+
+- Async UDF-style processing with batching (`asyncio.gather`)  
+- `semantic_extract` producing structured output validated by **Pydantic**  
+- A mocked `semantic_join` to illustrate similarity-based joins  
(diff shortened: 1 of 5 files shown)


YoungVor:
Hey Hanzala, thank you for joining the fenic community, we're excited to see what folks can do with fenic!

I'm curious about the vision for the playground, and the usecases you have in mind? 
If its a cost concern, you can set up a Google account with the free tier and play with the semantic layer in fenic.
If its more of a security or privacy concern, that's a reason for us to explore integrating with local options (like ollama, or litellm with a proxy server running a model).  

There are could be some benefits in mocking the LLM response from a testing perspective.  For e.g., if we wrote MockLLM to be MockOpenAIProvider (and replaced the provider/client classes with mocks), we could test that the semantic layer works correctly with complex queries without incurring completions costs.

bcallender:
Closing this as part of a PR backlog prune. The branch has gone stale against current `main`, and we do not expect to merge this no-API-key playground implementation in its current form. The underlying idea may still be useful, but it needs a current product decision and a fresh implementation shape before it can land safely. Thanks for the work here.

PR #190, opened 2025-09-07
fix: enforce that return types of all branches in conditional chains are equal and handle child exprs in IsNullExpr

rohitrastogi:
1. Validate that all branch return types in when/when/otherwise are equal
   - Previously did not validate this and did not call to_column_field for all branches
   - This led to invalid expression trees that fail during runtime and not planning time

2. Disallow Column.when() variant to be called on non-when expressions
   - PySpark allows this but effectively ignores the non-when expression
   - This violates principle of least surprise, so we error out instead

3. Call to_column_field for NullExpr children and add test validating for this

4. Change appended metric log level from debug to info
   - Too noisy for default configure_logging() setup

5. Clean up docstrings for when/when/otherwise
   - Make behavior more explicit
   - Remove confusing descriptions about "list of conditions"
   - Add notes about type consistency requirements

### DIFF
--- a/src/fenic/_backends/local/system_table_client.py
+++ b/src/fenic/_backends/local/system_table_client.py
@@ -605,7 +605,7 @@ class SystemTableClient:
                 ),
             )
             # trunk-ignore-end(bandit/B608)
-            logger.info(f"Appended metrics for execution {metrics.execution_id}")
+            logger.debug(f"Appended metrics for execution {metrics.execution_id}")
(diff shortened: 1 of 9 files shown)


bcallender:
`Change appended metric log level from debug to info` 
info to debug?

rohitrastogi:
> `Change appended metric log level from debug to info` info to debug?

updated

PR #191, opened 2025-09-08
feat: semantic.pdf_parse and support for google models

YoungVor:
### TL;DR

Added PDF parsing capabilities to Fenic with a new `semantic.parse_pdf()` function that leverages Google VLM  capabilities to extract content from PDF files.

### What changed?

- Added a new `ParsePDFExpr` expression to the logical plan
- Created a new `BaseSingleColumnFilePathOperator` class for file-based operations
- Added file-path concept to LMRequestMessages
- Implemented the `ParsePDF` semantic operator for processing PDF files
- Added model capability flags to identify which models support PDF parsing
- Added PDF file handling in the Google Gemini client implementation
    - Leverages Googles File Storage for Gemini, adding and cleaning up the files
- Extended the token counting system to handle PDF files
- PDF parsing options for inserting page-breaks and inserting descriptions of images it finds
    - All best effort, accuracy depends on the model
- Introduced testing concept for evaluating model accuracy

### Parsing options 

Page breaks, describe images, and ignoring images smaller than min size are best-effort options.  Some of them can’t be reliably unit tested because the model doesn’t reliably add this content

## Out of Scope in this PR:

- page selectors, ignoring headers/footers, etc.  This will require a pre-processing step with PyMuPDF
- OpenAI support (next PR)
- Sending base64 encoded pdfs, or converting to png and sending base64 encoded pngs (next PR)

### How to test?

- Added unit tests test_semantic_parse_pdf.py to check raw text content
- run semantic.parse_pdf(col("pdf_path")).show() on a column of known PDF paths and compare the output

### Why make this change?

PDF parsing is a common requirement for document processing workflows. By leveraging LLM vision capabilities, we can extract structured content from PDFs without requiring separate OCR tools. This feature enables users to process documents directly within Fenic pipelines, making it easier to build end-to-end document processing applications.


YoungVor:
* **#191** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/191?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#200](https://github.com/typedef-ai/fenic/pull/200) <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/200?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#219](https://github.com/typedef-ai/fenic/pull/219) <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/219?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/191?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#234** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/234?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/fenic/_inference/google/gemini_native_chat_completions_client.py
```diff
@@ -386,9 +421,19 @@ async def make_single_request(
             if e.code == 429:
                 return TransientException(e)
             else:
+                # Clean up file on fatal client errors
+                if file_obj:
+                    await self._client.files.delete(name=file_obj.name)
+                    logger.info(f"Deleted file {file_obj.name} from Google file store (fatal client error)")
```
bcallender: these can probably be `debug` level, after you're done testing.

bcallender: also, isn't this handled by the `finally`? it would execute there after we return the `FatalException`, I don't think we need it twice. The files will auto-delete after 48 hours anyway, so the consequences for not removing the file are not that bad.

src/fenic/_backends/local/semantic_operators/parse_pdf.py
```diff
@@ -0,0 +1,82 @@
...
+        self.model = model
+        self.model_alias = model_alias
+
+        super().__init__(
+            input=input,
+            request_sender=CompletionOnlyRequestSender(
+                model=model,
+                operator_name="semantic.parse_pdf",
+                inference_config=InferenceConfiguration(
+                    max_output_tokens=100_000, # TODO: ideally figure this out based on max PDF size, or allow user to decide
```
bcallender: i was surprised this works at all since we enforce that `max_output_tokens`  is less than the `max_output_tokens` value from the model's `CompletionModelParameters` , but we do that check back on the expression side, so it wouldn't be enforced here. 

As discussed, we can figure out some way of not setting an output token limit at all for this particular use case.

src/fenic/api/functions/semantic.py
```diff
@@ -587,3 +590,28 @@ def summarize(
         SemanticSummarizeExpr(Column._from_col_or_name(column)._logical_expr, format, temperature,
                               model_alias=resolved_model_alias)
     )
+
+
+@validate_call(config=ConfigDict(strict=True, arbitrary_types_allowed=True))
+def parse_pdf(
+    column: ColumnOrName,
+    model_alias: Optional[Union[str, ModelAlias]] = None,
+    output_format: Literal["json", "markdown"] = "markdown",
```
bcallender: yeah, i think what we would want to do is have a separate `extract_pdf` or `extract_from_pdf` where the user provides a pydantic schema that we coerce the output to instead of just asking the model to return JSON. They could probably do that with back-to-back calls to `parse_pdf` and `semantic.extract`, but that would mean more model calls.

src/fenic/core/_logical_plan/utils.py
```diff
@@ -85,6 +85,32 @@ def validate_completion_parameters(
     if temperature is not None and (temperature < 0 or temperature > completion_parameters.max_temperature):
         raise ValidationError(f"[{model_provider.value}:{model_config.model_name}] temperature must be between 0 and {completion_parameters.max_temperature}")
 
+def validate_model_support_for_pdf_parsing(model_alias: Optional[ResolvedModelAlias], resolved_session_config: ResolvedSessionConfig):
+    if not resolved_session_config.semantic.language_models:
```
bcallender: all of this is the same code as above right? we can get a utility method to validate a model alias and return the `completion_model_parameters` for it and extract the shared code.

src/fenic/_backends/local/semantic_operators/parse_pdf.py
YoungVor: yep!  Done.

src/fenic/api/functions/semantic.py
YoungVor: agreed.  No reason to return json, esp since we can convert from markdown if we want

src/fenic/_backends/local/semantic_operators/parse_pdf.py
YoungVor: Although, that means I need a pre-processing step to estimate the output tokens

src/fenic/_inference/google/gemini_native_chat_completions_client.py
YoungVor: Ah, looking at this again, this is wrong.  The behavior i wanted is to only clean up on non-retriable errors, or on completion, so we don't need to upload the file on the retry.  I think the right behavior is to keep this line, but instead of adding it to the finally, to add it to the end of the main code block

YoungVor: hmm, but the trouble with that is we can eventually fail after a large number of retries.  Maybe I'll just remove the caching behavior and load the file on every call and just keep the finally

src/fenic/_backends/local/semantic_operators/parse_pdf.py
YoungVor: We should also decouple the max_token limit we want to set on the model from the token output estimation

src/fenic/core/_logical_plan/utils.py
YoungVor: good call!  Cleaned this up

src/fenic/_backends/local/semantic_operators/base.py
```diff
@@ -246,6 +246,33 @@ def build_user_message(self, input: str) -> str:
...
+        
+        This overrides the parent method to pass file_path instead of user message.
+        """
+        return LMRequestMessages(
+            system=self.build_system_message(),
+            user_file_path=file_path,  # Pass file path instead of user message
+            examples=self.build_examples() if self.examples else [],
+        )
+
+    def build_user_message(self, input: str) -> str:
```
YoungVor: Don't need to override this

src/fenic/_backends/local/semantic_operators/base.py
```diff
@@ -246,6 +246,33 @@ def build_user_message(self, input: str) -> str:
         return input
 
 
+class BaseSingleColumnFilePathOperator(
```
YoungVor, line 249: Input is a column of filepaths - add comment.

src/fenic/_backends/local/semantic_operators/parse_pdf.py
```diff
@@ -0,0 +1,73 @@
...
+        if self.page_separator:
+            if "{page}" in self.page_separator:
+                parts.append(f"Insert the page separator '{self.page_separator}' as a markdown line for each page break, replacing the '{'{page}'}' pattern with the current page number.  If the document contains page numbers, do not include them in the output, instead replace them with this page separator.")
+            else:
+                parts.append(f"Don't include the page numbers in the output, instead insert the page separator '{self.page_separator}' as a markdown line for each page break.")
+
+
+        if self.describe_images:
+            if (self.describe_images_min_size > 0):
+                parts.append(f"For each image larger than {self.describe_images_min_size} pixels, describe them briefly in a markdown section with 'Image' in the title, preserving the output order.  Ignore all images smaller than {self.describe_images_min_size} pixels.")
```
YoungVor: remove min_image idea, not measurable, not dependable

src/fenic/_inference/types.py
```diff
@@ -15,7 +15,8 @@ class FewShotExample:
 class LMRequestMessages:
     system: str
     examples: List[FewShotExample]
-    user: str
+    user: Optional[str] = None
+    user_file_path: Optional[str] = None
 
     def to_message_list(self) -> List[Dict[str, str]]:
```
YoungVor: maybe change to to_openai_message_list

YoungVor: as per @bcallender will move to openai specific code

src/fenic/_backends/local/semantic_operators/parse_pdf.py
```diff
@@ -0,0 +1,71 @@
...
+                    max_output_tokens=None,
+                    temperature=0.0,  # Use deterministic parsing
+                    model_profile=model_alias.profile if model_alias else None,
+                ),
+            ),
+            examples=None,  # PDF parsing doesn't use examples
+        )
+
+    def build_system_message(self) -> str:
+        """Build system message for PDF parsing."""
```
rohitrastogi, line 69: prefer you use a jinja template to construct the prompts. you can see an example of this in other semantic operators

YoungVor: sg, will update

protos/logical_plan/v1/expressions.proto
```diff
@@ -260,6 +261,13 @@ message SemanticSummarizeExpr {
   optional ResolvedModelAlias model_alias = 4;
 }
 
+message SemanticParsePDFExpr {
+  LogicalExpr expr = 1;
+  optional ResolvedModelAlias model_alias = 2;
+  optional string page_separator = 4;
+  bool describe_images = 6;
```
bcallender: what happened to 3 and 5 😂

src/fenic/_backends/local/semantic_operators/parse_pdf.py
YoungVor: done

protos/logical_plan/v1/expressions.proto
bcallender: since this is the first appearance of this proto message, no need to worry about backwards compat

src/fenic/_backends/local/semantic_operators/parse_pdf.py
```diff
@@ -0,0 +1,77 @@
...
+from fenic._inference.language_model import InferenceConfiguration, LanguageModel
+from fenic.core._logical_plan.resolved_types import ResolvedModelAlias
+
+logger = logging.getLogger(__name__)
+
+
+class ParsePDF(BaseSingleColumnFilePathOperator[str, str]):
+    """Operator for parsing PDF files using language models with PDF parsing capabilities."""
+    SYSTEM_PROMPT = jinja2.Template(dedent("""\
+        "Transcribe the main content of this PDF document to clean, well-formatted markdown.\n"
```
bcallender: are these literal quotes? this is already a triple quoted string, so no need to do the line by line output formatting with the newlines like this, you can just format it literally. or remove the triple quotes and format it this way, but not both.

src/fenic/_backends/local/utils/doc_loader.py
```diff
@@ -83,6 +83,12 @@ def load_docs_from_folder(
         # Process files with the appropriate handler based on extension
         return DocFolderLoader._process_files(files, max_workers, content_type)
 
+    @staticmethod
+    def check_file_extensions(file_paths: List[str], valid_file_extension: Literal["md", "json", "pdf"]) -> bool:
+        """Check that the file extensions are valid."""
+        _ = DocFolderLoader._enumerate_files(file_paths, valid_file_extension, exclude_pattern=None, recursive=False)
```
bcallender, line 90: returns `bool` or nah?

src/fenic/_backends/local/utils/doc_loader.py
```diff
@@ -277,7 +283,16 @@ def _enumerate_files_local_fs(
...
-        """Enumerate files in a local fs folder based on include/exclude patterns."""
+        """Enumerate files in a local fs folder based on include/exclude patterns.
+
+        Returns:
+            List of file paths that match the include/exclude patterns.
+
+        Raises:
+            ValidationError: If the exclude pattern is invalid.
+            FileLoaderError: If the file extension is not supported.
+            ValueError: If the path does not exist.
```
bcallender: `FileNotFoundError`?

src/fenic/_inference/common_openai/openai_utils.py
```diff
@@ -0,0 +1,45 @@
...
+
+def convert_messages(lm_request_messages: LMRequestMessages) -> List[Dict[str, str]]:
+    messages = [{"role": "system", "content": lm_request_messages.system}]
+
+    for example in lm_request_messages.examples:
+        messages.append({"role": "user", "content": example.user})
+        messages.append({"role": "assistant", "content": example.assistant})
+
+    # Handle user message based on type of content
+    if lm_request_messages.user and lm_request_messages.user_file_path:
```
bcallender: these could all be combined if you always pass content as a list right? you can just create the contents list and add the file path and/or the user message, as long as you do the filepath first?

src/fenic/_inference/google/gemini_native_chat_completions_client.py
```diff
@@ -200,8 +207,7 @@ def estimate_tokens_for_request(self, request: FenicCompletionsRequest):
         input_tokens = self.count_tokens(request.messages)
         input_tokens += self._count_auxiliary_input_tokens(request)
 
-        # Estimate output tokens
-        output_tokens = self._get_max_output_tokens(request)
+        output_tokens = self._get_max_output_tokens(request) or 0
```
bcallender: i feel like if we have no estimate for the output tokens that we should just estimate that it will return the model's `max_output_tokens`? otherwise we will be gravely underestimating for the pdf parsing case.

src/fenic/core/_inference/model_catalog.py
```diff
@@ -1026,6 +1039,19 @@ def get_supported_completions_models_as_string(self) -> str:
                 all_models.append(f"{model_provider.value}:{model}")
         return ", ".join(sorted(all_models))
 
+    def get_models_supporting_pdf_parsing_as_string(self) -> str:
```
bcallender: doesn't need to be now, but we can express this as `get_models_for_criteria(self, Callable[[CompletionModelParameters], bool])` and you could just pass `lambda params: params.supports_pdf_parsing`, so we don't need a version of this for every capability.

src/fenic/core/_logical_plan/signatures/semantic.py
```diff
@@ -42,12 +42,19 @@ def register_semantic_signatures():
...
     FunctionRegistry.register("semantic.summarize", FunctionSignature(
         function_name="semantic.summarize",
         type_signature=Exact([StringType]),  # String input (model_alias is parameter)
         return_type=StringType
     ))
 
+    # Parse PDF - parse PDF files with OCR/VLMs
+    FunctionRegistry.register("semantic.parse_pdf", FunctionSignature(
+        function_name="semantic.parse_pdf",
+        type_signature=Exact([StringType]),  # String input (model_alias is parameter)
```
bcallender: this comment can be removed.

protos/logical_plan/v1/expressions.proto
YoungVor: makes sense, will update.  (wasn't worried about backward compatibility, just didn't update the indexing after removing some fields)

src/fenic/_backends/local/semantic_operators/parse_pdf.py
YoungVor: thanks, good catch.  Removed the extra \n

src/fenic/_backends/local/utils/doc_loader.py
YoungVor: nah :)

src/fenic/_backends/local/utils/doc_loader.py
YoungVor: makes sense, can update it

src/fenic/_inference/common_openai/openai_utils.py
YoungVor: Yeah, i like that.  I was thinking about doing that earlier but thought maybe there was an extra token penalty if you pass a list of one field rather than the user message directly, but its probably not worth worrying about

src/fenic/_inference/google/gemini_native_chat_completions_client.py
YoungVor: yeah, that's fair.  I fix this in the following PR, but its a better default to estimate the max

src/fenic/core/_inference/model_catalog.py
YoungVor: i like that.  Updated

src/fenic/_inference/common_openai/openai_utils.py
YoungVor: Ah, I fixed that in a later PR, but pulled it down here

src/fenic/core/_inference/model_catalog.py
```diff
@@ -1026,6 +1039,23 @@ def get_supported_completions_models_as_string(self) -> str:
...
+        Args:
+            criteria: A function that takes CompletionModelParameters and returns True if the model matches
+
+        Returns:
+            Comma-separated string of model names
+        """
+        all_models = []
+        for model_provider in ModelProvider:
+            for model in self._get_supported_completions_models_by_provider(model_provider).keys():
+                model_params = self._get_supported_completions_models_by_provider(model_provider).get(model)
```
bcallender: nit: no need to look it up twice, just use `items()` -- `for model_name, model_params in self._get_supported_completions_models_by_provider(model_provider).items()`

YoungVor: good call!  Updated

PR #201, opened 2025-09-11
feat: support date & timestamp data types in Fenic

german-typedef:
### TL;DR

Added support for Date and Timestamp data types throughout the Fenic system.

### What changed?

- Added `DateType` and `TimestampType` to the protocol buffer definitions
- Implemented date/time functions in the API including:
  - Extraction functions: `year()`, `month()`, `day()`, `hour()`, `minute()`, `second()`, `millisecond()`
  - Conversion functions: `to_date()`, `to_timestamp()`
  - Current time functions: `now()`, `current_date()`, `current_timestamp()`
  - Manipulation functions: `date_trunc()`, `date_add()`, `date_sub()`, `timestamp_add()`, `date_format()`, `datediff()`, `timestamp_diff`
- Updated type inference and conversion logic to properly handle date and timestamp types
- Modified ingestion coercions to preserve date/time types instead of converting to strings
- Support for literals for date & timestamp
- Added comprehensive tests for all new date/time functionality

### Notes

1. The formats used by `date_format`, `to_date` & `to_timestamp` are the "java like formats": (e.g. yyyy-MM-dd), polars uses the "chrono" time formats (Unix like: %Y-%m-%d). The code translates the format into chrono. This is also why for these functions I save the original format (to make sure that it can be serialized properly).
2. timestamp_diff rounds to int, in the case of year & month as these resolutions are not available in the duration class. The result of `typestamp_diff` is an Int64.
3. `current_date()` & `current_timestamp()` are just calls to the `NowExpr()`
4. Expression serialization is implemented for all the newly added expressions
5. To implement `timestamp_add`, I'm using `dt.offset_by()`, this requires passing the units as 1mo, 1y, etc. The interface exposed by fenic follows pyspark, so we have a quantity and a unit. Those are translated into what is required by `dt.offset_by()`

### Still to be done
1. Need to update the metrics to use proper timestamps


german-typedef:
* **#241** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/241?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#232** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/232?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#201** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/201?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/201?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/fenic/_backends/local/transpiler/expr_converter.py
```diff
@@ -1217,6 +1234,109 @@ def _convert_greatest_expr(self, logical: GreatestExpr) -> pl.Expr:
...
+    def _convert_date_diff_expr(self, logical: DateDiffExpr) -> pl.Expr:
+        return (self._convert_expr(logical.expr1) - self._convert_expr(logical.expr2)).dt.total_days()
+
+    @_convert_expr.register(TimestampDiffExpr)
+    def _convert_timestamp_diff_expr(self, logical: TimestampDiffExpr) -> pl.Expr:
+        duration_expr = self._convert_expr(logical.expr1) - self._convert_expr(logical.expr2)
+        if logical.unit == "year":
+            return (duration_expr.dt.total_days() / 365).cast(pl.Int64)
+        elif logical.unit == "month":
+            return (duration_expr.dt.total_days() / 30).cast(pl.Int64)
```
bcallender, line 1344: could this be done using `pl.date_range`? it takes into account things like calendar variations and DST, so something like 

```python
pl.date_range(
        start=self._convert_expr(logical.expr1),
        end=self._convert_expr(logical.expr2),
        interval="1mo",
        closed='left'
    )
```

bcallender: `time_range` also exists, if we don't have any date components (just a epoch timestamp)

src/fenic/__init__.py
```diff
@@ -75,6 +75,7 @@
     DataLikeType,
```
YoungVor, line 79: can we add the dt functions here too, so people can do fenic.year(col), fenic.to_date(col), etc without needing to import each function?

src/fenic/core/_logical_plan/expressions/dt.py
```diff
@@ -0,0 +1,550 @@
...
+    def to_column_field(self, plan: LogicalPlan, session_state: BaseSessionState) -> ColumnField:
+        col_expr = self.expr.to_column_field(plan, session_state)
+        if col_expr.data_type != StringType:
+            raise TypeError(f"Expected StringType, got {col_expr.data_type}")
+        return ColumnField(str(self), self.get_return_type())
+
+    @classmethod
+    def get_default_format(cls) -> str:
+        """Returns the default chrono format for the expression."""
+        raise NotImplementedError(f"get_default_format not implemented for {type(cls)}")
```
YoungVor: nit: i think you could also declare it @ abstractmethod and pass here

YoungVor: I'm impressed at the guy who made his git handle @ abstractmethod, he must get a lot of pings on random python projects :D

src/fenic/api/functions/dt.py
```diff
@@ -0,0 +1,492 @@
...
+    """Extract the year from a date column.
+
+    Args:
+        column: The column to extract the year from.
+
+    Returns:
+        A Column object with the year extracted.
+
+    Raises:
+        TypeError: If column type is not a DateType or TimestampType.
```
YoungVor, line 44: With a future PR we'll be able to cast a float or formatted strings to date/timestamp, right?

src/fenic/api/functions/dt.py
```diff
@@ -0,0 +1,492 @@
...
+
+    Returns:
+        A Column object with the month extracted.
+
+    Raises:
+        TypeError: If column type is not a DateType or TimestampType.
+
+    Example:
+        ```python
+        # dates: "2025-01-01", "2025-01-02", "2025-01-03"]
```
YoungVor: nit: might be nice for the example to span 2024-12-31 to 2025-01-01, so its more obvious 01 is the month

src/fenic/api/functions/dt.py
```diff
@@ -0,0 +1,492 @@
...
+    Returns:
+        A Column object with the DateType transformed.
+
+    Raises:
+        TypeError: If column type is not a StringType.
+
+    Notes:
+        - If format is not provided, the default format is "YYYY-MM-DD".
+        - The accepted formats should follow this pattern:
+          https://spark.apache.org/docs/latest/sql-ref-datetime-pattern.html
```
YoungVor, line 216: this would be the first time we directly reference spark documentation i think.  I know why you're doing it- that's a pretty complex feature to describe.

One issue though - we don't accept every pattern in that link, right?  We're staying away from timezones for e.g.

We could keep this more vague, provide a few more examples that show the different rules of the pattern, and say that it follows spark-like datetime patterns.. wdyt?  Or we could just say what we don't support in that link

src/fenic/api/functions/dt.py
```diff
@@ -0,0 +1,492 @@
...
+    Returns:
+        A Column object with the TimestampType transformed.
+
+    Raises:
+        TypeError: If column type is not a StringType.
+
+    Notes:
+        - If format is not provided, the default format is ISO 8601 with milliseconds.
+        - The accepted formats should follow this pattern:
+          https://spark.apache.org/docs/latest/sql-ref-datetime-pattern.html
```
YoungVor, line 246: same concern as above

src/fenic/api/functions/dt.py
```diff
@@ -0,0 +1,492 @@
...
+        ```
+    """
+    return Column._from_logical_expr(ToTimestampExpr(Column._from_col_or_name(column)._logical_expr, format))
+
+
+def now() -> Column:
+    """Get the current date and time.
+    
+    Returns:
+        A Column object with the current date and time.
```
YoungVor, line 264: should include the type.  returns type timestamp, iiuc?

src/fenic/api/functions/dt.py
```diff
@@ -0,0 +1,492 @@
...
+        # Output: [{'date': '<current date and time>'}]
+        ```
+    """
+    return Column._from_logical_expr(NowExpr())
+
+def current_date() -> Column:
+    """Get the current date.
+
+    Returns:
+        A Column object with the current date.
```
YoungVor, line 294: same for these.  This one is DateType?

src/fenic/api/functions/dt.py
```diff
@@ -0,0 +1,492 @@
...
+        df.select(date_trunc(col("date"), "year")).to_dicts()
+        # Output: [{'date': '2025-01-01'}, {'date': '2025-01-01'}, {'date': '2025-01-01'}]
+        ```
+    """
+    if unit not in ["year", "month", "day", "hour", "minute", "second", "millisecond"]:
+        raise ValueError(f"Unsupported unit: {unit}. Supported units are: year, month, day, hour, minute, second, millisecond.")
+
+    return Column._from_logical_expr(DateTruncExpr(Column._from_col_or_name(column)._logical_expr, unit.lower()))
+
+def date_add(column: ColumnOrName, days: Union[int, ColumnOrName]) -> Column:
```
YoungVor, line 333: are you not including the validate_call here so they can do types that can cast to int?  You could try out something like conint(strict=False) from pydantic to keep strict checking but allow int-type values for that field

YoungVor: it would be nice to keep the strict checking for the other fields

src/fenic/api/functions/dt.py
```diff
@@ -0,0 +1,492 @@
...
+        ```
+    """
+    if isinstance(quantity, int):
+        quantity_expr = LiteralExpr(quantity, IntegerType)
+    else:
+        quantity_expr = Column._from_col_or_name(quantity)._logical_expr
+
+    return Column._from_logical_expr(TimestampAddExpr(Column._from_col_or_name(column)._logical_expr, quantity_expr, unit.lower()))
+
+def date_format(column: ColumnOrName, format: str) -> Column:
```
YoungVor, line 422: strict validate_call, here and the other functions

tests/_backends/local/functions/test_date_functions.py
```diff
@@ -0,0 +1,367 @@
...
+class TestDateFunctions:
+    def test_year(self, df_with_date_types):
+        """Test the year function."""
+        default_year_results = [2025]*3
+        self._verify_results(df_with_date_types, "ts", year, IntegerType, "year", default_year_results)
+        self._verify_results(df_with_date_types, "date", year, IntegerType, "year", default_year_results)
+        self._verify_error(df_with_date_types, "col1", year, TypeError)
+
+    def test_month(self, df_with_date_types):
+        """Test the month function."""
```
YoungVor, line 151: nice and clean with the helper methods!

src/fenic/api/functions/dt.py
```diff
@@ -0,0 +1,492 @@
...
+    Example:
+        ```python
+        df.select(current_date()).to_dicts()
+        # Output: [{'date': '<current date>'}]
+        ```
+    """
+    return Column._from_logical_expr(NowExpr(as_date=True))
+
+@validate_call(config=ConfigDict(strict=True, arbitrary_types_allowed=True))
+def date_trunc(column: ColumnOrName, unit: str) -> Column:
```
YoungVor: can we make unit a Literal["year", "month", "day", "hour", "minute", "second", "millisecond"]? then with validate call you shouldn't need to validate outside of the function call sites

src/fenic/core/_logical_plan/expressions/dt.py
```diff
@@ -0,0 +1,550 @@
...
+
+    @staticmethod
+    def _get_polars_date_trunk_unit(unit: str) -> str:
+        """Returns the Polars date trunk unit for a given unit.
+        
+        For more information checkout the Polars documentation:
+        https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.dt.truncate.html
+        """
+        if unit not in POLARS_DURATION_UNITS_MAP:
+            raise ValueError(f"Unsupported unit: {unit}. Supported units are: year, month, day, hour, minute, second, millisecond.")
```
YoungVor: nit: can just print a list of the POLARS_DURATION_UNITS_MAP.keys().

src/fenic/core/_utils/schema.py
```diff
@@ -241,11 +252,13 @@ def _convert_polars_dtype_to_custom_dtype(
             for field in polars_dtype.fields
         ]
         return StructType(struct_fields=struct_fields)
-    elif isinstance(polars_dtype, (pl.Date, pl.Datetime)):
+    elif isinstance(polars_dtype, pl.Date):
+        return DateType
+    elif isinstance(polars_dtype, pl.Datetime):
         # pl.Time is not properly supported by DuckDB, it is suggested that users
```
YoungVor: does this comment still make sense?

YoungVor (commented):
Finished most of a pass, will finish a pass tomorrow or monday.  Looks great, thank you for doing this!  I have a few nits and some suggestions around the api

will writing to/from duckdb work and testing be a future pr?  And do you expect any work for remote catalogs/arrow to work?

src/fenic/_backends/local/physical_plan/utils.py
```diff
@@ -38,9 +36,7 @@ def apply_ingestion_coercions(df: pl.DataFrame) -> pl.DataFrame:
 
 
 def _build_target_dtype(dtype: pl.DataType) -> pl.DataType:
-    if dtype in [pl.Date, pl.Datetime]:
-        return pl.String
-    elif isinstance(dtype, pl.Array):
+    if isinstance(dtype, pl.Array):
```
YoungVor, line 38: is this just a noop now?  Can we just return the arg?  or am i missing a nuance with Array vs List

src/fenic/core/_logical_plan/expressions/dt.py
```diff
@@ -0,0 +1,550 @@
...
+
+
+# DATE CONVERSION EXPRESSIONS
+
+
+class DateConversionExpr(LogicalExpr):
+    """Base expression for date conversion operations."""
+    def __init__(self, expr: LogicalExpr, format: Optional[str]):
+        self.expr = expr
+        self.original_format = format
```
YoungVor, line 141: nit: is the original format always what the user specifies, and is it always guaranteed to be the spark/java format?

src/fenic/api/functions/dt.py
german-typedef: Correct, working on it right now.

src/fenic/__init__.py
german-typedef: How about adding the `dt `module, so we can do: `fenic.dt.year `(and all the other functions), similar to what we do with embeddings?

src/fenic/core/_logical_plan/expressions/dt.py
german-typedef: Awesome! Will change in the next iteration.

src/fenic/api/functions/dt.py
german-typedef: I like it. Agreed. Done in the next iteration.

src/fenic/api/functions/dt.py
german-typedef: I like the idea of just specifying that we don't support timezones (symbols: V, x, O, X, Z)

src/fenic/api/functions/dt.py
german-typedef: Added in the next iteration.

src/fenic/api/functions/dt.py
german-typedef: Correct, updated it accordingly.

src/fenic/core/_logical_plan/expressions/dt.py
german-typedef: You are right! Done in the next iteration.

src/fenic/core/_utils/schema.py
german-typedef: Removed in the next iteration.

src/fenic/_backends/local/physical_plan/utils.py
german-typedef: Let me double check why we needed to do that.

src/fenic/core/_logical_plan/expressions/dt.py
german-typedef: Correct, it is the one user provides and it should be the spark/java format.

src/fenic/_backends/local/transpiler/expr_converter.py
german-typedef: I looked at date_range. It works somewhat different from what we need, date_range will generate a range of dates (and also there is datetime_range) from the start to the end using the interval (start must be just one value, and end must be just one value), so if you have:
pl.date_range(start=2025-01-01, end=2025-01-05, interval="1d", closed="left') -> [2025-01-01, 2025-01-02, 2025-01-03, 2025-01-04] instead of "4" days.
I looked at the Spark documentation, timestamp_diff is not fully accurate, as it just checks for year and month that that more than a year or more than 30 days have passed). For year, it doesn't check for leap years, also for months it doesn't check the length of the month. For the next set of functions we have months_between that it should be precise (I'll need to figure out that one.

src/fenic/api/functions/dt.py
german-typedef: Updated.

src/fenic/api/functions/dt.py
german-typedef: Done, makes sense!

src/fenic/api/functions/dt.py
german-typedef: Added.

src/fenic/_backends/local/transpiler/expr_converter.py
bcallender: Ah yeah, I forgot to add that we would take the length of the list generated by `pl.date_range` -- in that example the length of the generated list is how we get the 4 days. If we don't need that precision here, makes sense to not do it for now, but might be useful for `months_between`

src/fenic/_backends/local/physical_plan/utils.py
bcallender: iirc, ideally we'd want to keep polars arrays as arrays so they keep their optimizations related to being a fixed size. there was some reason why we couldn't for now, hence this check.

src/fenic/core/_logical_plan/expressions/dt.py
```diff
@@ -0,0 +1,552 @@
...
+)
+
+POLARS_DURATION_UNITS_MAP = {
+    "year": "y",
+    "month": "mo",
+    "day": "d",
+    "hour": "h",
+    "minute": "m",
+    "second": "s",
+    "millisecond": "m",
```
bcallender: is this supposed to be the same as minute?

src/fenic/api/functions/dt.py
german-typedef: I missed :). Fixed.

src/fenic/core/_logical_plan/expressions/dt.py
german-typedef: Nope! great catch! Just a few cases to make sure all the paths are exercised. Thank you!

src/fenic/api/functions/dt.py
german-typedef: You will be able to cast an int or a string to a date / timestamp and viceversa.
Since dates are # of day since the epoch, and timestamps are the # of ms since epoch, int casting should be safe.

src/fenic/core/_logical_plan/expressions/dt.py
```diff
@@ -0,0 +1,552 @@
...
+        self.is_date = _check_date_type(self.expr, plan, session_state)
+
+        days_col = self.days.to_column_field(plan, session_state)
+        if days_col.data_type != IntegerType:
+            raise TypeError(f"Expected IntegerType, got {days_col.data_type}")
+
+        return ColumnField(str(self), DateType if self.is_date else TimestampType)
+    
+    def children(self) -> List[LogicalExpr]:
+        return []
```
cpard: is this correct? Shouldn't children return non empty values?

german-typedef: thank you! no, should be the expressions. I caught a couple of other places like that.

src/fenic/__init__.py
YoungVor: Perfect.  Works well with the short 'dt' module name!

src/fenic/api/functions/dt.py
```diff
@@ -0,0 +1,499 @@
...
+        A Column object with the DateType transformed.
+
+    Raises:
+        TypeError: If column type is not a StringType.
+
+    Notes:
+        - If format is not provided, the default format is "YYYY-MM-DD".
+        - The accepted formats should follow this pattern:
+          https://spark.apache.org/docs/latest/sql-ref-datetime-pattern.html
+        - Also keep in mind that time zones are not supported, so the following symbols won't work:
```
YoungVor: Nit: prefer wording in the documentation like "Time zones are not supported.  The following..."  or "Note: Time zones are not..."

src/fenic/api/functions/dt.py
```diff
@@ -0,0 +1,499 @@
...
+    """
+    if isinstance(days, int):
+        days_expr = LiteralExpr(days, IntegerType)
+    else:
+        days_expr = Column._from_col_or_name(days)._logical_expr
+
+    return Column._from_logical_expr(DateAddExpr(Column._from_col_or_name(column)._logical_expr, days_expr, sub=True))
+
+@validate_call(config=ConfigDict(strict=True, arbitrary_types_allowed=True))
+def timestamp_add(column: ColumnOrName, quantity: Union[int, ColumnOrName], unit: Literal["year", "month", "day", "hour", "minute", "second", "millisecond"]) -> Column:
```
YoungVor: Nit: can define the Literal as a named type at the top of the file, since you use it a few times

src/fenic/api/functions/dt.py
```diff
@@ -0,0 +1,499 @@
...
+        # Output: [{'date_diff': -1}, {'date_diff': 1}, {'date_diff': 4}]
+        ```
+    """
+    return Column._from_logical_expr(
+        DateDiffExpr(
+            Column._from_col_or_name(end)._logical_expr,
+            Column._from_col_or_name(start)._logical_expr))
+
+@validate_call(config=ConfigDict(strict=True, arbitrary_types_allowed=True))
+def timestamp_diff(start: ColumnOrName, end: ColumnOrName, unit: Literal["year", "month", "day", "hour", "minute", "second", "millisecond"]) -> Column:
```
YoungVor: nice, start/end is much more clear!

src/fenic/core/_logical_plan/expressions/dt.py
```diff
@@ -0,0 +1,552 @@
...
+
+
+def _check_date_type(expr: LogicalExpr, plan: LogicalPlan, session_state: BaseSessionState) -> bool:
+    """Checks if the expression is a date or timestamp type.
+
+    Returns True if the expression is a date type, False if it is a timestamp type.
+    Raises an error if the expression is not a date or timestamp type.
+    """
+    col_field = expr.to_column_field(plan, session_state)
+    if col_field.data_type in [DateType, TimestampType]:
```
YoungVor, line 383: nit: can just do a if, elif, else

tests/_backends/local/functions/test_date_functions.py
```diff
@@ -0,0 +1,401 @@
...
+    )
+
+    pl_df = pl_df.with_columns(
+        pl.col("ts_str").str.strptime(pl.Date, format="%Y-%m-%dT%H:%M:%S.%3f").alias("dt")
+    )
+
+    return local_session.create_dataframe(pl_df)
+
+class TestDateFunctions:
+    def test_year(self, df_with_date_types):
```
YoungVor, line 140: could we have a test or two with nested date and timestamp expressions, where you passs expressions like now() or dateAdd to another date function like date diff?

src/fenic/core/_logical_plan/expressions/dt.py
YoungVor: would be good to add some nested expression tests too

src/fenic/api/functions/dt.py
german-typedef: I'm making a change here to avoid casting to boolean (as it will be 1, so you'll get something like January 2nd 1970 that semantically doesn't make much sense).

src/fenic/core/_logical_plan/expressions/dt.py
german-typedef: I'm adding them in the next iteration.

src/fenic/core/_logical_plan/expressions/dt.py
```diff
@@ -0,0 +1,552 @@
...
+    def children(self) -> List[LogicalExpr]:
+        return [self.start, self.end]
+
+    def _eq_specific(self, other: TimestampDiffExpr) -> bool:
+        return self.start == other.start and self.end == other.end and self.unit == other.unit
+
+# UTILITY FUNCTIONS
+
+
+def _check_date_type(expr: LogicalExpr, plan: LogicalPlan, session_state: BaseSessionState) -> bool:
```
rohitrastogi: the semantics of this function are a bit odd:
The name '_check_date_type' makes the function seem that it just validates, but instead its encoding "True" for DateType, "False" for TimestampType

Would consider something more like:
```
def resolve_temporal_type(expr: LogicalExpr, plan: LogicalPlan, session_state: BaseSessionState) -> DataType:
    """Resolve the expression's type if it's DATE or TIMESTAMP, otherwise raise."""
    col_field = expr.to_column_field(plan, session_state)
    if col_field.data_type in (DateType, TimestampType):
        return col_field.data_type
    raise TypeError(f"Expected DateType or TimestampType, got {col_field.data_type}")
 ```

Comment (german-typedef) [reply src/fenic/core/_logical_plan/expressions/dt.py]: I like the suggestion. Modified accordingly.

File: src/fenic/core/_logical_plan/expressions/dt.py (hunk 23 above)
Comment (german-typedef) [reply src/fenic/core/_logical_plan/expressions/dt.py:383]: Updated, see Rohit's suggestion.

File: src/fenic/api/functions/dt.py (hunk 21 above)
Comment (german-typedef) [reply src/fenic/api/functions/dt.py]: Done.

File: src/fenic/api/functions/dt.py (hunk 20 above)
Comment (german-typedef) [reply src/fenic/api/functions/dt.py]: I like it, updated.

File: tests/_backends/local/functions/test_date_functions.py (hunk 24 above)
Comment (german-typedef) [reply tests/_backends/local/functions/test_date_functions.py:140]: please check the new method I've added to the TestDateFunctions class: test_nested_datetime_expressions.

File: src/fenic/api/column.py [hunk 26]
```diff
@@ -204,8 +204,12 @@ def cast(self, data_type: DataType) -> Column:
 
         - Numeric types (IntegerType, FloatType, DoubleType) can be cast between each other
         - Numeric types can be cast to/from StringType
+        - DateType and TimestampType can be cast between each other
+        - DateType and TimestampType can be cast to/from numeric types andStringType
```
rohitrastogi: nit: typo: `and StringType`

Generally speaking, I wonder if this information can be best represented as a table (n x n - where n is the number of types)

src/fenic/api/functions/dt.py
```diff
@@ -0,0 +1,500 @@
...
+    Returns:
+        A Column object with the year extracted.
+
+    Raises:
+        TypeError: If column type is not a DateType or TimestampType.
+
+    Example:
+        ```python
+        # dates: "2025-01-01", "2025-01-02", "2025-01-03"]
+        df.select(year(col("date"))).to_dicts()
```
rohitrastogi: here and elsewhere in this file, we probably want to show off the dt namespace
`df.select(dt.year...)`

I think we (try) to do this for the other other namespaces.

src/fenic/api/functions/dt.py
```diff
@@ -0,0 +1,500 @@
...
+    Returns:
+        A Column object with the date/timestamp column with the days substracted.
+
+    Raises:
+        TypeError: If column type is not a DateType or TimestampType, or if days is not an IntegerType.
+
+    Example:
+        ```python
+        # dates: "2025-01-01", "2025-02-01", "2025-03-01"]
+        df.select(date_sub(col("date"), 1)).to_dicts()
```
rohitrastogi: (here and elsewhere) we don't have a method called to_dicts(). It's called to_pydict().

src/fenic/core/_logical_plan/expressions/dt.py
```diff
@@ -0,0 +1,545 @@
...
+        pass
+
+
+class ToDateExpr(DateConversionExpr):
+    """Parses a string to a date."""
+    def __init__(self, expr: LogicalExpr, format: Optional[str]):
+        super().__init__(expr, format)
+
+    def __str__(self):
+        return f"TO_DATE({self.expr})"
```
rohitrastogi: we have conventionally (tried) to use "namespace.{function_lower_cased"}
so, `dt.to_date` in this case (applies here and elsewhere)

src/fenic/core/_utils/type_inference.py
```diff
@@ -1,9 +1,11 @@
+import datetime
```
rohitrastogi, line 1: Can you add some unit tests to tests/_backends/local/functions/test_lit.py to validate whether lit() works with the python datetime objects?

tests/_backends/test_ingestion_coercion.py
```diff
@@ -6,7 +6,7 @@
 
 
 def test_simple_date_coercion():
-    """Date column should be converted to string."""
+    """Date column should not beconverted to string."""
```
rohitrastogi: nit: typo

rohitrastogi: also, since we aren't coercing anymore, do we need these tests in this file?

src/fenic/api/column.py
german-typedef: Agreed, I think we can do something like this:
https://learn.microsoft.com/en-us/sql/t-sql/data-types/media/cast-and-convert-transact-sql/data-type-conversions.png?view=sql-server-ver17

I'll work on this to update the docs, probably outside of this PR.

src/fenic/api/functions/dt.py
german-typedef: Makes it clearer. Done.

src/fenic/api/functions/dt.py
german-typedef: Fixed.

src/fenic/core/_logical_plan/expressions/dt.py
german-typedef: Thanks! Updated for all the dt expressions.

src/fenic/core/_utils/type_inference.py
german-typedef: Done, I've added them to the lit primitive tests, we have others in different areas, but good to make sure the lit tests have an explicit validation for this.

tests/_backends/test_ingestion_coercion.py
german-typedef: I removed the test from this but added one to the test_date_functions to make sure we don't coerce any longer.

src/fenic/core/_logical_plan/expressions/dt.py
```diff
@@ -0,0 +1,545 @@
...
+        return [self.expr]
+
+
+class YearExpr(DatePartRelatedExpr):
+    """Expression for extracting the year from a date type column."""
+    def __init__(self, expr: LogicalExpr):
+        super().__init__(expr)
+
+    def __str__(self):
+        return f"dt.year({self.expr})"
```
YoungVor, line 55: nice, I like the str representations matching the namespace and function

src/fenic/_backends/local/transpiler/expr_converter.py
```diff
@@ -1217,6 +1235,109 @@ def _convert_greatest_expr(self, logical: GreatestExpr) -> pl.Expr:
...
+    def _convert_milli_second_expr(self, logical: MilliSecondExpr) -> pl.Expr:
+        converted_expr = self._convert_expr(logical.expr)
+        if logical.temporal_type == DateType:
+            return _get_zero_ts_part_from_date(converted_expr)
+        return converted_expr.dt.millisecond()
+
+    @_convert_expr.register(ToDateExpr)
+    def _convert_to_date_expr(self, logical: ToDateExpr) -> pl.Expr:
+        """Convert a string into a date."""
+        # logical.format is should be a chrono format.
```
rohitrastogi: nit: typo
remove "is"

src/fenic/api/functions/dt.py
```diff
@@ -0,0 +1,500 @@
...
+    SecondExpr,
+    TimestampAddExpr,
+    TimestampDiffExpr,
+    ToDateExpr,
+    ToTimestampExpr,
+    YearExpr,
+)
+from fenic.core.types.datatypes import IntegerType
+
+LITERAL_TIME_UNITS = Literal["year", "month", "day", "hour", "minute", "second", "millisecond"]
```
rohitrastogi: please move this to core/types/enums.py and follow the camel case naming convention.

Also, please make the the expressions accept the enum instead of a string for better type safety. Some embedding expressions and the fuzzy string expressions follow this pattern.

tests/_backends/local/functions/test_date_functions.py
```diff
@@ -0,0 +1,562 @@
...
+        assert result.schema["int_col"] == pl.Int64  # unchanged
+
+        # Check actual values are string representations
+        date_values = result["date_col"].to_list()
+        datetime_values = result["datetime_col"].to_list()
+        assert all(isinstance(val, datetime.date) for val in date_values)
+        assert all(isinstance(val, datetime.datetime) for val in datetime_values)
+
+    @classmethod
+    def _verify_results(
```
rohitrastogi, line 555: nit: I see how this utility is useful as the patterns are definitely the same for each test, but I find this reduces the readability of the tests. I tend to view the tests as documentation for examples of how the API works and find this challenging to grok on first glance.

german-typedef: Makes sense, for each of the tests that only had the helper methods calls, I've added a case with the specific use for that function, so it is clear on how it is supposed to be used, while keeping it concise when we have several calls for the same function by using the helper methods.

src/fenic/api/functions/dt.py
german-typedef: Done. Moved to the enums, and used it in the functions as well as in the expressions.

src/fenic/_backends/local/transpiler/expr_converter.py
german-typedef: Removed here and from the next function as well.

protos/logical_plan/v1/expressions.proto
```diff
@@ -524,4 +547,79 @@ message LeastExpr {
...
+message DateAddExpr {
+  LogicalExpr expr = 1;
+  LogicalExpr days = 2;
+  bool sub = 3;
+}
+
+message TimestampAddExpr {
+  LogicalExpr expr = 1;
+  LogicalExpr quantity = 2;
+  string unit = 3;
```
rohitrastogi, line 597: probably ideal to make this a proto enum (though we do this inconsistently right now)

PR #202, opened 2025-09-12
fix: when querying files in public S3 buckets/HuggingFace Datasets, the user should not have to provide credentials

bcallender:
### TL;DR

Allow users to pull public datasets from Hugging Face and publicaly accessible files from S3 without setting up an HF_TOKEN/AWS_ACCESS_KEY. 

### What changed?

- Enhanced error handling in `query_files` and `write_file` functions to provide more specific error messages
- Modified credential handling functions to return a boolean indicating if credentials were found
- Added graceful fallback for HuggingFace paths without credentials (will attempt to read public datasets)
- Added warning logs when credentials are missing but operation might still succeed
- Added error handling for HTTP status codes (404, 401, 403) with contextual messages
- Updated tests to match the new error handling behavior

### How to test?

- Test reading from S3 with valid credentials, invalid credentials, and no credentials
- Test reading from HuggingFace with valid token, invalid token, and no token
- Test writing to S3 with and without credentials
- Verify error messages are clear and actionable for each failure case

### Why make this change?

The previous implementation would fail with generic or confusing error messages when credential issues occurred. This change provides more user-friendly error messages that explain exactly what went wrong and how to fix it, improving the developer experience when working with remote data sources. It also allows reading from public HuggingFace datasets without requiring a token.

### DIFF
--- a/tests/_backends/local/io/test_reader.py
+++ b/tests/_backends/local/io/test_reader.py
@@ -42,6 +42,7 @@ from tests.conftest import _save_pdf_file
 
 COLUMNS = {"name", "age", "city"}
 
+
 def write_test_file(
     path: str,
     content: Union[pl.DataFrame, str],
@@ -90,7 +91,7 @@ David,33,Seattle"""
     schema = df.schema
     actual_columns = {field.name for field in schema.column_fields}
     assert (
-        actual_columns == COLUMNS
+            actual_columns == COLUMNS
     ), f"Expected columns {COLUMNS}, got {actual_columns}"
 
     # Collect the full DataFrame and verify data
@@ -110,7 +111,7 @@ David,33,Seattle"""
     schema = df.schema
     actual_columns = {field.name for field in schema.column_fields}
     assert (
-        actual_columns == COLUMNS
+            actual_columns == COLUMNS
     ), f"Expected columns {COLUMNS}, got {actual_columns}"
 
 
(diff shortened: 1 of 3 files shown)


bcallender:
* **#202** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/202?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/202?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

bcallender:
i removed the extra tests, i didn't realize they were already there, just gated. should be correct now.

tests/_backends/local/io/test_reader.py
```diff
@@ -9,7 +9,6 @@
 import polars as pl
 import pytest
 from botocore.session import get_session
-from duckdb import HTTPException
```
YoungVor, line 12: please make sure to run test_reader tests with `--test-huggingface-reads 1` and again with `--test-output-dir s3://<our public s3 bucket>/some_tmp_path` args

bcallender: (finally) done. the issue we were seeing was because when the `test_read_query_setup_with_huggingface_credentials_and_s3_credentials` test ran, it created a session using `get_or_create`, but never closed it. there is some...weird behavior that happens if you do that -- the next test that runs sets up the `local_session` using the fixture: 

```
@pytest.fixture
def local_session(local_session_config, request):
  ...
    session = Session.get_or_create(local_session_config)
    yield session
    session.stop()
    ...
``` 

but, since the `LocalSessionManager` is presumably still a singleton somewhere in memory, when that fixture runs, it just picks up the same session state it just had, with those test environment variables injected in the `BotoSession`. when the test fails, the session gets stopped, which is why the behavior is "fixed" for all the other tests in the suite.

PR #208, opened 2025-09-15
fix: remove o1-mini since it doesn't support json_schema structured output

YoungVor:
## What changed?

o1-mini does not support json_schema or structured outputs in its API, which we depend on for our semantic operation structured responses.


YoungVor:
* **#208** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/208?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/208?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

bcallender (approved):
we could probably finagle our way around the structured output support by failing if the model doesn't support it at runtime (like we do for openrouter), but dealing with the lack of things like system messages is definitely non-trivial. plus, this model will be retired soon anyway, and is already deprecated.

YoungVor:
> we could probably finagle our way around the structured output support by failing if the model doesn't support it at runtime (like we do for openrouter), but dealing with the lack of things like system messages is definitely non-trivial. plus, this model will be retired soon anyway, and is already deprecated.

True, and if we just fail when trying to provide a pydantic output, that definitely reduces how useful the model is, (a point in favor of just not supporting a deprecated model)

PR #209, opened 2025-09-15
fix: gemini-2.5-pro default model without profiles

YoungVor:
## Why the change

Gemini-2.5-pro does not allow disabled reasoning.  We had special checks for the Profile path, but we weren't handling the default model path with no profiles properly.

With this change, user can configure gemini-2.5-pro in a single line without profiles and we default the thinking budget to 'low'.  Also introduced concept of ‘supports-disabled-reasoning’ to our model-params to remove special case code

## How to test

Run unit tests with gemini-2.5-pro and removing all profiles to ensure default behavior works

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -335,18 +335,20 @@ def configure_language_model(model_provider: ModelProvider, model_name: str) -> 
             )
     elif model_provider == ModelProvider.GOOGLE_DEVELOPER:
         if model_parameters.supports_reasoning:
+            profiles = {
+                "auto": GoogleDeveloperLanguageModel.Profile(thinking_token_budget=-1),
+                "low": GoogleDeveloperLanguageModel.Profile(thinking_token_budget=1024),
+                "medium": GoogleDeveloperLanguageModel.Profile(thinking_token_budget=4096),
+                "high": GoogleDeveloperLanguageModel.Profile(thinking_token_budget=8192),
+            }
+            if model_parameters.supports_disabled_reasoning:
+                profiles["thinking_disabled"] = GoogleDeveloperLanguageModel.Profile()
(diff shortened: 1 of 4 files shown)


YoungVor:
* **#209** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/209?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/209?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tests/conftest.py
```diff
@@ -335,17 +335,19 @@ def configure_language_model(model_provider: ModelProvider, model_name: str) ->
...
                 tpm=500_000,
-                profiles={
-                    "thinking_disabled": GoogleDeveloperLanguageModel.Profile(),
-                    "auto": GoogleDeveloperLanguageModel.Profile(thinking_token_budget=-1),
-                    "low": GoogleDeveloperLanguageModel.Profile(thinking_token_budget=1024),
-                    "medium": GoogleDeveloperLanguageModel.Profile(thinking_token_budget=4096),
-                    "high": GoogleDeveloperLanguageModel.Profile(thinking_token_budget=8192),
-                },
+                profiles=profiles,
                 default_profile="auto",
```
bcallender: can this be `"thinking_disabled" if model_parameters.supports_disabled_reasoning else "auto"`? As much as possible we want to disable generation of reasoning tokens for the unit tests if we are testing a model like `gemini-2.5-flash`

bcallender (approved):

tests/conftest.py
YoungVor: ah, good point.  Done

PR #210, opened 2025-09-15
fix: raise if a request can't be handled by max token capacity

YoungVor:
## Current behavior

- configuring tpm to something low will result in the program hanging forever if a single request has a higher token estimate than the max capacity.  
- when model_client process_queue caught an exception, the client stops processing the tasks, but nothing cancels the pending tasks so we never shut down.
- google models were retrying on catch-all errors

## What changed

- On detecting max tpm too low for request's estimated tokens, cancel model client tasks queue and raise error
- When _process_queue catches a non-cancel exception, clear the queue and shutdown
- google models  fail (raise FatalException) in catch-all error handling

### DIFF
--- a/src/fenic/_inference/google/gemini_native_chat_completions_client.py
+++ b/src/fenic/_inference/google/gemini_native_chat_completions_client.py
@@ -350,7 +350,7 @@ class GeminiNativeChatCompletionsClient(
             else:
                 return FatalException(e)
         except Exception as e:  # noqa: BLE001 – catch-all mapped to Fatal
-            return TransientException(e)
+            return FatalException(e)
 
     def _prepare_schema(self, response_format: ResolvedResponseFormat) -> dict[str, Any]:
         """Google Gemini does not support additionalProperties in JSON schemas, even if it is set to False.
--- a/src/fenic/_inference/model_client.py
+++ b/src/fenic/_inference/model_client.py
@@ -283,7 +283,6 @@ class ModelClient(Generic[RequestT, ResponseT], ABC):
                 self._register_thread_exception(queue_item, exception)
             except asyncio.QueueEmpty:
                 break
-
         cancel_future = asyncio.run_coroutine_threadsafe(
             self._cancel_in_flight_requests(), self._event_loop
         )
@@ -562,6 +561,12 @@ class ModelClient(Generic[RequestT, ResponseT], ABC):
             raise
         except Exception as e:
             logger.error(f"Error in worker for model {self.model}: {e}", exc_info=True)
+            for task in self.pending_requests:
+                self._register_thread_exception(task, e)
+            self.pending_requests = []
+            self.request_queue.cancel()
+            self.retry_queue.cancel()
+            self.shutdown()
 
 
     async def _process_single_request(self, queue_item: QueueItem[RequestT]):
--- a/src/fenic/_inference/rate_limit_strategy.py
+++ b/src/fenic/_inference/rate_limit_strategy.py
@@ -5,6 +5,7 @@ from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
 
 from fenic._constants import MINUTE_IN_SECONDS
+from fenic.core.error import ExecutionError
 
 
 @dataclass
@@ -86,8 +87,12 @@ class RateLimitStrategy(ABC):
 
         Returns:
             bool: True if there was enough capacity and it was consumed, False otherwise.
+
+        Raises:
+            ConfigurationError: If there is insufficient capacity to handle the request.
         """
         pass
+
 
     @abstractmethod
     def context_tokens_per_minute(self) -> int:
@@ -137,6 +142,7 @@ class UnifiedTokenRateLimitStrategy(RateLimitStrategy):
             bool: True if there was enough capacity and it was consumed, False otherwise.
         """
         now = time.time()
+        self._check_max_rate_limits(token_estimate)
         available_tokens = self.unified_tokens_bucket._get_available_capacity(now)
         available_requests = self.requests_bucket._get_available_capacity(now)
         has_request_capacity = available_requests >= 1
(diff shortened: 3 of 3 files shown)


YoungVor:
* **#210** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/210?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/210?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/fenic/_inference/google/gemini_native_chat_completions_client.py
```diff
@@ -368,7 +368,7 @@ async def make_single_request(
             else:
                 return FatalException(e)
         except Exception as e:  # noqa: BLE001 – catch-all mapped to Fatal
-            return TransientException(e)
+            return FatalException(e)
```
bcallender, line 353: what specific exceptions are you catching here that need to be mapped to `FatalException`? I'm wary of this change just because I have seen some weird behavior where transient network issues cause the `geni-ai` client to raise `httpx` exceptions which don't get mapped to `ServerError` or `ClientError`

src/fenic/_inference/rate_limit_strategy.py
```diff
@@ -207,6 +219,8 @@ def check_and_consume_rate_limit(self, token_estimate: TokenEstimate) -> bool:
             bool: True if there was enough capacity and it was consumed, False otherwise.
         """
         now = time.time()
+        breakpoint()
```
bcallender: remove?

src/fenic/_inference/rate_limit_strategy.py
```diff
@@ -223,6 +237,14 @@ def check_and_consume_rate_limit(self, token_estimate: TokenEstimate) -> bool:
             self.requests_bucket._set_capacity(available_requests, now)
         return has_capacity
 
+    def _check_max_rate_limits(self, token_estimate: TokenEstimate):
+        """Checks if the strategy is configured with enough capacity to handle the request.
+        """
+        if self.input_tpm < token_estimate.input_tokens:
+            raise ConfigurationError(f"Insufficient capacity to handle the request. Input TPM limit is {self.input_tpm} but request requires {token_estimate.input_tokens} input tokens.  Please configure the model with more capacity.")
```
bcallender: add language here indicating the request requirements are an estimate.

src/fenic/_inference/rate_limit_strategy.py
```diff
@@ -150,6 +156,12 @@ def check_and_consume_rate_limit(self, token_estimate: TokenEstimate) -> bool:
 
         return has_capacity
 
+    def _check_max_rate_limits(self, token_estimate: TokenEstimate):
+        """Checks if the strategy is configured with enough capacity to handle the request.
+        """
+        if self.tpm < token_estimate.total_tokens:
+            raise ConfigurationError(f"Insufficient capacity to handle the request. TPM limit is {self.tpm} but request requires {token_estimate.total_tokens} tokens.  Please configure the model with more capacity.")
```
bcallender: nit: Probably a `ConfigurationError` wrapped in an `ExecutionError`? I think that it makes sense that any error that is returned after the plan starts executing would be an `ExecutionError`

bcallender (approved):
anything that fixes a hanging issue is a good fix in my book! i would just be careful with the changes to exception handling in the gemini chat completions client -- I think we should probably map most other exceptions to `FatalException` but there are a class of transient network errors that we should catch and map to `TransientException` imo

src/fenic/_inference/google/gemini_native_chat_completions_client.py
YoungVor: i actually didn't catch anything here, i thought i was initially before realizing the exception would be higher up in _process_queue.

I changed this because i thought it was the intention to map non-specific exceptions to Fatal (its what we tell the linter, and its how openai chat completions works).

src/fenic/_inference/rate_limit_strategy.py
YoungVor: good call ;)

src/fenic/_inference/rate_limit_strategy.py
YoungVor: good point, it did appear as execution error anyway.  I'll change it

src/fenic/_inference/rate_limit_strategy.py
YoungVor: will do!

src/fenic/_inference/google/gemini_native_chat_completions_client.py
YoungVor: I'll add a few more retries for google retriable errors like RESOURCE_EXHAUSTED, ServiceUnavailable, InternalServerError, and DeadlineExceeded.

I won't add the same 404 retry we had for openai, we can always add it if they end up having a similar problem with stale cloudflare endpoints

YoungVor: Closing the conversation here: Thinking again, i think we can keep all ServerErrors as retryable.  If we find errors that don't map to ServerError or ClientError, i think we should default to failing.  It looks like those transient 404s will map to ClientError anyway, so we can decide to add a special case for that in future if needed

src/fenic/_inference/model_client.py
```diff
@@ -562,6 +561,12 @@ async def _process_queue(self):
             raise
         except Exception as e:
             logger.error(f"Error in worker for model {self.model}: {e}", exc_info=True)
+            for task in self.pending_requests:
+                self._register_thread_exception(task, e)
+            self.pending_requests = []
```
rohitrastogi, line 566: @YoungVor what was the purpose of this change? if _process_queue fails because of _process_single_request, we raise an exception, register the thread exception and eventually raise an execution error. we shouldn't shutdown - the caller should decide what to do with the execution error.

YoungVor: @rohitrastogi  good catch.  The purpose is to catch and raise if a request's token estimation is too high for the model's total tpm limit, but I wrongly shutdown instead of properly raising the exception in the user thread and continuing to process the queue item.  Added the fix here: https://github.com/typedef-ai/fenic/pull/228

PR #211, opened 2025-09-15
fix: warn if a request's tokens may exceed model's context window limit

YoungVor:
## What changed

- adding a WARN if a request's estimated tokens exceeds the model's max context window size

### DIFF
--- a/src/fenic/_inference/model_client.py
+++ b/src/fenic/_inference/model_client.py
@@ -30,7 +30,7 @@ from fenic._inference.token_counter import (
     TokenCounter,
     Tokenizable,
 )
-from fenic.core._inference.model_catalog import ModelProvider
+from fenic.core._inference.model_catalog import ModelProvider, model_catalog
 from fenic.core._inference.model_provider import ModelProviderClass
 from fenic.core._logical_plan.resolved_types import ResolvedResponseFormat
 from fenic.core.metrics import LMMetrics
@@ -483,6 +483,7 @@ class ModelClient(Generic[RequestT, ResponseT], ABC):
 
                 # Only enqueue if this is a new, unique request
                 if estimated_tokens is not None:
+                    self._check_token_estimate_against_model_limits(estimated_tokens)
                     num_unique_requests += 1
                     total_token_estimate += estimated_tokens
                     queue_item = QueueItem(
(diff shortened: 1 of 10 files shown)


YoungVor:
* **#211** <a href="https://app.graphite.com/github/pr/typedef-ai/fenic/211?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/typedef-ai/fenic/211?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/fenic/_inference/model_client.py
```diff
@@ -484,6 +487,8 @@ def _submit_batch_requests(self,
 
                 # Only enqueue if this is a new, unique request
                 if estimated_tokens is not None:
+                    if self.model_max_context_window_length and estimated_tokens.total_tokens > self.model_max_context_window_length:
```
bcallender: We should add a separate case for when just the input would be too long:

`estimated_tokens.input_tokens > self._model_parameters.content_window_length` -- Advise the user to use the text chunking functionality to break up their inputs into small enough chunks for the context window length.

src/fenic/_inference/model_client.py
```diff
@@ -128,6 +128,9 @@ def __init__(
         self.rate_limit_strategy = rate_limit_strategy
         self.context_tokens_per_minute = rate_limit_strategy.context_tokens_per_minute()
         self.token_counter = token_counter
+        # model limits
+        model_parameters = model_catalog.get_completion_model_parameters(model_provider, model)
```
bcallender: actually, scratch that, this is a little bit harder than it seems. the `ModelClient` is the super class of both `Embedding` and `Completions` clients, so you can't assume that there will be `completion_model_parameters` for the model you're working with, it could just as easily be an embedding model. 

the reason why this issue was tricky was finding an elegant way to build that logic into some place where you have access to the model parameters and the token estimate, but without having to write the same code in every concrete model client.

bcallender: it's possible to do it here. you basically would have to pass the `EmbeddingModelParameters`/`CompletionModelParameters` from each concrete client implementation to the `__init__` here. if it's an  `EmbeddingModelParameters`, we want to validate that the input isn't larger than `max_input_size`. If it's a `CompletionModelParameters` we want to do the same validation you're already doing below.

YoungVor: that makes sense.  Though i think for CompletionModelParameters we can still have a separate check for input size too.

Basically the thought is to have separate checking/warning logic for completion vs embeddings

src/fenic/_inference/model_client.py
YoungVor: done

YoungVor:
@bcallender  I think @rohitrastogi has mentioned we could catch these issues when the message is first getting created, by refactoring so that the model parameters are available during the message creation in the semantic_operator classes, but this is an alternative solution that lets us check whether our estimates will be too large for an embeddings or language model.  Lmk if you think this is a worthwhile solution, or if we should hold off until we can try the larger refactor

bcallender:
Closing this as part of a PR backlog prune. The branch has gone stale against current `main`, and we do not expect to merge this context-window warning implementation in its current form. If we still want this exact warning behavior, we should revive it as a fresh, focused PR against current `main` and reconcile it with the token-limit safeguards now present in reduce/PDF/model-client paths. Thanks for the work here.

PR #213, opened 2025-09-16
feat: add PhysicalPlanOptimizer library and implement MergeDuckDBNodesRule

rohitrastogi:
1. Add `PhysicalPlanOptimizer` library to mirror `LogicalPlanOptimizer`.
2. Implement `MergeDuckDBNodesRule` to optimize DuckDB operations by merging adjacent DuckDB-related nodes into a single `MergedDuckDBExec`.

   * Traverses the plan bottom-up and flattens fusable DuckDB children for simpler execution.
   * Preserves the original subtree in `merge_root` so runtime execution can traverse it and register SQL views, allowing DuckDB to fuse operators internally.
   * `children` contains flattened nodes exposed to the outer engine, simplifying plan edges while maintaining correctness.
   * Only merges non-cached DuckDB nodes, leaving cached results and other nodes unaffected.


tests/_backends/local/physical_plan/test_physical_optimizer.py
```diff
@@ -0,0 +1,292 @@
...
+
+        # Apply optimization
+        optimizer = PhysicalPlanOptimizer(session_state, [MergeDuckDBNodesRule()])
+        result = optimizer.optimize(file_source)
+
+        # Should not be optimized (single node doesn't benefit from merging)
+        assert result.optimized is False
+        assert result.plan is file_source
+
+    def test_cache_boundary_prevents_merging(self, local_session: Session):
```
YoungVor, line 163: I assume the cache_boundary tests will just need a little tweak after you add support for cache (and support for duckdb client to connect to both the cache and catalog databases)

src/fenic/_backends/local/physical_plan/base.py
```diff
@@ -153,15 +155,21 @@ def _execute(self, child_dfs: List[pl.DataFrame]) -> pl.DataFrame:
...
     def build_lineage(self) -> LineageGraph:
         """Return the lineage of the physical plan."""
         leaf_nodes = []
-        root_node, _ = self._build_lineage(leaf_nodes)
+        root_node, _ = self.build_node_lineage(leaf_nodes)
         return LineageGraph(root_node, leaf_nodes)
 
-    def _build_lineage(
+    @abstractmethod
+    def build_node_lineage(
```
YoungVor, line 172: Not in scope for this PR, but thinking aloud - it would be good to have something help us guarantee that all physical/logical nodes have unit tests for their lineage and execute functionality.  @bcallender added that cool registry for all nodes to catch if anything gets implemented without the serde part, and he also uses that registry to ensure there is a test for every serde node.  Maybe we could add something similar for physical and logical nodes

src/fenic/_backends/local/physical_plan/optimizer/merge_duckdb_nodes.py
```diff
@@ -0,0 +1,114 @@
...
+        self, node: PhysicalPlan, session_state: LocalSessionState, original: PhysicalPlan
+    ) -> PhysicalPlan:
+        """Recursively optimize the plan tree bottom-up."""
+        # Optimize children first
+        optimized_children = [
+            self._optimize(child, session_state, original=child) for child in node.children
+        ]
+
+        # If children changed, make a new node with updated children
+        if optimized_children != node.children:
```
YoungVor, line 56: do we not need an __eq__ and _eq_specific_ in physical plan nodes?

src/fenic/_backends/local/physical_plan/optimizer/merge_duckdb_nodes.py
```diff
@@ -0,0 +1,114 @@
...
+        ]
+        if not fusable_children:
+            return node
+
+        # Flatten children
+        new_children = []
+        fusable_indices = {i for i, _ in fusable_children}
+        for i, child in enumerate(node.children):
+            if i in fusable_indices:
+                new_children.extend(child.children)
```
YoungVor, line 84: because you traverse bottom up, you know that the child's children are non-fusable 'non-duckdb' children.  You could consider having a check with InternalError for this assumption (but i don't feel strongly)

YoungVor (approved):
Minor comments, but looks awesome!

tests/_backends/local/physical_plan/test_physical_optimizer.py
rohitrastogi: yep

src/fenic/_backends/local/physical_plan/base.py
rohitrastogi: yeah, maybe a good idea

src/fenic/_backends/local/physical_plan/optimizer/merge_duckdb_nodes.py
rohitrastogi: we have tests for this.

src/fenic/_backends/local/physical_plan/optimizer/merge_duckdb_nodes.py
rohitrastogi: I do think they would be good to add generally speaking.  They aren't required for this code (instance equality is sufficient for correctness here) because we return child nodes untouched (no new instance) if there are no fusion opportunities.

YoungVor: makes sense!

PR #219, opened 2025-09-17
feat: Add page chunking to semantic parse_pdf, refactor token estimation

YoungVor:
### TL;DR

Gemini models especially are limited by the output token limit, only 8K tokens (for flash-2.0). Added PDF parsing by implementing chunking-by-page for large documents, based on the models output token limit.

Also refactored file support in LMRequestMessage (added LMRequestFile), and ability for gemini to send chunked pdfs in requests

### Chunks
- Added support for processing PDFs in chunks based on page count and output token limits
   - Chunking happens once, when requests are built
   - These chunks are saved as bytes in the LMRequestMessage, in new LMRequestFile class
   -`ParsePDF` operator will separate long PDFs into chunks with separate requests per chunk, then in postprocess, will concat the responses for each chunk into a single coherent output, honoring the page delimiter option
   - Added internal configuration to force PDF max chunk size (page count): PDF_MAX_PAGES_CHUNK
   - Added unit tests to test_parse_pdf to test various chunk sizes 

### Token estimations and LMRequestFile
- Token estimation optionally uses the pdf chunk bytes in the LMRequestMessages
- Added utility functions for PDF processing: pdf_to_base64, get_pdf_text, and get_pdf_pages
- Added Fixed token estimation for requests with PDF files for Gemini (258 tokens per page)

### Unit tests
1. Added test matrix to semantic_parse_pdf unit tests to run against various gemini  models
2. Added Unit tests for parse_pdf operator - testing various page chunk sizes, page deliminator options, and token limits/output with a mock llm, and validating the output
3. Added unit tests for token estimations
4. Added unit tests for semantic_parse_pdf that internally set the page batch size to 1 before sending requests to various llms and checking the semantic output 

### Stress tests
1. Run pdfparsing against hundreds of pages of financial documents using supported gemini models 
  2. With various chunking settings
  3. make sure they return markdown without any errors

## Not in scope of this PR

- OpenAI support
- Test harness for stress/advanced eval testing
- evaluating model accuracy

PR #207, opened 2025-09-15
fix: support for openai o1-mini model

YoungVor:
### What changed?

o1-mini is an older model that doesn't support a lot of the fields most other openai models support.  This PR is to add back support for o1-mini and work around those fields

- doesn't support reasoning
- doesn’t support system messages
- doesn't support json_schema (structured responses)

PR #200, opened 2025-09-10
feat: openai pdf_parse support and proper token estimation for pdf_parse

YoungVor:
### TL;DR

Add support for PDF parsing across multiple LLM providers by implementing base64 encoding for PDF files and proper token counting.

### What changed?

- Added PDF file handling capabilities to OpenAI and other LLM providers
- convert_messages is only used by the request, and encodes the file data
- All token estimations and dedup hash encodings use common LMRequestMessages class
- Enhanced token counter to handle PDFs and images with appropriate model-specific strategies
- Added utility functions for PDF processing: `pdf_to_base64`, `get_pdf_text_and_image_sizes`, and `pdf_to_base64_images`
- Fixed token estimation for requests with PDF files
- Add unit/eval tests for supported openai models

### How to test?

1. Run the PDF parsing tests 

### Why make this change?

This change enables PDF parsing across multiple LLM providers, not just Google's Gemini models. By implementing base64 encoding for PDFs and proper token counting, we can now use OpenAI models (like GPT-4o, GPT-4.1, and o-series models) and other providers to parse PDF documents.

PR #220, opened 2025-09-17
fix: doc loader should cast content in physical node and small api fixes

YoungVor:
## Changes

- doc_loader content cast to markdown happens in physical node with polars_plugins
- replace 'doc_path' with file_path in pdf_metadata() to be consistent
- take 'markdown' or 'json' in the docs signature
- Move path args validation for docs() to the top level api
- Allow Path types in docs() API for consistency
- Added equality tests


YoungVor:
* **#200** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/200?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#207](https://github.com/typedef-ai/fenic/pull/207) <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/207?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#219](https://github.com/typedef-ai/fenic/pull/219) <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/219?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#191** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/191?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#220** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/220?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/220?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/fenic/api/io/reader.py
```diff
@@ -266,27 +264,27 @@ def _read_file(
     def docs(
             self,
             paths: Union[str, list[str]],
-            data_type: Union[MarkdownType, JsonType],
+            data_type: Literal["markdown", "json"],
```
rohitrastogi: maybe content_type? it's not really a data type

src/fenic/api/io/reader.py
```diff
@@ -266,27 +264,27 @@ def _read_file(
...
             exclude: Optional[str] = None,
             recursive: bool = False,
     ) -> DataFrame:
-        r"""Load a DataFrame from a list of paths of documents (markdown or json).
+        r"""Load a DataFrame with the document contents of a list of paths (markdown or json).
 
         Args:
             paths: Glob pattern (or list of glob patterns) to the folder(s) to load.
             data_type: Data type that will be used to cast the content of the files.
```
rohitrastogi: casting is an implementation detail, not sure we need to mention it

src/fenic/api/io/reader.py
```diff
@@ -310,37 +308,30 @@ def docs(
...
             paths = [paths]
 
-        valid_file_extension = "md" if data_type == MarkdownType else "json"
+        if data_type not in ["markdown", "json"]:
+            raise UnsupportedFileTypeError(f"{data_type}, must be 'markdown' or 'json'")
+
         logical_node = DocSource.from_session_state(
             paths=paths,
-            valid_file_extension=valid_file_extension,
+            valid_file_extension="md" if data_type == "markdown" else "json",
```
rohitrastogi: why not just pass content_type all the way through and then convert to the valid file extension/data type only when needed?

src/fenic/api/io/reader.py
YoungVor: agreed, simplied to say its the content type of the files

src/fenic/api/io/reader.py
YoungVor: agreed, updated

src/fenic/_backends/local/utils/doc_loader.py
```diff
@@ -38,15 +40,15 @@ class DocFolderLoader:
     @validate_call(config=ConfigDict(strict=True))
```
rohitrastogi: This isn't a public API, so doesn't need validate call.

src/fenic/_backends/local/physical_plan/source.py
```diff
@@ -128,14 +132,14 @@ class DocSourceExec(PhysicalPlan):
     def __init__(
             self,
             paths: list[str],
-            valid_file_extension: str,
+            content_type: str,
```
rohitrastogi: We should use an enum - more type safe than a string. We should accept the literal as input but then use an enum all the way through imo

src/fenic/api/io/reader.py
YoungVor: updated

src/fenic/_backends/local/utils/doc_loader.py
YoungVor: makes sense!  removed.

src/fenic/_backends/local/physical_plan/source.py
YoungVor: its cleaner that way, and more efficient on the wire.  Done!

src/fenic/_backends/local/utils/doc_loader.py
YoungVor: Ah, it breaks a test where we catch 'None' as one of the inputs.  This is because we aren't validating the path glob inputs at the API level.  

I'll add the validation

protos/logical_plan/v1/enums.proto
```diff
@@ -54,3 +54,9 @@ enum FuzzySimilarityMethod {
   JARO = 4;
   HAMMING = 5;
 }
+
+enum DocContentType {
```
german-typedef, line 58: Love it!

protos/logical_plan/v1/enums.proto
```diff
@@ -54,3 +54,9 @@ enum FuzzySimilarityMethod {
   JARO = 4;
   HAMMING = 5;
 }
+
```
rohitrastogi, line 57: Consider moving this to core/types/enums.py instead of the utils file.

Also, we can use a literal, as long as we use the literal consistently!

YoungVor: ah good point!  CHanged to literal for consistency, and moved to enums.py

PR #221, opened 2025-09-17
refactor: separate concerns between building sql queries for duckdb io from executing queries

rohitrastogi:
**Summary:**
This PR separates the responsibilities of generating DuckDB SQL strings from executing those SQL queries. Previously, SQL generation and execution were tightly coupled, making it harder to implement certain optimizations and abstractions.

**Details:**

* Introduces a clear boundary between:

  1. Generating SQL strings for DuckDB I/O based on file parameters.
  2. Executing the SQL queries themselves.
* This separation is necessary for the `MergedDuckDBNode` physical plan node, whose `execute()` function now has a clean interface to:

  * Apply the DuckDB node execution.
  * Create a view for the resulting data.
* Improves test coverage for SQL generation to ensure correctness and maintainability.

**Impact:**

* Enables easier integration of DuckDB node execution with view creation.
* Makes the system more modular and testable, particularly for optimizations involving cached or merged DuckDB nodes.


src/fenic/_backends/local/utils/io_utils.py
```diff
@@ -113,93 +98,270 @@ def write_file(
...
+    else:
+        return f"Failed to read from Hugging Face. {e}"
+
+
 def _fetch_and_validate_s3_credentials(s3_session: BotoSession) -> Tuple[ReadOnlyCredentials, str]:
     """Check if s3 credentials are present and valid."""
     credentials = s3_session.get_credentials()
     if not credentials:
-        raise ConfigurationError("Unable to locate AWS credentials.")
+        raise ConfigurationError("AWS credentials were not found. Configure AWS credentials (env/aws_config) to read or write to S3.")
```
bcallender, line 350: nit: we're not really using these error messages since we're always catching them and logging a warning, do we even need to raise here or could we just return `None` and have an if-check for that?

src/fenic/_backends/local/utils/io_utils.py
```diff
@@ -1,19 +1,46 @@
...
+    FloatType,
+    IntegerType,
+    StringType,
+)
+from fenic.core.types.schema import Schema
 
 logger = logging.getLogger(__name__)
 
+@dataclass
+class ReadQueryResult:
```
bcallender, line 33: nit: perhaps `ResolvedReadQuery` or `ValidatedReadQuery`? `Result` suffix makes me think this is what is getting returned from a read query.

bcallender (approved):
just a couple of nits, otherwise this looks great

src/fenic/_backends/local/utils/io_utils.py
german-typedef: +1

german-typedef (approved):

src/fenic/_backends/local/utils/io_utils.py
rohitrastogi: We don't always catch them. We only catch them when we want to swallow the error and warn.

src/fenic/_backends/local/utils/io_utils.py
rohitrastogi: done.

PR #223, opened 2025-09-18
docs: move pre-requisite note to the start of the python tests command

yashmjain:
The pre-requisite note "**_Note: All tests require a valid OpenAI/Anthropic API key set in the environment variables_**." was originally placed after the Python test command. However, it should be presented beforehand so that users can configure the required environment variables prior to executing the tests.


CONTRIBUTING.md
```diff
@@ -105,6 +105,8 @@ uv run --group docs mkdocs serve
 
 ## ✅ Running Tests
 
+> ⚠️ Note: All tests require a valid OpenAI/Anthropic API key set in the environment variables.
```
YoungVor: Actually, this info is out of date - we support a growing list of providers -google, cohere, openrouter (in the next release).  Can we change it to say "tests require a valid API key for the language and embedding model providers they're configured to run with", or something to that effect

YoungVor (commented):
Thank you Yash for contributing to Fenic!  Looks good, after one minor change

CONTRIBUTING.md
YoungVor: We can also update this after the merge

YoungVor (approved):

CONTRIBUTING.md
yashmjain: Hi @YoungVor : I have modified the comments as per the review comments

YoungVor (changes requested):
Hey @yashmjain , thank you for the contribution!  Before making a commit, could you set up authentication signing?  You can follow the instructions here: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

bcallender:
Closing this as part of a PR backlog prune. The branch has gone stale against current `main`, and we do not expect to merge it in this form. If the underlying idea is still useful, we should revive it as a fresh, focused issue or PR against current `main`. Thanks for the work here.

PR #225, opened 2025-09-18
refactor: use tmpdir in all unit tests to avoid lingering duckdb/wal files

rohitrastogi:
### PR: Use temporary directories for DuckDB sessions and ensure WAL cleanup

**Problem:**
Some unit tests were seeing “lingering” DuckDB state between runs. This happened because:

* Local DuckDB sessions created `.duckdb` files in a persistent directory.
* Only the `.duckdb` file was deleted between tests; the `.duckdb.wal` file could persist if a test failed or the connection wasn’t closed properly.
* When a new session opened, DuckDB recovered the WAL file, restoring leftover state and making tests appear to leak data.

**Solution:**
* Use **temporary directories** in session configs so each test gets a fresh, isolated DuckDB directory that is wiped out automatically.
* Updated cleanup logic to delete both the `.duckdb` and `.duckdb.wal` files to ensure no leftover WAL data affects subsequent tests.

**Impact:**

* Tests now run in fully isolated environments.
* Eliminates false state leakage caused by orphaned WAL files.
* Ensures deterministic, repeatable unit test behavior for local DuckDB sessions.


bcallender (approved):
nice! now i can finally clean up all of the `.duckdb` files that are hanging out in my `tests` directory!

PR #228, opened 2025-09-21
fix: raise exception in user thread instead of shutting down when request above model tpm

YoungVor:
## What changed?

https://github.com/typedef-ai/fenic/pull/210/files fixes a bug where we hang forever if a single request token estimate is above the model's tpm limit, but it does so by incorrectly shutting down the processing queue.  This fix patches it so we raise the exception in the user thread instead.

## How to test?

Run unit tests with semantic operations, set model configuration with low tpm (like 10), make sure the semantic operations raise immediately

### DIFF
--- a/src/fenic/_inference/model_client.py
+++ b/src/fenic/_inference/model_client.py
@@ -544,29 +544,30 @@ class ModelClient(Generic[RequestT, ResponseT], ABC):
                     self.pending_requests = queue_items
 
                 # Iterate through pending requests and process those with available capacity
+                processed_requests = []
                 for queue_item in self.pending_requests:
-                    if self._check_and_consume_rate_limit(queue_item.estimated_tokens):
-                        task = asyncio.create_task(
-                            self._process_single_request(queue_item)
-                        )
-                        self._track_inflight_task(task)
-                        self.pending_requests.remove(queue_item)
-                    else:
-                        # Sleep for a short duration to wait for rate limit to refill to avoid busy-waiting
-                        await asyncio.sleep(MILLISECOND_IN_SECONDS)
+                    try:
+                        if self._check_and_consume_rate_limit(queue_item.estimated_tokens):
+                            task = asyncio.create_task(
+                                self._process_single_request(queue_item)
+                            )
+                            self._track_inflight_task(task)
+                            processed_requests.append(queue_item)
+                        else:
+                            # Sleep for a short duration to wait for rate limit to refill to avoid busy-waiting
+                            await asyncio.sleep(MILLISECOND_IN_SECONDS)
 
-                    await self._maybe_backoff()
+                        await self._maybe_backoff()
+                    except Exception as e:
+                        logger.error(f"Fatal error in request worker for model {self.model}: {e}", exc_info=True)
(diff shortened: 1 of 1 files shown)


YoungVor:
* **#228** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/228?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/228?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

bcallender (approved):

src/fenic/_inference/model_client.py
```diff
@@ -545,28 +545,27 @@ async def _process_queue(self):
...
+                                self._process_single_request(queue_item)
+                            )
+                            self._track_inflight_task(task)
+                            self.pending_requests.remove(queue_item)
+                        else:
+                            # Sleep for a short duration to wait for rate limit to refill to avoid busy-waiting
+                            await asyncio.sleep(MILLISECOND_IN_SECONDS)
+
+                        await self._maybe_backoff()
+                    except asyncio.CancelledError:
```
rohitrastogi: We don't need asyncio.CancelledError in here. The outer try catch will handle this.  asyncio.CancelledError will not be caught by Exception because it implements BaseException directly.

src/fenic/_inference/model_client.py
```diff
@@ -545,28 +545,27 @@ async def _process_queue(self):
...
+                        raise
+                    except Exception as e:
+                        logger.error(f"Fatal error in request worker for model {self.model}: {e}", exc_info=True)
+                        self._register_thread_exception(queue_item, e)
                         self.pending_requests.remove(queue_item)
```
rohitrastogi: We probably don't want to be mutating the list we are iterating over (I know we are already doing this). We can collect all the elements we process and remove from self.pending_requests at the very end.

src/fenic/_inference/model_client.py
```diff
@@ -545,28 +545,27 @@ async def _process_queue(self):
...
+                            self.pending_requests.remove(queue_item)
+                        else:
+                            # Sleep for a short duration to wait for rate limit to refill to avoid busy-waiting
+                            await asyncio.sleep(MILLISECOND_IN_SECONDS)
+
+                        await self._maybe_backoff()
+                    except asyncio.CancelledError:
+                        raise
+                    except Exception as e:
+                        logger.error(f"Fatal error in request worker for model {self.model}: {e}", exc_info=True)
```
rohitrastogi, line 562: I probably wouldn't use the word fatal here because we aren't killing the actor.

YoungVor: Good point, will update the wording

src/fenic/_inference/model_client.py
YoungVor: yep, makes sense.  Will change so we remove items after completing the iteration

src/fenic/_inference/model_client.py
YoungVor: got it!  Removed.

PR #232, opened 2025-09-22
feat: add tz support for timestamp type

german-typedef:
### TL;DR

Added timezone support to TimestampType in Fenic.

### Timestamp behavior

Similar to PySpark behavior.
- Timestamps in fenic are timezone aware.  All timestamps are represented as Timestamps in Microseconds against UTC
--  Naive timestamps (w/o timezone) can be loaded directly to fenic, but they are treated as UTC timestamps
- Users can create schemas with arbitrary timestamps and timezones 
- DuckDB sessions a convert underlying data to UTC 
-- DuckDB session will use local timezone when outputting data, which will affect some SQL commands
-- Naive timestamps will be treated as UTC
-- written as TIMESTAMPTZ
- Casting works between timestamps, strings, and ints.  Timestamps with Timezone is always returned.

### Special case behaviors: Naive timestamps
- 'Naive Timestamp', or Timestamp(timezone=None), does not exist in fenic.  We always default to UTC
- User CAN load naive timestamp values directly into fenic through polars
-- However, it will be treated as UTC by fenic functions, rust cast, and duckdb.  Any transformation will first add the UTC timezone.
- User CAN load strings that follow ISO standard. 
-- for e.g.: `df({ts_str:["2025-02-02 11:00:00-07:00"]}).cast(TimestampType)` will return a column of timestamps converted to UTC
- User CANNOT load NAIVE timestamps following ISO standard by casting.
-- `df({ts_str:["2025-02-02 11:00:00"]}).cast(TimestampType)` will fail silently and return [None]

### Special case behaviors: SQL and file source
- Duckdb does not save the timezone metadata, rather it converts all timestamps to the session/connection's timezone offset.
- DuckDB also uses microsecond intervals
- When using the catalog, we save the fenic schema and can convert them back to the original timezone.
- When using SQL functions or writing to and loading from files, these timezones will convert to UTC.

### Timezone Expressions:
- to_utc_timestamp (similar to spark function)
- from_utc_timestamp (similar to spark function)

### What changed?
- Modified `TimestampType` in Rust to include an optional timezone parameter
- Updated the Python `_TimestampType` class to store timezone information
- Enhanced conversion functions between Polars and Fenic data types to preserve timezone information
- Updated equality comparison for timestamp types to consider timezone
- Implemented proper string representation for timestamp types with timezone
- Unit tests for handling native, UTC, and arbitrary timezones
- Unit tests updated to expect TimestampType(timezone=)
- All entry points into DuckDB need to run apply_ingestion_coercions so we convert back to timestamp with UTC.  To do that, we needed to allow a path through saving tables that skips coercing Lists to Arrays (affects embedding types only, we want to keep the otpimization to treat embeddings as arrays)
-- added apply_ingestion_coercions to catalog, cache read, and lineage code paths (skipping coerce lists to arrays)


german-typedef:
* **#241** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/241?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#232** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/232?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/232?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#201** <a href="https://app.graphite.dev/github/pr/typedef-ai/fenic/201?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/fenic/_backends/schema_serde.py
```diff
@@ -141,6 +142,8 @@ def serialize_data_type(data_type: DataType) -> Dict[str, Any]:
         serialized["format"] = data_type.format
     elif isinstance(data_type, DocumentPathType):
         serialized["format"] = data_type.format
+    elif isinstance(data_type, _TimestampType):
+        serialized["timezone"] = data_type.timezone
```
YoungVor: we should have a TimestampType(timezone!=None) to test_datatype_serde

src/fenic/api/functions/dt.py
```diff
@@ -500,3 +503,81 @@ def timestamp_diff(start: ColumnOrName, end: ColumnOrName, unit: DateTimeUnit) -
...
+    Notes:
+        This function takes a timestamp which is timezone-agnostic,
+        and interprets it as a timestamp in UTC,
+        and renders that timestamp as a timestamp in the given timezone.
+
+    Example:
+        ```python
+        # ts: "2025-01-01 10:00:00", "2025-02-01 11:00:00", "2025-03-01 12:00:00"]
+        df.select(dt.from_utc_timestamp(col("ts"), "America/Los_Angeles")).to_pydict()
+        # Output: [{'ts': '2025-01-01 02:00:00'}, {'ts': '2025-02-01 03:00:00'}, {'ts': '2025-03-01 04:00:00'}]
```
YoungVor: do we not print the timezone when showing a TimestampType(timezone)?  How does the user keep track of what the type is, by printing the schema?

src/fenic/api/functions/dt.py
```diff
@@ -500,3 +503,81 @@ def timestamp_diff(start: ColumnOrName, end: ColumnOrName, unit: DateTimeUnit) -
...
+        ```
+    """
+    return Column._from_logical_expr(
+        FromUTCTimestampExpr(
+            Column._from_col_or_name(column)._logical_expr,
+            tz))
+
+@validate_call(config=ConfigDict(strict=True, arbitrary_types_allowed=True))
+def convert_timezone(sourceTz: Optional[str], targetTz: str, sourceTs: ColumnOrName) -> Column:
+    """Converts the timestamp without time zone sourceTs from the sourceTz time zone to targetTz.
```
YoungVor: I think we should have a little more detail in the docstring for this one.

Is it a requirement that the sourceTs type does not have tz set?  What happens if it does, is it ignored, or preferred over the sourceTz arg, or raise exception?

What happens if the sourceTz  is none, do we treat source timestamp as UTC?

src/fenic/api/functions/dt.py
```diff
@@ -500,3 +503,81 @@ def timestamp_diff(start: ColumnOrName, end: ColumnOrName, unit: DateTimeUnit) -
...
+
+    Args:
+        sourceTz: The timezone of the source timestamp.
+        targetTz: The timezone of the target timestamp.
+        sourceTs: The column containing the timestamp.
+
+    Returns:
+        A Column object with the timestamp converted to the given timezone.
```
YoungVor: Related to a previous message, with this support, I have a question on how is the user meant to manage TimestampType(timezone="something") vs TimestampType(None), and how we display it for them.

src/fenic/core/types/datatypes.py
```diff
@@ -216,13 +216,15 @@ def __eq__(self, other: object) -> bool:
...
+    timezone: Optional[str] = None
+
     def __str__(self) -> str:
         """Return a string representation of the timestamp type.
 
         Returns:
             The string "TimestampType".
         """
-        return "TimestampType"
+        return f"TimestampType(timezone={self.timezone})"
```
YoungVor: This will be TimestampType(timezone=None) if timezone isn't set/for UTC.  Should we print TimestampType, or TimestampType(UTC) in this case?

Also, nit: polars just prints the timezone, no 'timezone=':
`datetime[ns, America/New_York]`
`datetime[ns]` (for no timezone set)

src/fenic/core/_logical_plan/expressions/dt.py
```diff
@@ -374,17 +376,94 @@ def _eq_specific(self, other: TimestampDiffExpr) -> bool:
...
-    """Resolve the expression's type if it's DATE or TIMESTAMP, otherwise raise an error."""
+    """Resolve the expression's type if it's DATE or TIMESTAMP, otherwise raise an error.
+
+    Currently used for FromUTCTimestampExpr and ToUTCTimestampExpr, as in these methods it assumes the column
+    shouldn't have a timezone.
+    """
     col_field = expr.to_column_field(plan, session_state)
     if col_field.data_type in [DateType, TimestampType]:
         return col_field.data_type
     raise TypeError(f"Expected one of DateType or TimestampType, got {col_field.data_type}")
```
YoungVor: We should raise an explicit message if we expect Timestamp() with no timezone but get Timestamp(timezone)

src/fenic/core/_logical_plan/expressions/dt.py
```diff
@@ -374,17 +376,94 @@ def _eq_specific(self, other: TimestampDiffExpr) -> bool:
...
 
 
+def _resolve_temporal_type_with_timezone(expr: LogicalExpr, plan: LogicalPlan, session_state: BaseSessionState) -> bool:
+    """Resolve the expression's type if it's DATE or TIMESTAMP, otherwise raise an error."""
+    col_field = expr.to_column_field(plan, session_state)
+    if col_field.data_type in [DateType, TimestampType]:
+        return col_field.data_type
+    elif isinstance(col_field.data_type, _TimestampType):
+        return col_field.data_type
+    raise TypeError(f"Expected one of DateType or TimestampType, got {col_field.data_type}")
```
YoungVor, line 426: Similar to above - We should raise an explicit message if we expect Timestamp(timezone) but get Timestamp

src/fenic/core/_logical_plan/expressions/dt.py
```diff
@@ -374,17 +376,94 @@ def _eq_specific(self, other: TimestampDiffExpr) -> bool:
...
+    def _eq_specific(self, other: ConvertTimezoneExpr) -> bool:
+        return (self.source_timezone == other.source_timezone and
+            self.target_timezone == other.target_timezone and
+            self.expr == other.expr)
+
+
 # UTILITY FUNCTIONS
 
 
 def _resolve_temporal_type(expr: LogicalExpr, plan: LogicalPlan, session_state: BaseSessionState) -> bool:
```
YoungVor: nit: this is really temporal_type_without_timezone

YoungVor (commented):
Went through the code and api, haven't finished going through the tests yet.  My big question is how we are presenting the timezone to the user, and that we should be fairly explicit in documenting the experience for the user.

* is it ok that we don't print the timezone as part of the timestamp value itself?
* how do we represent Timstamp(UTC) vs Timestamp(default) vs Timestamp(timezone)
* we raise with explicit messages if we expect Timestamp() vs Timestamp(timezone) and vice versa

src/fenic/api/functions/dt.py
```diff
@@ -500,3 +503,81 @@ def timestamp_diff(start: ColumnOrName, end: ColumnOrName, unit: DateTimeUnit) -
...
+        # Output: [{'ts': '2025-01-01 02:00:00'}, {'ts': '2025-02-01 03:00:00'}, {'ts': '2025-03-01 04:00:00'}]
+        ```
+    """
+    return Column._from_logical_expr(
+        FromUTCTimestampExpr(
+            Column._from_col_or_name(column)._logical_expr,
+            tz))
+
+@validate_call(config=ConfigDict(strict=True, arbitrary_types_allowed=True))
+def convert_timezone(sourceTz: Optional[str], targetTz: str, sourceTs: ColumnOrName) -> Column:
```
YoungVor: should there be a sourceTz if we expect a TimestampType(timezone) as the input?

src/fenic/api/functions/dt.py
YoungVor: The timezone does get printed in the value of a Timestamp with timezone set

src/fenic/core/types/datatypes.py
YoungVor: Since we default to making timestamps tz aware and UTC, I'm less concerned about this.  tz=None is still possible if the user loads polars/python vales with naive values directly into fenic though.

src/fenic/api/functions/dt.py
YoungVor: changed the API after updating the design to have Timestamp always default to timezone = UTC

src/fenic/_backends/schema_serde.py
YoungVor: made timezone a required argument, so this isn't necessary

src/fenic/_backends/local/transpiler/expr_converter.py
```diff
@@ -1302,12 +1309,16 @@ def _convert_to_date_expr(self, logical: ToDateExpr) -> pl.Expr:
     def _convert_to_timestamp_expr(self, logical: ToTimestampExpr) -> pl.Expr:
         """Convert a string into a timestamp."""
         # logical.format should be a chrono format.
-        return self._convert_expr(logical.expr).str.strptime(dtype=pl.Datetime, format=logical.format)
+        converted_expr = self._convert_expr(logical.expr).str.strptime(dtype=pl.Datetime, format=logical.format)
+        # If the format includes timezone, the input must specify its own timezone
+        if _format_has_tz_tokens(logical.format):
```
rohitrastogi, line 1313: We can make the behavior of forcing to UTC vectorized instead of the Python for loop.

src/fenic/_backends/local/physical_plan/utils.py
```diff
@@ -39,6 +39,9 @@ def _build_target_dtype(dtype: pl.DataType) -> pl.DataType:
         return pl.List(_build_target_dtype(dtype.inner))
     elif isinstance(dtype, pl.List):
         return pl.List(_build_target_dtype(dtype.inner))
+    elif isinstance(dtype, pl.Datetime):
+        # DuckDB always uses UTC as its session timezone, so we set UTC here.
+        return pl.Datetime(time_unit="us", time_zone=dtype.time_zone or "UTC")
```
rohitrastogi: If we always set this to UTC, we don't need to worry about setting the DuckDB timezone (since we'll always correctly cast this to "UTC" from all DuckDB codepaths)

src/fenic/_backends/local/transpiler/expr_converter.py
```diff
@@ -190,6 +194,9 @@ def _literal_to_polars_expr(value: Any, data_type: DataType) -> pl.Expr:
             if value is None:
                 return pl.lit(None, dtype=convert_custom_dtype_to_polars(data_type))
 
+            if isinstance(data_type, _TimestampType):
+                return pl.lit(pytype_datetime_to_utc(value), dtype=pl.Datetime(time_zone="UTC"))
```
rohitrastogi: nit: just consider inlining this function since it's not going to be used elsewhere.

src/fenic/_backends/local/utils/io_utils.py
```diff
@@ -366,5 +375,5 @@ def _fetch_hf_token() -> Optional[str]:
 
 def _configure_duckdb_conn(db_conn: duckdb.DuckDBPyConnection) -> duckdb.DuckDBPyConnection:
     """Common configuration for duckdb connections."""
-    db_conn.execute("set arrow_large_buffer_size=true;")
+    db_conn.execute("set arrow_large_buffer_size=true; SET TIMEZONE='UTC';")
```
rohitrastogi: here and elsewhere, not sure this is required if we just correctly cast on ingestion from duckdb.

src/fenic/core/_logical_plan/plans/source.py
```diff
@@ -22,7 +25,7 @@ def __init__(
             source: pl.DataFrame,
             session_state: Optional[BaseSessionState] = None,
             schema: Optional[Schema] = None):
-        self._source = source
```
rohitrastogi, line 25: nit: maybe worth an inline

tests/_backends/local/functions/test_date_functions.py
```diff
@@ -205,9 +247,22 @@ def test_to_timestamp(self, df_with_date_types):
         """Test the to_timestamp function."""
         df = df_with_date_types.select(to_timestamp(col("ts_str")).alias("ts2"))
         self._verify_results_on_df(df, TimestampType, "ts2", TS_FULL_RESULTS)
+        # We default all timestamps to UTC
+        assert df.to_pydict()['ts2'][0].tzinfo == zoneinfo.ZoneInfo(key='UTC')
 
         self._verify_error(df_with_date_types, "col1", to_timestamp, TypeError)
 
+    def test_to_timestamp_tz(self, df_with_date_types):
```
rohitrastogi: this test name doesn't describe the function we are actually testing

tests/_backends/local/functions/test_greatest_least.py
```diff
@@ -135,17 +136,34 @@ def test_least(greatest_least_source):
...
         df = df_source.select(greatest(col("datetime_col"), col("datetime_col2")).alias("greatest"))
         results = df.to_pydict()
-        assert results["greatest"] == [datetime.datetime(2023, 12, 25, 14, 30), datetime.datetime(2024, 1, 1, 9, 18)]
+        # We convert naive datetimes to UTC.
+        assert results["greatest"] == [datetime.datetime(2023, 12, 25, 14, 30, tzinfo=datetime.timezone.utc), datetime.datetime(2024, 1, 1, 9, 18, tzinfo=datetime.timezone.utc)]
+
+
+        df = df_source.select(least(col("datetime_col_tz_la"), col("datetime_col_tz_la2")).alias("least"))
+        results = df.to_pydict()
+        print(results)
```
rohitrastogi: nit: remove the print statement

src/fenic/_backends/local/transpiler/expr_converter.py
YoungVor: spoke in person, we replaced this with the same pattern - if format has timezone , use polars replace_timestamp.  If format does not, use polars convert_timestamp expr.

src/fenic/_backends/local/physical_plan/utils.py
YoungVor: Spoke in person, we discovered we still need the apply_ingestion_coercions when loading tables (which means we need to optionally skip coercing Lists), but otherwise we apply coercions in every entrypoint from duckdb, so this is correct.

src/fenic/core/_logical_plan/plans/source.py
YoungVor: sure, changed to InMemorySource class method

src/fenic/_backends/local/transpiler/expr_converter.py
YoungVor: done

tests/_backends/local/functions/test_date_functions.py
YoungVor: updated to date_format

tests/_backends/local/functions/test_greatest_least.py
YoungVor: done

src/fenic/_backends/local/utils/io_utils.py
YoungVor: removed

YoungVor: removed since we're casting on all ingest paths now

src/fenic/core/_logical_plan/plans/source.py
```diff
@@ -62,6 +64,31 @@ def with_children(self,
     def _eq_specific(self, other: InMemorySource) -> bool:
         return self._source.equals(other._source)
 
+    @classmethod
```
rohitrastogi: This code doesn't handle nested container types (structs, lists, etc) that might have timestamps embedded in them. Also, after thinking about this further, I think we actually don't need this code since we are calling apply_ingestion_coercions in the physical source.

We do need to update apply_ingestion_coercions to handle timestamps where datetime is none (this will only be exercised for polars sources).

src/fenic/_backends/local/physical_plan/source.py
```diff
@@ -69,7 +69,7 @@ def execute_node(self, child_dfs: List[pl.DataFrame]) -> pl.DataFrame:
 
         file_format = self.file_format.lower()
         df = query_files(paths=self.paths, file_type=file_format, s3_session=self.session_state.s3_session, **self.options)
-        return apply_ingestion_coercions(df)
+        return apply_ingestion_coercions(df, coerce_array=True)
```
rohitrastogi, line 72: We need to call this on cache reads too, otherwise timestamps won't be in UTC. We may need to handle this for lineage calculations too, but I'm having a hard time thinking through it.

Probably good to add some tests for timestamp handling from cache.

You might have been correct that doing this at the connection/cursor level may have been simpler, but if we take that approach, we should just override the .cursor() method on a wrapped connection object to also set the time so we don't have to call a special method every time its used.

src/fenic/_backends/local/catalog.py
```diff
@@ -717,6 +721,12 @@ def _does_view_exist(self, cursor: duckdb.DuckDBPyConnection, view_identifier: T
                 f"Failed to check if view: `{view_identifier.db}.{view_identifier.table}` exists"
             ) from e
 
+    def _resolve_table_schema(self, cursor: duckdb.DuckDBPyConnection, table_identifier: TableIdentifier, table:pl.DataFrame) -> pl.DataFrame:
```
rohitrastogi: Why did we add this code?

src/fenic/_backends/local/physical_plan/transform.py
```diff
@@ -396,8 +396,8 @@ def execute_node(self, child_dfs: List[pl.DataFrame]) -> pl.DataFrame:
         for child_df, view_name in zip(child_dfs, self.view_names, strict=False):
             cursor.register(view_name, child_df)
         try:
-            pl_result = cursor.execute(self.query).pl()
-            return apply_ingestion_coercions(pl_result)
+            pl_result = cursor.execute(f"{self.query}").pl()
```
rohitrastogi: nit: no need for the format string

src/fenic/_backends/local/transpiler/expr_converter.py
```diff
@@ -1431,5 +1464,19 @@ def _like_to_regex(pattern_expr: pl.Expr, case_insensitive: bool = False) -> pl.
 
     return regex_expr
 
+def _format_has_tz_tokens(fmt: str):
```
rohitrastogi: nit: missing return type hint

rust/src/dtypes/types.rs
```diff
@@ -64,9 +64,12 @@ impl FenicDType {
...
             FenicDType::BooleanType => DataType::Boolean,
             FenicDType::DateType => DataType::Date,
-            // TODO: Timezone support will be added later.
-            // For datetime, we'll use nanoseconds as the time unit.
-            FenicDType::TimestampType => DataType::Datetime(TimeUnit::Nanoseconds, None),
+            // For datetime, we'll use microseconds as the time unit.
+            // TODO(DY) don't need opt_try_new, we know this works
+            FenicDType::TimestampType => DataType::Datetime(
+                TimeUnit::Microseconds,
+                TimeZone::opt_try_new("UTC".into()).unwrap(),
```
rohitrastogi: nit: don't need opt_try_new. you can just do TimeZone::UTC, I think.

src/fenic/_backends/local/physical_plan/source.py
YoungVor: Good idea, will add a few cache tests for timestamp handling.  

I'll try to think through the lineage piece

src/fenic/core/_logical_plan/plans/source.py
YoungVor: Ah, good point.  Removed

src/fenic/_backends/local/catalog.py
YoungVor: looks like old code left over from a previous pr or merge.  Removed

rust/src/dtypes/types.rs
YoungVor: done, the idiomatic way is with from_static

src/fenic/_backends/local/physical_plan/source.py
YoungVor: I think all we need is the apply_coercions on the temp_df client's read_df, that seems to be the only entrypoint for lineage.  Added that and updated the lineage unit tests.

I don't *think* we need any tests beyond the simple forward and simple backwards test for testing the timestamp coercion in lineage, but let me know if you think its worth having a more complex test

rohitrastogi: yeah i think this makes sense

src/fenic/_backends/schema_serde.py
```diff
@@ -172,6 +173,8 @@ def _deserialize_data_type(data: Dict[str, Any]) -> DataType:
         return TranscriptType(format=data["format"])
     elif type_name == "DocumentPathType":
         return DocumentPathType(format=data["format"])
+    elif type_name == "TimestampType":
+        return TimestampType
```
bcallender: nit: you don't have to add this here if the type is not parameterized -- you can just add it to the `_TYPE_MAPPING` constant and it will be automatically mapped. having it here implies that there is some parameterization we need to add to it.

src/fenic/api/functions/dt.py
```diff
@@ -235,7 +235,7 @@ def to_timestamp(column: ColumnOrName, format: Optional[str] = None) -> Column:
         format: The format of the timestamp string.
 
     Returns:
-        A Column object with the TimestampType transformed.
+        A Column object with the TimestampType transformed.  If timezone is not in the format, it sets the timezone to UTC.  If the timezone is in the format, it converts the timezone to UTC.
```
bcallender: nit: `it` is unclear in this case, at least to me. Would prefer something like: 
```
A Column object with the `TimestampType` type, with a UTC timezone. If the provided `format` contains a timezone specifier, the result rows will be converted from the `format` timezone to UTC.
```

src/fenic/api/functions/dt.py
```diff
@@ -491,12 +489,110 @@ def timestamp_diff(start: ColumnOrName, end: ColumnOrName, unit: DateTimeUnit) -
...
+
+    Notes:
+        - In fenic, the TimestampType data type is always in UTC, which is not timezone-agnostic.
+        - Spark also represents all timestamps as not timezone-agnostic. The only difference with Spark is that the session timestamp is always UTC instead of local.
+        - Similarly to Spark to_utc_timestamp function, this function treats the input timestamp as timezone-agnostic, sets it to the requested timezone (without changing the timestamp), then converts the timestamp to UTC.
+        - see Spark documentation for more details: https://spark.apache.org/docs/4.0.1/api/python/reference/pyspark.sql/api/pyspark.sql.functions.to_utc_timestamp.html
+
+    Example:
+        ```python
+        df.select("timestamp", to_timestamp(col("timestamp"), "yyyy-MM-dd HH:mm:ss").alias("la_time")).show()
```
bcallender: nit: inconsistent use of `dt.<func>` vs no import prefix

src/fenic/api/functions/dt.py
```diff
@@ -491,12 +489,110 @@ def timestamp_diff(start: ColumnOrName, end: ColumnOrName, unit: DateTimeUnit) -
...
+
+    Returns:
+        A Column object with timestamp rendered in UTC.
+
+    Raises:
+        TypeError: If column type is not a TimestampType.
+
+    Notes:
+        - In fenic, the TimestampType data type is always in UTC, which is not timezone-agnostic.
+        - Spark also represents all timestamps as not timezone-agnostic. The only difference with Spark is that the session timestamp is always UTC instead of local.
```
bcallender: nit: I know this is saying that _fenic_ is different than spark because _fenic_ always treats the session timestamp as UTC instead of the local user's time, but this sentence is ambiguous imo, it could be interpreted the opposite way.

src/fenic/api/functions/dt.py
```diff
@@ -491,12 +489,110 @@ def timestamp_diff(start: ColumnOrName, end: ColumnOrName, unit: DateTimeUnit) -
...
+        ```
+    """
+    return Column._from_logical_expr(
+        ToUTCTimestampExpr(
+            Column._from_col_or_name(column)._logical_expr,
+            tz))
+
+@validate_call(config=ConfigDict(strict=True, arbitrary_types_allowed=True))
+def from_utc_timestamp(column: ColumnOrName, tz: str) -> Column:
+    """Takes a timestamp in UTC, converts it to the provided timezone, and renders that timestamp as a timestamp in UTC.
```
bcallender: nit: `convert` makes it seem like the final output will be a timestamp type with the provided timezone, rather than the same timestamp `adjusted` to the target timezone, rendered in UTC. 

I think by clarifying that it takes a Column of `TimestampType`, it makes it clear that the timestamp will already be UTC and it makes this a little less confusing:

```
Accepts a Column[TimestampType] (UTC). For each input instant `t`, computes the wall‑clock time in timezone `tz` at `t`, then re‑expresses that wall‑clock time as a UTC timestamp. In other words, it shifts by the timezone offset at that instant: `out = t − offset(tz, t)`. 
```

I would also note that: 

- nulls are preserved
- Use when to “apply” `tz` to a UTC instant and keep the resulting wall‑clock, still stored as UTC.
- Use to_utc_timestamp for the inverse (treats naive/local values in tz, then converts to UTC).

src/fenic/api/functions/dt.py
bcallender: Would also add some usage notes: 

```
- Use this when data contains local/wall‑clock timestamps from `tz` and the actual UTC instants are required.
- For the inverse operation (UTC → local wall‑clock, then re‑expressed in UTC), see from_utc_timestamp.
```

src/fenic/api/functions/dt.py
```diff
@@ -491,12 +489,110 @@ def timestamp_diff(start: ColumnOrName, end: ColumnOrName, unit: DateTimeUnit) -
...
     return Column._from_logical_expr(
         TimestampDiffExpr(
             Column._from_col_or_name(start)._logical_expr,
             Column._from_col_or_name(end)._logical_expr,
             unit))
+
+
+@validate_call(config=ConfigDict(strict=True, arbitrary_types_allowed=True))
+def to_utc_timestamp(column: ColumnOrName, tz: str) -> Column:
+    """Takes a timestamp, treats it as timezone-agnostic, interprets it as a timestamp in the provided timezone, and converts that timestamp to UTC.
```
bcallender: I think this is more clear than `from_utc_timestamp`, but might still be a little less ambiguous:

```
Accepts a Column[TimestampType] and interprets each value as wall-clock time in the specified timezone `tz`. Returns the equivalent instant in UTC.
```

src/fenic/_backends/schema_serde.py
YoungVor: thanks! Yes, it used to be parametereized in an old version.  Removed.

src/fenic/api/functions/dt.py
YoungVor: I like it.  Tweaked to say the timestamp offset will be modified

src/fenic/api/functions/dt.py
YoungVor: good catch.  Updated

src/fenic/api/functions/dt.py
YoungVor: done.  A few tweaks, and changed to t + offset (the example is -8 from UTC)

src/fenic/api/functions/dt.py
YoungVor: i like it.  Done

src/fenic/api/functions/dt.py
YoungVor: done.  makes sense not to mention the timezone agnostic part here since we go into detail in the notes

bcallender (approved):
I think the API is clearer now. It's still kind of confusing, but I think that is just the nature of working with timestamps and timezones, the existing pyspark docs have the same problems.

PR #245, opened 2025-10-04
refactor: reorganize array expressions in proto for LLM-friendly development

bcallender:
- Group all array expressions together with inline documentation
- Add comments guiding future array function additions
- Preserve existing field numbers for backwards compatibility
- No functional changes, pure organization

PR #243, opened 2025-10-01
feat: add support for array_distinct

bcallender:

PR #242, opened 2025-10-01
feat: add support for sum_discinct, count_distinct, approx_count_distinct

bcallender:
feat: add support for sum_discinct, count_distinct, approx_count_distinct

fix behavior to be more in line with pyspark semantics

PR #247, opened 2025-10-04
feat: implement PySpark regexp_* text functions

bcallender:
- Add regexp_count: count regex pattern matches
- Add regexp_extract: extract specific capture group from match
- Add regexp_extract_all: extract all matches of a capture group (uses extract_groups for PySpark compatibility)
- Add regexp_instr: find 1-based position of regex match
- Add regexp_substr: extract first substring matching pattern
- Validate regex patterns using py_validate_regex
- Add Polars backend implementations with PySpark compatibility
- Add type signatures for all functions
- Add protobuf definitions and serde support
- Add comprehensive tests including edge cases
- All functions accept string or column expressions for patterns


bcallender:
* **#249** <a href="https://app.graphite.com/github/pr/typedef-ai/fenic/249?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#247** <a href="https://app.graphite.com/github/pr/typedef-ai/fenic/247?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/typedef-ai/fenic/247?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#243** <a href="https://app.graphite.com/github/pr/typedef-ai/fenic/243?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#245** <a href="https://app.graphite.com/github/pr/typedef-ai/fenic/245?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#242** <a href="https://app.graphite.com/github/pr/typedef-ai/fenic/242?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/fenic/api/functions/text.py
```diff
@@ -640,6 +645,255 @@ def regexp_replace(
...
+    return Column._from_logical_expr(
+        RegexpCountExpr(Column._from_col_or_name(src)._logical_expr, pattern_expr)
+    )
+
+
+@validate_call(config=ConfigDict(strict=True, arbitrary_types_allowed=True))
+def regexp_extract(
+    src: ColumnOrName, pattern: Union[Column, str], idx: Union[Column, int]
+) -> Column:
+    r"""Extract a specific regex group from a string.
```
YoungVor, line 696: Pyspark doc isn't super clear on this either, but we are implementing it so its the group from the first match, right?  Can we mention that here?

src/fenic/api/functions/text.py
```diff
@@ -640,6 +645,255 @@ def regexp_replace(
...
+    Extracts a capture group matched by the regex pattern.
+    Group 0 is the entire match, group 1+ are capture groups.
+
+    Args:
+        src: The input string column or column name
+        pattern: The regex pattern with capture groups (can be a string literal or column expression)
+        idx: The group index to extract (0 = entire match, 1+ = capture groups)
+
+    Returns:
+        Column: A string column containing the extracted group (null if no match)
```
YoungVor: interesting - pyspark behavior is to return an empty string, not null.

YoungVor: nit: if regex did not match, or the specified group did not match

src/fenic/core/_logical_plan/signatures/text.py
```diff
@@ -184,6 +184,52 @@ def register_text_signatures():
         )
     )
 
+    # Regex functions
+    FunctionRegistry.register(
```
YoungVor, line 188: Can we support String-like logical types like Markdown and Json?

src/fenic/api/functions/text.py
```diff
@@ -640,6 +645,255 @@ def regexp_replace(
...
+        pattern_expr = pattern._logical_expr
+    else:
+        pattern_expr = lit(pattern)._logical_expr
+    return Column._from_logical_expr(
+        RegexpCountExpr(Column._from_col_or_name(src)._logical_expr, pattern_expr)
+    )
+
+
+@validate_call(config=ConfigDict(strict=True, arbitrary_types_allowed=True))
+def regexp_extract(
```
YoungVor, line 693: Weird, regexp_extract only accepts string patterns and ints for pattern/idx, not columns.  But not the case for regexp_extract_all, which makes not sense to me.  I don't think its worth changing here

src/fenic/api/functions/text.py
```diff
@@ -640,6 +645,255 @@ def regexp_replace(
...
+
+@validate_call(config=ConfigDict(strict=True, arbitrary_types_allowed=True))
+def regexp_substr(src: ColumnOrName, pattern: Union[Column, str]) -> Column:
+    r"""Extract the first substring matching a regex pattern.
+
+    Returns the first substring that matches the pattern. Returns null if no match is found.
+
+    Args:
+        src: The input string column or column name
+        pattern: The regex pattern to search for
```
YoungVor, line 860: nit: here and the other APIs- can mention pattern to search for is a string or column of strings, idx is an int or column of ints, etc.

YoungVor: can do it like rlike : "other (Union[str, Column]): <description>"

src/fenic/core/_logical_plan/signatures/text.py
YoungVor: actually it looks like the other text stuff like rlike can't handle markdown either, so probably out of scope for this pr.  Please disregard.

User can always cast to string

tests/_backends/local/functions/test_string_functions.py
```diff
@@ -839,3 +839,161 @@ def test_trim_whitespace(whitespace_trim_df):
...
+    assert result["username"][3] is None
+
+
+def test_regexp_extract_entire_match(local_session):
+    """Test regexp_extract with group 0 (entire match)."""
+    data = {"text": ["Price: $123.45", "Cost: $67.89", "No price"]}
+    df = local_session.create_dataframe(data)
+
+    result = df.select(
+        text.regexp_extract("text", r"\$\d+\.\d+", 0).alias("price")
```
YoungVor: nit, could still have the capture groups in the regex and test that we still return whole match

tests/_backends/local/functions/test_string_functions.py
```diff
@@ -839,3 +839,161 @@ def test_trim_whitespace(whitespace_trim_df):
...
+
+
+def test_regexp_extract(local_session):
+    """Test regexp_extract function."""
+    data = {"email": ["user@domain.com", "admin@example.org", "invalid", None]}
+    df = local_session.create_dataframe(data)
+
+    # Extract username (group 1)
+    result = df.select(
+        text.regexp_extract("email", r"([^@]+)@", 1).alias("username")
```
YoungVor: Can we include some tests for group not matching, though pattern matches?  I think there are two interesting cases:

whole pattern matches, but group id is out of capture group range
whole pattern matches, but the group does not match (with some 'or' expression)

tests/_backends/local/functions/test_string_functions.py
```diff
@@ -839,3 +839,161 @@ def test_trim_whitespace(whitespace_trim_df):
...
+    assert result["url"][1] is None
+    assert result["url"][2] == "http://test.org"
+
+
+def test_regexp_invalid_pattern(local_session):
+    """Test that invalid regex patterns are caught."""
+    data = {"text": ["test"]}
+    df = local_session.create_dataframe(data)
+
+    with pytest.raises(ValidationError, match="Invalid regex pattern"):
```
YoungVor: good to include a column containing an invalid regex too

src/fenic/_backends/local/transpiler/expr_converter.py
```diff
@@ -878,6 +883,115 @@ def _convert_regexp_split_expr(self, logical: RegexpSplitExpr) -> pl.Expr:
...
+            
+            if group_idx == 0:
+                # Group 0: extract entire match using extract_all
+                return physical_expr.str.extract_all(pattern)
+            else:
+                # Group 1+: extract all matches, then use extract_groups on each to get specific capture group
+                # extract_groups returns a struct with numbered fields ("1", "2", etc.)
+                return (
+                    physical_expr.str.extract_all(pattern)
+                    .list.eval(pl.element().str.extract_groups(pattern).struct[str(group_idx)])
```
YoungVor: its not ideal to run two separate regex expressions across the series...

would it be possible to modify the regex pattern itself so we only capture the group we want?  Either by removing other group parens or adding a 'Lookbehind/Lookahead' to the group of choice?

I suppose this also involves 2 regexes across series (if the regex is a column) so maybe it doesn't buy us much...

YoungVor: Ah, plus, that approach doesn't work if the regex matches but the groups don't match.

pyspark documentation seems to indicate idx is based on the set of returned captures, not based on the index of the capture in the regex

src/fenic/_backends/local/transpiler/expr_converter.py
```diff
@@ -878,6 +883,115 @@ def _convert_regexp_split_expr(self, logical: RegexpSplitExpr) -> pl.Expr:
...
+        # We'll extract the match and then find its position in the original string
+        if isinstance(logical.pattern, LiteralExpr) and isinstance(logical.idx, LiteralExpr):
+            # Extract the match
+            match_expr = physical_expr.str.extract(pattern_expr, int(logical.idx.literal))
+            # Find position: use str.find on the match within the original string
+            # Add 1 to convert from 0-based to 1-based indexing (PySpark compatibility)
+            # Return 0 if no match, null if input is null
+            return pl.when(physical_expr.is_null()).then(None).otherwise(
+                pl.when(match_expr.is_null()).then(pl.lit(0)).otherwise(
+                    physical_expr.str.find(match_expr, literal=True) + 1
```
YoungVor, line 938: don't you still need the str.extract step for catpuring the desired group in the literal case?

YoungVor: oh, sorry, thats the match_expr.  Please disregard

src/fenic/_backends/local/transpiler/expr_converter.py
```diff
@@ -878,6 +883,115 @@ def _convert_regexp_split_expr(self, logical: RegexpSplitExpr) -> pl.Expr:
...
+
+    @_convert_expr.register(RegexpInstrExpr)
+    def _convert_regexp_instr_expr(self, logical: RegexpInstrExpr) -> pl.Expr:
+        physical_expr = self._convert_expr(logical.expr)
+        pattern_expr = self._convert_expr(logical.pattern)
+        idx_expr = self._convert_expr(logical.idx)
+
+        # Polars doesn't have a built-in for this, so we need to implement it
+        # We'll extract the match and then find its position in the original string
+        if isinstance(logical.pattern, LiteralExpr) and isinstance(logical.idx, LiteralExpr):
```
YoungVor: Does this also work if only idx is literal but pattern is not?

tests/_backends/local/functions/test_string_functions.py
```diff
@@ -839,3 +839,161 @@ def test_trim_whitespace(whitespace_trim_df):
...
+    assert result["position"][3] is None
+
+
+def test_regexp_instr_with_email(local_session):
+    """Test regexp_instr finding email position."""
+    data = {"text": ["Contact: user@domain.com", "No email here", "Email: admin@test.com"]}
+    df = local_session.create_dataframe(data)
+
+    result = df.select(
+        text.regexp_instr("text", r"[^@\s]+@[^@\s]+", 0).alias("email_pos")
```
YoungVor: since there is different polars code for it, could you add some _instr tests where the pattern and idx are columns?

tests/_backends/local/functions/test_string_functions.py
YoungVor: Also, since there is different polars code for it, could you add some tests where idx is a column?

src/fenic/core/_logical_plan/signatures/text.py
bcallender: Yeah this is where having something like implicit casting would be helpful.

src/fenic/api/functions/text.py
bcallender: so in general i don't put the python types in the docstring -- when the doc generation runs it will expand the types automatically, and the in-IDE docs will show the types as part of the function definition. I can expand on the column data types that are accepted though.

src/fenic/api/functions/text.py
bcallender: ah good catch. let me just make it also return empty string.

src/fenic/api/functions/text.py
bcallender: Yeah I didn't catch that, very weird. `regexp_extract` is much older than `regexp_extract_all` in the library, which is maybe part of it? I decided to simplify by making idx a non expression, which makes the polars translation a lot easier as well.

tests/_backends/local/functions/test_string_functions.py
bcallender: good catch! there is an issue when that code path is taken around the index, will need to fix.

src/fenic/_backends/local/transpiler/expr_converter.py
bcallender: yeah, this implementation was a mess. I took another pass at it and decided to just implement the dynamic case in rust directly with a plugin.

src/fenic/api/functions/text.py
YoungVor: Ah, TIL.  That makes sense.  Thanks!

>> I can expand on the column data types that are accepted though.
not sure what you mean, do you mean there are column data types that are exceptions in the doc generation?

src/fenic/_backends/local/transpiler/expr_converter.py
YoungVor: nice

src/fenic/_backends/local/transpiler/expr_converter.py
bcallender: I'm not sure what you mean here? I thought `corresponding to the regex group index.` was the index of the group in the regex

bcallender: I also replaced this implementation with a pure rust implementation, so should be decently performant 👍

src/fenic/api/functions/text.py
bcallender: Oh I mean i can clarify what the `DataType` of the column has to be -- like if the column itself must be a string.

YoungVor: right, makes sense

src/fenic/api/functions/text.py
```diff
@@ -640,6 +645,254 @@ def regexp_replace(
...
+        pattern_expr = lit(pattern)._logical_expr
+    return Column._from_logical_expr(
+        RegexpExtractExpr(
+            Column._from_col_or_name(src)._logical_expr, pattern_expr, idx
+        )
+    )
+
+
+@validate_call(config=ConfigDict(strict=True, arbitrary_types_allowed=True))
+def regexp_extract_all(
```
YoungVor, line 745: haha, I'm not sure why for regexp_extract, idx as a column isn't necessary, but a column idx is desirable for extract_all and instr. pyspark really is confounding sometimes

rust/src/regex/mod.rs
```diff
@@ -9,3 +13,147 @@ pub fn py_validate_regex(regex: &str) -> PyResult<()> {
...
+        let idx_opt = idx_series.get(if idx_is_literal { 0 } else { i });
+
+        let value = match (text_opt, pattern_opt, idx_opt) {
+            (Some(text), Some(pattern), Some(idx)) => {
+                // Validate index is non-negative
+                if idx < 0 {
+                    Some(Series::new_empty(PlSmallStr::EMPTY, &DataType::String))
+                } else {
+                    // Get or compile regex, return error if invalid
+                    if !regex_cache.contains_key(pattern) {
```
YoungVor: nit: consider creating a helper function for duplicate regex map creation code

rust/src/regex/mod.rs
```diff
@@ -9,3 +13,147 @@ pub fn py_validate_regex(regex: &str) -> PyResult<()> {
...
+        let idx_opt = idx_series.get(if idx_is_literal { 0 } else { i });
+
+        let value = match (text_opt, pattern_opt, idx_opt) {
+            (Some(text), Some(pattern), Some(idx)) => {
+                // Validate index is non-negative
+                if idx < 0 {
+                    Some(0)
+                } else {
+                    // Get or compile regex, return error if invalid
+                    if !regex_cache.contains_key(pattern) {
```
YoungVor: nit: claude tells me rust provides a way to turn the lookup/insert into an enum, so there is only ever one lookup:

let regex = match regex_cache.entry(pattern.to_string()) {
    Entry::Occupied(e) => e.into_mut(),
    Entry::Vacant(e) => {
        match Regex::new(pattern) {
            Ok(re) => e.insert(re),
            Err(err) => { /* error handling */ }
        }
    }
};

pretty minor performance gain, but kind of cool

rust/src/regex/mod.rs
```diff
@@ -9,3 +13,147 @@ pub fn py_validate_regex(regex: &str) -> PyResult<()> {
...
+                    // Find all matches
+                    for captures in regex.captures_iter(text) {
+                        // idx=0 is whole match, idx=1+ are capture groups
+                        if let Some(matched) = captures.get(idx_usize) {
+                            matches.push(matched.as_str());
+                        }
+                    }
+
+                    // Return as Series
+                    Some(Series::from_iter(matches))
```
YoungVor: nit: claude says you can be more explicit about the type to avoid type inference:

Some(StringChunked::from_iter_values(
    PlSmallStr::EMPTY, 
    matches.into_iter()
).into_series())

rust/src/regex/mod.rs
```diff
@@ -9,3 +13,147 @@ pub fn py_validate_regex(regex: &str) -> PyResult<()> {
         Err(error) => Err(PyValueError::new_err(error.to_string())),
     }
 }
+
+#[polars_expr(output_type=Int32)]
```
YoungVor, line 37: For all of these functions, we could check that the group idx isn't negative, or group idx isn't higher than the total captures in the regex, and bail early.

Its a little annoying since there are regexp capture functions that live in python, but maybe it could be a plugin, similar to py_validate_regex?

YoungVor, line 37: I know there are python unit tests, but might be worth having some rust unit tests too

YoungVor: Up to you.. if they're easy to generate

tests/_backends/local/functions/test_regexp_functions.py
```diff
@@ -0,0 +1,480 @@
...
+    result = df.select(
+        text.regexp_extract_all(col("text"), col("pattern"), col("idx")).alias("matches")
+    ).to_polars()
+
+    matches_list = result["matches"].to_list()
+    assert matches_list[0] == []
+    assert matches_list[1] == []
+    assert matches_list[2] == []
+
+def test_regexp_extract_all_with_nested_capture_groups(local_session):
```
YoungVor, line 241: thank you for adding this!

bcallender: no worries! i learned something about how nested capture groups worked, so that was...fun 😂 . I feel much better about the implementation with these tests and the various fixes they provoked, thanks!

src/fenic/_backends/local/transpiler/expr_converter.py
bcallender: We discussed this offline -- added the `test_regexp_extract_all_with_nested_capture_groups` test to demonstrate that the behavior is inline with expectations when there are ambiguous or confusing match patterns for nested capture groups.

src/fenic/api/functions/text.py
bcallender: these functions are quite new, pyspark 3.5.0 I think, so i think the way the API is designed has changed. It's odd though, in pyspark, instead of 0 being the default, its `None`, but it's still 1-indexed to align with the existing functions, which is just...confusing.

rust/src/regex/mod.rs
bcallender: the issue is that the group idx can differ for each row, which is why i check if it's when we generate the regex. I'm not sure what the second part of the comment means here.

bcallender: added some for the utility functions

YoungVor (approved):
Looks clean and awesome Brandon, thank you for dealing with all the nits! :D

bcallender:
### Merge activity

* **Nov 19, 11:54 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/typedef-ai/fenic/247).
* **Nov 19, 11:54 PM UTC**: @bcallender merged this pull request with [Graphite](https://app.graphite.com/github/pr/typedef-ai/fenic/247).

PR #275, opened 2025-11-22
feat: Add remove_stopwords and detect_language (Part 1 of Issue #274)

Aaryan-549:
## Summary
This PR implements the first half of **Issue #274**, focusing on lightweight NLP preprocessing functions.

## Changes
- Implemented `text.remove_stopwords()` (supports standard ISO 639-1 languages).
- Implemented `text.detect_language()` (with optional confidence scores).
- Added comprehensive unit tests in `tests/api/functions/test_text.py`.
- Updated `pyproject.toml` to handle optional NLP dependencies.

## Implementation Notes
I explicitly split this feature into two PRs to manage dependency weight:
1.  **This PR:** Lightweight operations (Stopwords/Language Detection).
2.  **Follow-up PR:** Heavy operations (Lemmatization/Stemming) which require larger libraries like `spaCy`.

## Checklist
- [x] Tests added
- [x] Documentation updated
- [x] Linter passed


src/fenic/api/functions/text.py
```diff
@@ -1430,3 +1430,176 @@ def compute_fuzzy_token_set_ratio(column: ColumnOrName, other: Union[Column, str
...
+    Args:
+        column: Input text column to process
+        language: ISO 639-1 language code. Supported languages:
+            - "en" (English)
+            - "es" (Spanish)
+            - "fr" (French)
+            - "de" (German)
+            - "it" (Italian)
+            - "pt" (Portuguese)
+        custom_stopwords: Optional list of custom stopwords to use instead of language-based
```
YoungVor, line 1484: The customization is a good idea.  Should we take it a step further?  Somehow give the user the option to either
* add stopwords to the default list
* replace the default stopwords list (like you do here)

tests/_backends/local/functions/test_nlp_functions.py
```diff
@@ -0,0 +1,290 @@
...
+
+
+def test_remove_stopwords_english(stopwords_test_df):
+    """Test English stopword removal."""
+    result = stopwords_test_df.select(
+        text.remove_stopwords(col("text"))
+    ).to_polars()
+
+    # "The", "over", "the" should be removed
+    assert "The" not in result["text"][0]
```
YoungVor: I think it would be better both for readability and for completeness testing to define the expected output strings as a const near where you define the input strings, and just assert the answer is equivalent

tests/_backends/local/functions/test_nlp_functions.py
```diff
@@ -0,0 +1,290 @@
+"""Tests for NLP text preprocessing functions."""
+
+import pytest
+
+from fenic import col, text
+from fenic.core.error import ValidationError
+
+
+@pytest.fixture
+def stopwords_test_df(local_session):
```
YoungVor, line 18: would be great to include a test with simple markdown column

rust/src/nlp/stopwords.rs
```diff
@@ -0,0 +1,157 @@
+use std::collections::{HashMap, HashSet};
+use std::sync::LazyLock;
+
+/// Stopwords for multiple languages
+/// Based on common NLP stopword lists
```
YoungVor: Can you go into more detail here?  For e.g., are these taken from a commonly used NLTK list?  Would be good to add that detail to the api docstring as well

protos/logical_plan/v1/expressions.proto
```diff
@@ -662,6 +668,25 @@ message MdExtractHeaderChunks {
   int32 header_level = 2;
 }
 
+// NLP text preprocessing expressions
+message RemoveStopwordsExpr {
+  LogicalExpr expr = 1;
+  LogicalExpr language = 2;
+}
+
+message RemoveCustomStopwordsExpr {
```
YoungVor: Any reason not to have a single Logical and Physical expr node with optional language and stopwords arguments?  (or just send empty stopwords if there's no customization)

src/fenic/_backends/local/transpiler/expr_converter.py
```diff
@@ -1331,6 +1335,30 @@ def _convert_md_chunk_by_headings_expr(self, logical: MdExtractHeaderChunks) ->
...
+    @_convert_expr.register(RemoveCustomStopwordsExpr)
+    def _convert_remove_custom_stopwords_expr(self, logical: RemoveCustomStopwordsExpr) -> pl.Expr:
+        physical_expr = self._convert_expr(logical.expr)
+        stopwords_expr = self._convert_expr(logical.stopwords)
+        return physical_expr.nlp.remove_custom_stopwords(stopwords_expr)
+
+    @_convert_expr.register(DetectLanguageExpr)
+    def _convert_detect_language_expr(self, logical: DetectLanguageExpr) -> pl.Expr:
+        physical_expr = self._convert_expr(logical.expr)
+        return physical_expr.nlp.detect_language()
```
YoungVor, line 1340: I think you could have just one rust implementation of detect languages, and if the confidence option is not toggled, just return the 'language' field

physical_expr.nlp.detect_language().struct.field("language")

src/fenic/_backends/local/polars_plugins/nlp.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def __init__(self, expr: pl.Expr) -> None:
+        """Initialize an NLP Namespace with a Polars expression.
+
+        Args:
+            expr: A Polars expression containing the text data for NLP operations.
+        """
+        self.expr = expr
+
+    def remove_stopwords(self, language: IntoExpr) -> pl.Expr:
```
YoungVor, line 22: Can we have one function for remove_stopwords, and one for detect_languages?

protos/logical_plan/v1/expressions.proto
```diff
@@ -662,6 +668,25 @@ message MdExtractHeaderChunks {
   int32 header_level = 2;
 }
 
+// NLP text preprocessing expressions
+message RemoveStopwordsExpr {
+  LogicalExpr expr = 1;
+  LogicalExpr language = 2;
```
YoungVor, line 674: This should just be string.  LogicalExpr is for an expression that will be evaluated further up the plan/dag, but you are just passing a string in the api, not a Column.

If you want to make language, or the customized list a Column so the user can change the list based on the row you can do that too, though I do think static input like you have is fine for this usecase.

protos/logical_plan/v1/expressions.proto
```diff
@@ -662,6 +668,25 @@ message MdExtractHeaderChunks {
...
+}
+
+message RemoveCustomStopwordsExpr {
+  LogicalExpr expr = 1;
+  LogicalExpr stopwords = 2;
```
YoungVor: stopwords I think can just be a repeated string

src/fenic/api/functions/text.py
```diff
@@ -1430,3 +1430,176 @@ def compute_fuzzy_token_set_ratio(column: ColumnOrName, other: Union[Column, str
...
+        ```
+    """
+    from fenic.core._logical_plan.expressions.nlp import RemoveCustomStopwordsExpr, RemoveStopwordsExpr
+
+    column_expr = Column._from_col_or_name(column)._logical_expr
+
+    if custom_stopwords is not None:
+        # Use only custom stopwords (language parameter is ignored in this branch)
+        stopwords_list = list(custom_stopwords)
+        stopwords_expr = LiteralExpr(stopwords_list, ArrayType(StringType))
```
YoungVor, line 1554: If you just want a single value for the operation, you don't need to convert this to a literal expression, you can simply put the python value (stopwords list, language string) directly to the logical node.

Converting to expressions is useful if you want to give the user the ability to pass a column(str), not just a string

src/fenic/api/functions/text.py
```diff
@@ -1430,3 +1430,176 @@ def compute_fuzzy_token_set_ratio(column: ColumnOrName, other: Union[Column, str
         other_expr = other._logical_expr
 
     return Column._from_logical_expr(FuzzyTokenSetRatioExpr(Column._from_col_or_name(column)._logical_expr, other_expr, method))
+
+
+# NLP Text Preprocessing Functions
+
+@validate_call(config=ConfigDict(strict=True, arbitrary_types_allowed=True))
+def remove_stopwords(column: ColumnOrName, language: str = "en", custom_stopwords: Optional[List[str]] = None) -> Column:
```
YoungVor: Does this work only for String columns, or does it work for Markdown and Json columns?

src/fenic/api/functions/text.py
```diff
@@ -1430,3 +1430,176 @@ def compute_fuzzy_token_set_ratio(column: ColumnOrName, other: Union[Column, str
...
+            - "fr" (French)
+            - "de" (German)
+            - "it" (Italian)
+            - "pt" (Portuguese)
+        custom_stopwords: Optional list of custom stopwords to use instead of language-based
+            stopwords (case-insensitive). When provided, the language parameter is ignored
+            and only the specified custom stopwords are removed.
+
+    Returns:
+        Text column with stopwords removed, separated by single spaces
```
YoungVor: Will it collapse multiple removed stop words into one space? ( for e.g.: `subject is all about any below tasks` would condense to `subject tasks` instead of `subject      tasks`)

YoungVor (changes requested):
Thank you for putting this together Aaryn, I'm excited to see these changes in Fenic!

I left some comments, but my main feedback is around:
1. There's a lot of potential reduce the interface surface areas and deduplicate (single expression with flags rather than two separate expressions, single rust functions with flags, etc).  You have the right idea keeping it simple on the API layer by having a single call with flags - you should be able to plumb that down the stack
2. For the API - do you intend to allow users to include Columns for their stopword customizations, or keep it static for the operation?

Aaryan-549:
Thanks for the detailed feedback @YoungVor! I appreciate the suggestions on simplifying the architecture.

> Thank you for putting this together Aaryn, I'm excited to see these changes in Fenic!
> 
> I left some comments, but my main feedback is around:
> 
> 1. There's a lot of potential reduce the interface surface areas and deduplicate (single expression with flags rather than two separate expressions, single rust functions with flags, etc).  You have the right idea keeping it simple on the API layer by having a single call with flags - you should be able to plumb that down the stack

I structured it with separate `RemoveStopwordsExpr` and `RemoveCustomStopwordsExpr` to follow the pattern I saw in other parts of the codebase (e.g., separate expressions for different text operations). However, I completely agree that a single expression with optional parameters would be cleaner.

Would you prefer I refactor to:
```python
# Single logical expression
RemoveStopwordsExpr(expr, language=Optional[str], custom_stopwords=Optional[LogicalExpr])

**Comment by Aaryan-549** [comment]
>2. For the API - do you intend to allow users to include Columns for their stopword customizations, or keep it static for the operation?

Great question! The stopword lists are based on commonly used NLP corpora, similar to NLTK's stopwords collections. They're manually curated high-frequency words (100-200+ per language) that typically add little semantic value.

I've updated the API docstring to clarify this (see latest commit).

File: protos/logical_plan/v1/expressions.proto (hunk 9 above)
Comment (Aaryan-549) [reply protos/logical_plan/v1/expressions.proto]: You're absolutely right - using `string` for language and `repeated string` for stopwords would be much simpler. I initially used `LogicalExpr` thinking it would allow future flexibility for Column inputs, but static strings make more sense for the common case.

Should I refactor the proto to:
```protobuf
message RemoveStopwordsExpr {
  LogicalExpr expr = 1;
  string language = 2;  // Instead of LogicalExpr
  repeated string stopwords = 3;  // Instead of LogicalExpr
}

File: src/fenic/_backends/local/transpiler/expr_converter.py (hunk 6 above)
Comment (Aaryan-549) [reply src/fenic/_backends/local/transpiler/expr_converter.py:1340]: Yeah, I think having one Rust function that always returns the struct, then extracting just the language field in Python when `return_confidence=False` is going to be much cleaner.

I'll refactor to:
- Single `detect_language()` Rust function returning struct  
- Python wrapper extracts `.struct.field("language")` when confidence not needed

This eliminates `DetectLanguageWithConfidenceExpr` and simplifies the code significantly.

File: src/fenic/_backends/local/polars_plugins/nlp.py (hunk 7 above)
Comment (Aaryan-549) [reply src/fenic/_backends/local/polars_plugins/nlp.py:22]: Agreed - the polars plugin layer can have just:
- `nlp.remove_stopwords(language, stopwords)` - single function  
- `nlp.detect_language()` - always returns struct

The Python API can handle the conditional logic for which arguments to pass.

File: src/fenic/api/functions/text.py (hunk 11 above)
Comment (Aaryan-549) [reply src/fenic/api/functions/text.py]: Currently it works on String columns. Markdown and JSON columns would work if they're already string-typed, but there's no special handling for their structure.

The stopword removal operates on whitespace-tokenized text, so it should work on any text content. Should I add explicit tests/documentation for Markdown and JSON column behavior?

File: src/fenic/api/functions/text.py (hunk 12 above)
Comment (Aaryan-549) [reply src/fenic/api/functions/text.py]: Yes, multiple consecutive removed stopwords collapse into a single space. For example:

Input: `"subject is all about any below tasks"`  
Output: `"subject tasks"` (not `"subject      tasks"`)

I've clarified this in the documentation (see latest commit).

Comment (YoungVor) [reply src/fenic/api/functions/text.py]: great!  Thank you

File: tests/_backends/local/functions/test_nlp_functions.py [hunk 13]
```diff
@@ -288,3 +290,32 @@ def test_detect_language_short_text(local_session):
...
+## Another section
+
+More text with stopwords."""
+
+    data = {"markdown": [markdown_text]}
+    df = local_session.create_dataframe(data)
+
+    # Convert to string and remove stopwords
+    result = df.select(
+        text.remove_stopwords(col("markdown")).alias("cleaned")
```
YoungVor, line 335: Does it work if you pass MarkdownType? I.e., pass col("markdown").cast(MarkdownType)

YoungVor:
> Thanks for the detailed feedback @YoungVor! I appreciate the suggestions on simplifying the architecture.
> 
> > Thank you for putting this together Aaryn, I'm excited to see these changes in Fenic!
> > I left some comments, but my main feedback is around:
> > 
> > 1. There's a lot of potential reduce the interface surface areas and deduplicate (single expression with flags rather than two separate expressions, single rust functions with flags, etc).  You have the right idea keeping it simple on the API layer by having a single call with flags - you should be able to plumb that down the stack
> 
> I structured it with separate `RemoveStopwordsExpr` and `RemoveCustomStopwordsExpr` to follow the pattern I saw in other parts of the codebase (e.g., separate expressions for different text operations). However, I completely agree that a single expression with optional parameters would be cleaner.
> 
> Would you prefer I refactor to:
> 
> ```python
> # Single logical expression
> RemoveStopwordsExpr(expr, language=Optional[str], custom_stopwords=Optional[LogicalExpr])
> ```

Yes please!  We've evolved our standards as we went, so it doesn't surprise me if we have some older code we could make cleaner in future

YoungVor:
> > 2. For the API - do you intend to allow users to include Columns for their stopword customizations, or keep it static for the operation?
> 
> Great question! The stopword lists are based on commonly used NLP corpora, similar to NLTK's stopwords collections. They're manually curated high-frequency words (100-200+ per language) that typically add little semantic value.
> 
> I've updated the API docstring to clarify this (see latest commit).

Thank you!  That's quite clear.  Please also add the comment to the stopwords.rs

src/fenic/api/functions/text.py
YoungVor: It would be good to include that case in a unit test and check the single space

src/fenic/_backends/local/transpiler/expr_converter.py
YoungVor: Sounds perfect!

And I think you can do a similar refactor to have a single remove_stopwords rust function taking an optional list of custom words

src/fenic/api/functions/text.py
YoungVor: I see you added the test, but to clarify, I meant accepting a column of MarkdownType or JsonType as well as StringType.  There might be some gotchas to getting the expression to take MarkdownType, so I'm fine with keeping it as StringType in this PR.  User can always cast to string first

src/fenic/api/functions/text.py
YoungVor: Have you given thought to this point?  It could be interesting to allow the user to pass row-specific stopword lists or languages, but keeping it as a static function-level input I think is fine.  In the latter case, you can pass the python value (without converting to an expression)

src/fenic/core/_logical_plan/expressions/nlp.py
```diff
@@ -0,0 +1,80 @@
...
+
+class RemoveStopwordsExpr(ValidatedSignature, UnparameterizedExpr, LogicalExpr):
+    """Remove stopwords from text based on language."""
+
+    function_name = "text.remove_stopwords"
+
+    def __init__(self, expr: LogicalExpr, language: LogicalExpr):
+        self.expr = expr
+        self.language = language
+        self._validator = SignatureValidator(self.function_name)
```
YoungVor: You will need to add the function signatures to the registry in _logical_plan/signatures/text.py.

Did you run the entire 'local' unit test suite?  I think the test_registry.py test should fail if it catches new functions with unregistered signatures.

YoungVor (commented):
Second pass, thank you for addressing some of the concerns!  I think you need to register the function signatures (only two if you refactor both to take optionals), and I think its fine to keep the signature as StringType

Aaryan-549:
Hi @YoungVor,

Sorry for the delay, I had Endsems going on haha. Thank you for the detailed feedback! I've implemented all the changes you requested:

### 1. Unified Expression Architecture 
Consolidated `RemoveStopwordsExpr` and `RemoveCustomStopwordsExpr` into a single unified expression with an optional `custom_stopwords` parameter. This simplifies the API and makes the codebase more maintainable.

**Updated files:**
- `rust/src/nlp/mod.rs` - Single `remove_stopwords` function handling both cases
- `protos/logical_plan/v1/expressions.proto` - Updated `RemoveStopwordsExpr` with optional field
- `src/fenic/core/_logical_plan/expressions/nlp.py` - Unified Python expression class
- `src/fenic/core/_serde/proto/expressions/nlp.py` - Updated serialization handlers
- `src/fenic/api/functions/text.py` - Unified public API
- `src/fenic/_backends/local/transpiler/expr_converter.py` - Updated transpiler

### 2. Function Signatures Added to Registry 
Added proper function signatures for `text.remove_stopwords` and `text.detect_language` to the registry in `src/fenic/core/_logical_plan/signatures/text.py`. This ensures proper type validation.

### 3. Documentation Added 
Added comprehensive documentation to `rust/src/nlp/stopwords.rs` explaining:
- Source and curation of stopword lists (similar to NLTK's collections)
- What stopwords are and their purpose
- Supported languages
- High-frequency word characteristics

### 4. Single Space Collapse Test 
Added test case `test_remove_stopwords_single_space_collapse` in `tests/_backends/local/functions/test_nlp_functions.py` to verify that consecutive removed stopwords collapse to a single space (e.g., "subject is all about any below tasks" → "subject tasks").

bcallender:
Closing this as part of a PR backlog prune. Stopword removal and language detection may still be useful, but this branch is stale against current `main`, large, and had unresolved requested changes. This is not a rejection of the idea itself; it needs a current owner and a fresh implementation shape before it can land safely. Thanks for the substantial work here.

PR #313, opened 2026-06-09
feat: adaptive output-token estimation + settlement for rate limiting

bcallender:
## Problem

The client-side rate limiter reserves output tokens at a worst-case ceiling (`max_completion_tokens` + a static reasoning budget — e.g. +8192 for medium reasoning, 1.5× thinking budget on Gemini, up to 0.6× max output on OpenRouter) for **every** row, and never reconciles the reservation against actual usage. Actual token counts come back on every response but flow only to metrics. Result: the TPM bucket is chronically over-reserved and throughput sits far below the provider's real limit.

## What this does

Three coordinated changes, all completions-only, no scheduler changes:

1. **Settlement (the backbone).** New `RateLimitStrategy.settle(reserved, actual)`: when a response completes, the bucket is corrected by `reserved − actual`. Reconciles the bucket to actual usage in both directions — refunding the over-reservation (the common case) and debiting further when a request runs past it (the p95 tail) — and neither increases 429 risk (a refund returns capacity the provider never charged; a debit only tightens the limiter). Always on.
2. **Adaptive estimation (the accelerator).** A per-client `OutputTokenEstimator` learns the distribution of actual output tokens per `(profile, max_completion_tokens)` and reserves p95 (p99 for reasoning profiles) × `safety_margin` instead of the static ceiling — clamped ≤ the old ceiling, so reservations only ever shrink. Anthropic's separated input/output TPM buckets are settled independently (input reconciled against actual `prompt_tokens`, correcting the static 1.05× tokenizer fudge as a bonus).
3. **Decoupling.** The learned value affects only the rate-limiter reservation. The API `max_tokens` cap path is untouched — zero truncation risk.

## Measured results

**Live API** (`benchmarks/rate_limit_stress.py`: gpt-4.1-nano, tpm=20k, naive `max_output_tokens=2048`, 40 rows of a varied-output extraction task averaging ~12 actual output tokens/row):

| Mode | Wall-clock | Reserved out tokens | Reservation efficiency |
| --- | --- | --- | --- |
| Pre-feature (ceiling reserve, no settle) | 207.6s | 81,920 | 0.006 |
| Settlement only (`enabled=False`) | 3.2s | 81,920 | 0.006 |
| Adaptive + settlement (post-warm-up) | **1.2s** | **1,160** | **0.409** |

**174.6× end-to-end**, identical outputs, identical cost ($0.0009/phase), zero provider 429s. Settlement delivers the bulk (65×); adaptive estimation tightens reservations ~70× and adds another ~2.7× by admitting the whole batch in one burst. The hermetic harness (`tests/_inference/rate_limit_harness/`) shows the same shape in simulation (23× / 0.40→0.79 efficiency).

## Error-classification fixes (found via this work)

- **OpenAI embeddings:** `insufficient_quota` 429s were classified Transient and burned the full exponential-backoff budget (~5 min of sleeps per batch) — a drained account turned a ~16-min CI suite into a 2-hour crawl. Now Fatal (mirrors the chat path).
- **Anthropic:** 429s carrying `x-should-retry: false` (hard spend-cap breach) are now Fatal; plain per-minute 429s stay Transient. 529 `overloaded_error` is now correctly Transient — the async anthropic SDK (0.54.0) maps 529 to `InternalServerError` (no 529 branch, unlike the sync client), so it previously fell into the `AnthropicError` catch-all and **permanently failed** the request.

## Config

Default-on with one dial, mirroring the `llm_response_cache` wiring:

```python
SemanticConfig(..., adaptive_token_estimation=AdaptiveTokenEstimationConfig(safety_margin=1.3))
# or disable adaptive estimation (settlement stays on):
AdaptiveTokenEstimationConfig(enabled=False)
```

`None` = enabled with defaults (`safety_margin=1.15`). New `LMMetrics.num_reserved_output_tokens` exposes reservation efficiency.

## Safety / compatibility

- Reservation always ≤ the previous static ceiling; cold start (<30 samples) and small batches transparently use the old ceiling; a non-positive ceiling (`max_completion_tokens=None`, no file estimate) is preserved exactly.
- API caps, retries, backoff, dedup, embeddings reservation paths: unchanged.
- Reconciliation is post-success bookkeeping, defensively wrapped — can never fail a completed request.
- Known deferral: settle-after-backoff re-injection during 429 storms measured at ~0.4% of bucket/event; tracked in TD-3375 with a fix sketch.

## Testing

- 13 hermetic test files (estimator incl. zero-ceiling, settlement incl. separated-bucket clamps, config, wiring, per-provider routing for OpenAI/Anthropic/Gemini/OpenRouter, thinking-token settlement, disabled-path regression, OpenAI quota + Anthropic 429/529 classification) — no API keys needed.
- `tests/_inference/rate_limit_harness/`: simulated-provider perf harness driving the real queue/backoff/settle path.
- `benchmarks/rate_limit_stress.py`: manual live-API benchmark (never collected by CI) — source of the numbers above.

Design doc: `specs/adaptive_token_estimation_design.md` (includes both measured-results tables and the pre-existing-issues register).


src/fenic/_inference/output_token_estimator.py
```diff
@@ -0,0 +1,68 @@
...
+        if not self._enabled:
+            return static_ceiling
+        with self._lock:
+            samples = self._samples.get(key)
+            if samples is None or len(samples) < self._min_samples:
+                return static_ceiling
+            ordered = sorted(samples)
+        q = _REASONING_QUANTILE if reasoning else _DEFAULT_QUANTILE
+        modeled = _quantile(ordered, q) * self._safety_margin
+        return max(1, min(static_ceiling, math.ceil(modeled)))
```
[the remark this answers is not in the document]
bcallender: > `OutputTokenEstimator.reserve()` can return 1 even when `static_ceiling` is 0 because of `max(1, min(static_ceiling, ...))`. ... Clamp to `[0, static_ceiling]` and short-circuit non-positive ceilings.

Addressed in ab7d6ad: `reserve()` now short-circuits non-positive ceilings (`if static_ceiling <= 0: return static_ceiling`) before any estimation, so a 0 ceiling (e.g. `max_completion_tokens=None` with no file estimate and no reasoning budget) reserves exactly 0 — matching the pre-feature static path. The docstring was updated to state the contract, and `test_zero_ceiling_preserved_even_when_warm` regression-guards it (warm estimator, 0 ceiling, asserts 0).

bcallender:
* **#313** <a href="https://app.graphite.com/github/pr/typedef-ai/fenic/313?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/typedef-ai/fenic/313?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/fenic/_inference/model_client.py
```diff
@@ -290,6 +289,56 @@ def reset_metrics(self):
         """Reset all metrics for this model client to their initial values."""
         pass
 
+    def _estimator_key(self, request: RequestT) -> tuple[Optional[str], Optional[int]]:
+        """Key for the output-token estimator: (profile_hash, max_completion_tokens)."""
+        max_tokens = getattr(request, "max_completion_tokens", None)
+        return (self.get_profile_hash_for_request(request), max_tokens)
```
YoungVor: nit: key is `(profile_hash, max_completion_tokens)` only — distinct ops at the same profile+max_tokens share one distribution, so p95 over a bimodal mix under-reserves the heavier op. Bounded by the static-ceiling clamp; a code comment would help.

src/fenic/_inference/model_client.py
```diff
@@ -290,6 +289,56 @@ def reset_metrics(self):
...
+        whether adaptive estimation is enabled. Settlement is pure accounting that
+        can only refund unused reservation — it never increases 429 risk.
+        ``enabled=False`` only affects the up-front *reservation size* (falls back
+        to the static ceiling); the settle() call always executes.
+        """
+        if reserved.total_tokens <= 0:
+            return
+        # Record the reservation unconditionally so the metric is populated even
+        # if observe() or settle() raise below.
+        self.get_metrics().num_reserved_output_tokens += reserved.output_tokens
```
YoungVor, line 366: nit: incremented only on success, but each retried dispatch re-consumes the bucket — so reservation-efficiency reads better than reality under retries.

YoungVor:
nit (PR description): "settlement … can only refund … never increases 429 risk" — `settle()` has a debit branch when `actual > reserved` (the p95 tail), as its own docstring notes ("debits further when under-reserved"). Worth tightening the wording.

YoungVor (approved):
Spoke in person about how decoupling the estimates for rate-limit reasoning effort from visible budget for openrouter is unnecessary, since we don't actually throttle OpenRouter completions with tpm.

Other than a few minor nits my friend claude found, looks good!

src/fenic/_inference/model_client.py
bcallender: > distinct ops at the same profile+max_tokens share one distribution, so p95 over a bimodal mix under-reserves the heavier op. Bounded by the static-ceiling clamp; a code comment would help.

Addressed in 8a43673. Rather than just a comment, added `structured_output.schema_fingerprint` (already on the request — it's the response cache key) to the key, so structured ops with different schemas, and structured vs unstructured ops at the same `max_completion_tokens`, no longer pool. Deliberately did NOT add `operation_name`: it's not on the request, and finer keys fragment the sample pool, slowing the `min_samples=30` warm-up the estimator depends on.

One correction to the bound: the static-ceiling clamp caps *over*-reservation only — it does not bound under-reservation. Under-reservation of a minority heavy op is absorbed by the **retry queue** (settlement refunds on success, so it can't prevent an in-flight 429); always-on settlement keeps the bucket honest. Net: a misfit only dampens the adaptive accelerator, never the settlement backbone. The docstring now spells this out, and points at per-row input-aware estimation as the better future lever.

src/fenic/_inference/model_client.py
bcallender: > incremented only on success, but each retried dispatch re-consumes the bucket — so reservation-efficiency reads better than reality under retries.

Correct — confirmed. The bucket is debited per dispatch attempt (incl. retried ones, same frozen `estimated_tokens`), but `num_reserved_output_tokens` is counted once on eventual success, so `actual / reserved` reads optimistic under retry storms. Added a comment at the increment documenting this and cross-referencing TD-3375 (the retry N-consume / 1-settle issue), where the bucket-side fix lives. Left the metric per-success so it stays paired with `num_output_tokens` (both per-logical-request) for the efficiency ratio.

bcallender:
> "settlement … can only refund … never increases 429 risk" — `settle()` has a debit branch when `actual > reserved` (the p95 tail) … Worth tightening the wording.

Good catch — fixed in 7bf9588 across the PR description, the `AdaptiveTokenEstimationConfig` docstring, and `_reconcile_completion`. They now say settlement reconciles in **both** directions: refunds the over-reservation (common case) and debits further when a request runs past its reservation (the p95 tail). The "never increases 429 risk" conclusion stands, just correctly justified — a refund only returns capacity the provider never charged, and a debit only makes the limiter more conservative.

PR #314, opened 2026-06-10
test: add semantic operator execution guardrails

bcallender:
## Summary

This PR adds local semantic-operator guardrails that make the execution path easier to verify before we tune runtime behavior. It locks in representative golden outputs, adds a deterministic semantic.reduce → model-client tripwire, and introduces a repeatable peak-RSS benchmark protocol so future memory-footprint improvements can cite comparable before/after evidence instead of relying on ad hoc observations.

The branch also includes the semantic.reduce compatibility fix needed after the Polars stack update changed the grouped `map_batches` callback shape. Follow-up review tightened that fix to avoid Python `to_list()`/DataFrame round-trips in the hot preprocessing path.

## Linear issues

- TD-3317: M0 golden-output correctness tests for semantic operators
- TD-3319: M0 reduce→model_client integration API tripwire
- TD-3316: M0 peak-memory regression harness for semantic operators
- TD-3381: M0 semantic memory benchmark matrix and PR evidence protocol

## What changed

- Added golden-output coverage for semantic.join, semantic.sim_join, and semantic.reduce local execution behavior.
- Added a deterministic semantic.reduce integration tripwire proving the operator reaches the LanguageModel / ModelClient completions API with the expected request shape.
- Fixed semantic.reduce for current Polars grouped `map_batches` callback behavior while preserving compatibility with direct Series-of-Series callers.
- Avoided unnecessary Python materialization in semantic.reduce preprocessing by using direct struct-field extraction for the no-context/no-sort path and `struct.unnest()` when a DataFrame is needed.
- Added a process-isolated semantic-operator memory harness plus focused tests.
- Documented the smoke-test versus evidence-grade benchmark matrix and PR evidence protocol for memory-improvement claims.

## Evidence / validation

- [x] `git diff --check`
- [x] `trunk check --fix --no-progress src/fenic/_backends/local/semantic_operators/reduce.py tests/_backends/local/functions/test_semantic_reduce.py` — checked 3 files, no issues.
- [x] `uv run pytest -q tests/_backends/local/semantic_operators/test_reduce.py tests/_backends/local/functions/test_semantic_reduce.py tests/test_benchmark_semantic_operator_memory.py` — 25 passed, 5 warnings.
- [x] Earlier PR package validation: `uv run --env-file /home/hermes/.hermes/.env pytest -q tests/_backends/local/dataframe/test_semantic_join.py tests/_backends/local/dataframe/test_semantic_sim_join.py tests/_backends/local/functions/test_semantic_reduce.py tests/test_benchmark_semantic_operator_memory.py` — 36 passed, 1 xfailed, 5 warnings before the later empty-input xfail cleanup.
- [x] `uv run python tools/benchmark_semantic_operator_memory.py --help` — help renders the benchmark protocol guidance.
- [x] `env -u OPENAI_API_KEY -u ANTHROPIC_API_KEY -u GEMINI_API_KEY uv run python tools/benchmark_semantic_operator_memory.py --cases semantic_reduce,map_extract_chain --rows 2 --right-rows 2 --groups 1 --label pr-package-offline-check --json` — both cases passed without provider credentials.

## Reviewer notes

- TD-3318/provider idle-gap instrumentation is intentionally excluded from this PR.
- The semantic.reduce empty-input golden-output case now passes and is no longer marked xfail.
- Cross-group semantic.reduce execution is now delegated to Polars grouped `map_batches` scheduling for the real execution path. Local fake-model review observed grouped callbacks still running concurrently; the explicit `Reduce.execute()` ThreadPoolExecutor is no longer the concurrency owner.
- `.hermes/` evidence and `.venv/` are preserved locally but ignored/untracked and not included in the PR.

## Release / risk

- Risk: Low/Medium — mostly test/tooling/protocol guardrails, plus a semantic.reduce runtime compatibility fix for current Polars grouped `map_batches` behavior. The compatibility path has focused semantic.reduce coverage and avoids the preprocessing overhead called out during review.
- Rollback: Revert this PR; no migration or production state change is required.

### DIFF
--- a/src/fenic/_backends/local/semantic_operators/reduce.py
+++ b/src/fenic/_backends/local/semantic_operators/reduce.py
@@ -155,7 +155,10 @@ class Reduce:
 
     def _preprocess_group(self, group: pl.Series) -> Tuple[str, pl.Series]:
         """Preprocess group by adding context and ordering the data column."""
-        group_df = pl.DataFrame(group.to_list())
+        if not self.group_context_names and not self.descending:
+            return self.user_instruction, group.struct.field(DATA_COLUMN_NAME)
+
+        group_df = group.struct.unnest()
 
         # Step 0: Build user instruction from group context if present
         if self.group_context_names:
--- a/src/fenic/_backends/local/semantic_operators/reduce.py
+++ b/src/fenic/_backends/local/semantic_operators/reduce.py
@@ -1,6 +1,5 @@
 import logging
 import math
-from concurrent.futures import ThreadPoolExecutor
 from textwrap import dedent
 from typing import List, Optional, Tuple
 
@@ -82,12 +81,25 @@ class Reduce:
             self._user_instruction_template = jinja2.Template(self.user_instruction)
 
     def execute(self) -> pl.Series:
-        """Execute parallel reduction across all groups."""
-        with ThreadPoolExecutor(max_workers=min(10, len(self.input))) as executor:
-            results = list(
-                executor.map(lambda x: self._reduce_group(*x), enumerate(self.input))
-            )
-        return pl.Series(results)
+        """Execute reduction over Polars group batches.
+
+        ``expr_converter`` calls this from ``map_batches`` after ``implode``. With
+        current Polars, the callback receives one Series of struct rows for the
+        current group, and ``returns_scalar=True`` expects one output value for
+        that group. Unit tests and direct callers may still pass a Series whose
+        items are grouped Series values; preserve that shape as multiple groups.
+        """
+        if len(self.input) == 0:
+            return pl.Series([None])
+
+        first_group = self.input[0]
+        if isinstance(first_group, pl.Series):
+            return pl.Series([
+                self._reduce_group(group_index, group)
+                for group_index, group in enumerate(self.input)
+            ])
+
+        return pl.Series([self._reduce_group(0, self.input)])
 
     def _reduce_group(self, group_index: int, group: pl.Series) -> str | None:
         """Reduces a single group of documents hierarchically until a single output is obtained.
@@ -143,9 +155,11 @@ class Reduce:
 
     def _preprocess_group(self, group: pl.Series) -> Tuple[str, pl.Series]:
         """Preprocess group by adding context and ordering the data column."""
+        group_df = pl.DataFrame(group.to_list())
+
         # Step 0: Build user instruction from group context if present
         if self.group_context_names:
-            first_row_data = group.first()
+            first_row_data = group_df.row(0, named=True)
             group_context = {name: first_row_data[name] for name in self.group_context_names}
             user_instruction = self._user_instruction_template.render(**group_context)
         else:
@@ -155,14 +169,7 @@ class Reduce:
         if self.descending:
             sort_column_names = [f"{SORT_KEY_COLUMN_NAME}_{i}" for i in range(len(self.descending))]
 
-            # Build a dict of sort key name -> Series extracted from struct field
-            sort_key_columns = {
-                name: group.struct.field(name)
-                for name in sort_column_names
-            }
-
-            # Create a DataFrame from those sort keys (no copy)
-            sort_keys_df = pl.DataFrame(sort_key_columns)
+            sort_keys_df = group_df.select(sort_column_names)
 
             # Compute sorted indices with arg_sort_by
             sorted_idx = sort_keys_df.select(
@@ -174,9 +181,9 @@ class Reduce:
             ).to_series()
 
             # Gather the data column in sorted order
-            sorted_data = group.struct.field(DATA_COLUMN_NAME).gather(sorted_idx)
+            sorted_data = group_df[DATA_COLUMN_NAME].gather(sorted_idx)
         else:
-            sorted_data = group.struct.field(DATA_COLUMN_NAME)
+            sorted_data = group_df[DATA_COLUMN_NAME]
 
         return user_instruction, sorted_data
 
--- a/tests/_backends/local/functions/test_semantic_reduce.py
+++ b/tests/_backends/local/functions/test_semantic_reduce.py
@@ -18,7 +18,7 @@ from fenic.api.session import (
     Session,
     SessionConfig,
 )
-from fenic.core.error import ExecutionError, PlanError, ValidationError
+from fenic.core.error import PlanError, ValidationError
 
 
 def _install_deterministic_reduce_model(local_session, monkeypatch):
@@ -175,10 +175,6 @@ def test_semantic_reduce_golden_output_ordering_and_nulls(local_session, monkeyp
     ]
 
 
-@pytest.mark.xfail(
-    raises=ExecutionError,
-    reason="semantic.reduce currently creates a ThreadPoolExecutor with max_workers=0 for empty grouped inputs.",
-)
 def test_semantic_reduce_golden_empty_result(local_session, monkeypatch):
     _install_deterministic_reduce_model(local_session, monkeypatch)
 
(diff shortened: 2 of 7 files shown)


tools/benchmark_semantic_operator_memory.py
```diff
@@ -0,0 +1,577 @@
...
+        k=args.k,
+        label=args.label,
+    )
+    case_results = [_run_case(case, config) for case in args.cases]
+    failures = [case for case in case_results if not case.get("ok")]
+    payload = {
+        "schema_version": 1,
+        "label": args.label,
+        "generated_at": dt.datetime.now(dt.UTC).isoformat(),
+        "git": {
```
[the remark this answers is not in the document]
bcallender: Fixed in 6ed4d2c: the harness now uses `datetime.timezone.utc` for Python 3.10 compatibility.

tools/benchmark_semantic_operator_memory.py
```diff
@@ -0,0 +1,577 @@
...
+    if queue.empty():
+        return {
+            "ok": False,
+            "case": case,
+            "error": f"child exited with code {process.exitcode} before reporting",
+            "peak_rss_bytes": 0,
+        }
+    result = queue.get()
+    result["exit_code"] = process.exitcode
+    return result
```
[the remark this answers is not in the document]
bcallender: Fixed in 6ed4d2c: the worker result is now retrieved with non-blocking queue reads after `join()`, treating `Empty`/errors as the failure case instead of relying on `Queue.empty()`.

src/fenic/_backends/local/semantic_operators/reduce.py
```diff
@@ -143,9 +155,11 @@ def _reduce_group(self, group_index: int, group: pl.Series) -> str | None:
 
     def _preprocess_group(self, group: pl.Series) -> Tuple[str, pl.Series]:
         """Preprocess group by adding context and ordering the data column."""
+        group_df = pl.DataFrame(group.to_list())
```
YoungVor: I think group.struct.unnest() is faster because we don't convert to python dict before converting back to a dataframe

YoungVor: also, as a nit, could probably avoid unnest/building a dataframe enitrely if you don't need sorting or group context (and just return group.struct.field(DATA_COLUMN_NAME) )

src/fenic/_backends/local/semantic_operators/reduce.py
```diff
@@ -82,12 +81,25 @@ def __init__(
             self._user_instruction_template = jinja2.Template(self.user_instruction)
 
     def execute(self) -> pl.Series:
-        """Execute parallel reduction across all groups."""
-        with ThreadPoolExecutor(max_workers=min(10, len(self.input))) as executor:
```
YoungVor, line 86: is "no runtime implementation change / Risk: Low.” in the PR description accurate?  We were doing the each reduce group's LLM calls in parallel before, but Claude flagged that Polars will do these serially now.

I don't think its worth changing anything if we don't notice any performance impact, but worth mentioning the behavior change in the PR description

tests/_backends/local/functions/test_semantic_reduce.py
```diff
@@ -1,19 +1,95 @@
...
+
+    requests = batch["requests"]
+    assert len(requests) == 1
+    request = requests[0]
+    assert isinstance(request, FenicCompletionsRequest)
+    assert request.max_completion_tokens == 37
+    assert request.top_logprobs is None
+    assert request.structured_output is None
+    assert request.model_profile is None
+    assert request.temperature in (0.25, None)
```
YoungVor, line 87: nit: could determine whether this should be None based on looking up the model passed to the test

src/fenic/_backends/local/semantic_operators/reduce.py
YoungVor: we already use polar's map_batches for the other semantic operators, now this is no different

YoungVor (approved):
Looks good, pending a few minor nits

src/fenic/_backends/local/semantic_operators/reduce.py
bcallender: Fixed in 38dd1c7: `_preprocess_group` now uses the direct `group.struct.field(DATA_COLUMN_NAME)` fast path when there is no sorting/group context, and `group.struct.unnest()` when a DataFrame is needed. This avoids the Python `to_list()` -> DataFrame round-trip.

src/fenic/_backends/local/semantic_operators/reduce.py
bcallender: Updated in 38dd1c7 / PR body: agreed that the original risk wording was stale after the Polars compatibility fix. The PR now calls out the semantic.reduce runtime compatibility change and notes that cross-group execution is delegated to Polars grouped `map_batches` scheduling rather than the old explicit `ThreadPoolExecutor`. A focused fake-model check showed grouped callbacks still run concurrently under current Polars, so I did not restore a nested executor.

tests/_backends/local/functions/test_semantic_reduce.py
bcallender: Leaving this one as a non-blocking test nit for now. The assertion intentionally allows both `0.25` and `None` because the model/client path can normalize temperature depending on the configured model profile; tightening it would require making the test model-profile-aware rather than improving the tripwire's core coverage.

bcallender:
Follow-up on the semantic.reduce threadpool / Polars concurrency question.

I do **not** think removing the `ThreadPoolExecutor` was strictly required for compatibility with the newer Polars API. The required compatibility change was adapting the grouped aggregation path from the old `agg_list=True` shape to the current:

```python
struct.implode().map_batches(..., returns_scalar=True)
```

With that shape, Polars invokes the Python callback per group. In the normal grouped execution path, `SemanticReduce.execute()` is therefore usually handed one group at a time, so the old internal `ThreadPoolExecutor(max_workers=min(10, len(self.input)))` would not be the mechanism providing cross-group parallelism anymore. Cross-group scheduling is delegated to Polars.

I sanity-checked that with two local smoke probes on this branch using Polars `1.35.2`:

1. **Reduce-like grouped Polars callback**
   - 24 groups × 8 docs/group × ~20k chars/doc
   - callback slept `0.25s`
   - expected serial time: ~`6.0s`
   - observed elapsed: ~`1.62s`
   - Polars thread pool size: `4`
   - observed max callback overlap: `4`

2. **Actual `Reduce(...).execute()` path with a fake sleepy model**
   - same 24-group long-corpus setup
   - fake `get_completions()` slept `0.25s`
   - expected serial time: ~`6.0s`
   - observed elapsed: ~`1.60s`
   - observed max completion overlap: `4`
   - speedup vs serial: ~`3.7x`

I also checked the old threadpool-style `execute()` against the new callback shape, and it was not inherently incompatible. It just no longer owns useful cross-group fanout in the grouped Polars path.

So my conclusion is:

> The PR does not need to remove the threadpool for correctness, but with the current `implode().map_batches(..., returns_scalar=True)` API, semantic.reduce no longer controls cross-group concurrency directly. Polars does. If the API shape were different — e.g. if `Reduce.execute()` received a batch containing many groups — adding threads there could make sense for a large latency-bound workload. In the current shape, that control belongs at the Polars scheduler/thread-pool layer, not inside `Reduce.execute()`.

I filed a follow-up ticket to turn the ad hoc smoke check into a repeatable benchmark/smoke guardrail without making CI depend on brittle wall-clock timing: [TD-3420](https://linear.app/typedef-ai/issue/TD-3420/add-semanticreduce-concurrency-smoke-benchmark-for-polars-grouped-map).

PR #318, opened 2026-06-11
perf: bound semantic similarity join memory

bcallender:
## Summary
- Bounds semantic similarity join vector-search memory by processing left-side embeddings in batches instead of one eager Polars map over the full left side.
- Uses a managed temporary LanceDB directory so sim_join vector-index artifacts are cleaned after execution.
- Avoids materializing synthetic normalized vector columns for derived `semantic.sim_join` expressions in the final result.
- Pins `openai==2.40.0` to avoid the `openai>=2.41.0` request-header regression that surfaced in CI while this PR was being validated.

## Linear issues
- TD-3320

## What changed
- Added `left_batch_size` to the local semantic sim join executor and slices left embeddings through LanceDB search in bounded chunks.
- Replaced persistent UUID-named LanceDB temp dirs with `TemporaryDirectory` under the existing vector-index root.
- Added explicit include/drop controls for normalized join-vector columns so derived expressions do not widen the result.
- Added direct no-network regression coverage for batching, temp-dir cleanup, synthetic vector-column dropping, and existing custom embedding golden output.
- Temporarily pinned OpenAI client below 2.41.0 so unrelated header inflation does not mask the sim_join regression/validation signal.

## Evidence / validation
- `uvx ruff check src/fenic/_backends/local/semantic_operators/sim_join.py src/fenic/_backends/local/physical_plan/join.py tests/_backends/local/dataframe/test_semantic_sim_join.py` → pass
- `OPENAI_API_KEY=*** .venv/bin/python -m pytest -q tests/_backends/local/dataframe/test_semantic_sim_join.py::test_semantic_sim_join_custom_embeddings_golden_output tests/_backends/local/dataframe/test_semantic_sim_join.py::test_semantic_sim_join_can_skip_normalized_vector_columns tests/_backends/local/dataframe/test_semantic_sim_join.py::test_semantic_sim_join_left_batching_preserves_output tests/_backends/local/dataframe/test_semantic_sim_join.py::test_semantic_sim_join_removes_lancedb_temp_dir tests/_backends/local/dataframe/test_semantic_sim_join.py::test_semantic_sim_join_with_incompatible_embeddings` → `5 passed, 5 warnings`
- OpenAI pin validation:
  - `uv run pytest tests/examples/test_examples.py -q` → `11 passed`
  - focused semantic cluster / sim_join CI repro tests with `openai==2.40.0` → `3 passed`
  - GitHub CI for head `bcba005a40a0` is green, including `test 3.11, local, true`, `trunk check`, `Trunk Check`, and `allcheckspassed`.
- Memory smoke harness, same parameters (`sim_join`, rows=64, right_rows=32, dimensions=8, k=2):
  - before: `ok=True`, `result_rows=128`, `peak_rss_bytes=441061376`, `elapsed_ms=13883.341`
  - after: `ok=True`, `result_rows=128`, `peak_rss_bytes=410025984`, `elapsed_ms=4042.508`
- Larger-scale memory harness runs, comparing `origin/main` / `4168f5c` against PR head `bcba005a40a0`:

| matrix | before peak RSS | after peak RSS | delta | result rows | elapsed before → after |
|---|---:|---:|---:|---:|---:|
| `rows=2048, right_rows=1024, embedding_dimensions=64, k=10` | 462,057,472 | 407,810,048 | -54,247,424 (-11.7%) | 20,480 | 8.6s → 18.4s |
| `rows=4096, right_rows=2048, embedding_dimensions=64, k=10` | 489,304,064 | 470,720,512 | -18,583,552 (-3.8%) | 40,960 | 8.2s → 44.1s |
| `rows=8192, right_rows=4096, embedding_dimensions=64, k=10` | 573,063,168 | 529,420,288 | -43,642,880 (-7.6%) | 81,920 | 12.2s → 45.1s |

Command shape used for each before/after pair:

```bash
uv run python tools/benchmark_semantic_operator_memory.py \
  --cases sim_join \
  --rows <rows> \
  --right-rows <right_rows> \
  --embedding-dimensions 64 \
  --k 10 \
  --label TD-3320-<before|after>-sim-join-<size> \
  --json
```

### 2026-06-12 performance recheck / hold recommendation

A clean recheck after moving the PR worktree to the canonical Fenic worktree location did **not** reproduce the original 44–45s after-run outliers, but it also does **not** support landing this PR as-is for memory savings. The current implementation still materializes the complete match set and final eager output; only the left-side LanceDB search loop is chunked. Profiling shows the changed path spends its time in repeated Polars `map_elements` / collect passes over left batches, so batching adds execution overhead without meaningfully bounding end-to-end peak RSS.

Recheck evidence is stored locally under `.hermes/evidence/perf-recheck-2026-06-12/`.

| matrix | before elapsed | after elapsed | runtime ratio | before RSS | after RSS | RSS delta |
|---|---:|---:|---:|---:|---:|---:|
| `2048x1024` | 7.14s | 4.77s | 0.67× | 441.8 MiB | 463.5 MiB | +21.7 MiB (+4.9%) |
| `4096x2048` | 5.41s | 9.12s | 1.69× | 466.1 MiB | 491.0 MiB | +24.9 MiB (+5.3%) |
| `8192x4096` | 9.90s | 17.81s | 1.80× | 547.5 MiB | 537.9 MiB | -9.6 MiB (-1.8%) |

Recommendation: hold/redesign the memory-bound sim_join approach unless we can demonstrate a workload that currently OOMs and is made safe by this bounded path. The temp-dir cleanup and OpenAI pin may still be worth salvaging separately.

## Reviewer notes
- This does not change user-facing nearest-neighbor semantics; batching is only on the left-side search loop.
- Scope boundary from adversarial review: this PR does not make sim_join fully streaming end-to-end. Normalized input columns, the complete `matches_df`, and the final eager output are still materialized; the bounded part is the left-side LanceDB search loop.
- The larger-scale harness shows lower peak RSS on the tested matrices, but also shows the current sequential batching path is slower at larger row counts. Treat this PR as a memory-bound/safety improvement, not a throughput improvement; batch-size/concurrency tuning should be follow-up work if needed.
- The `len(right) > 5000` LanceDB-index path was not separately covered by the new temp-dir cleanup test; the cleanup mechanism uses the same `TemporaryDirectory` lifecycle, but this is a reasonable follow-up if reviewers want indexed-path coverage.
- Full `test_semantic_sim_join.py` still requires real model credentials for network-backed embedding tests, so focused validation used the no-network/direct custom-embedding tests plus the existing memory benchmark harness.
- Local pre-commit was blocked by Trunk failing to install `gitleaks` because its cached Go toolchain directory hit a chmod permission error; the touched files are Python-only, and the explicit validation above was run after the commit contents were staged.

## Release / risk
- Risk is localized to local semantic similarity join execution plus the temporary OpenAI dependency pin.
- Primary regression risks are result ordering/schema changes and derived-expression column restoration; both are covered by focused tests in this PR.
- The OpenAI pin should be revisited once the upstream/client header regression is understood or fixed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ dependencies = [
   "polars>=1.34.0,<1.36.0",
   "tiktoken>=0.11.0",
   "lancedb>=0.22.0,<0.25.0",
-  "openai>=1.101.0",
+  "openai==2.40.0",
   "sqlglot>=29.0.1,<31.0.0",
   "pandas>=2.2.2",
   "cloudpickle>=3.1.1",
--- a/src/fenic/_backends/local/physical_plan/join.py
+++ b/src/fenic/_backends/local/physical_plan/join.py
@@ -259,20 +259,30 @@ class SemanticSimilarityJoinExec(PhysicalPlan):
         )
 
         # TODO(rohitrastogi): Avoid regenerating embeddings if semantic index already exists
-        result = SemanticSimJoin(left_df, right_df, self.k, self.similarity_metric).execute()
+        result = SemanticSimJoin(
+            left_df,
+            right_df,
+            self.k,
+            self.similarity_metric,
+            include_left_on=maybe_left_name is not None,
+            include_right_on=maybe_right_name is not None,
+        ).execute()
 
         if self.similarity_score_column:
             result = result.rename({DISTANCE_COL_NAME: self.similarity_score_column})
         else:
             result = result.drop(DISTANCE_COL_NAME)
 
-        # Restore original column names or drop temporary columns
-        result = restore_column_after_join(
-            result, maybe_left_name, LEFT_ON_COL_NAME
-        )
-        result = restore_column_after_join(
-            result, maybe_right_name, RIGHT_ON_COL_NAME
-        )
+        # Restore original column names. Derived temporary join columns are not
+        # materialized by SemanticSimJoin, so there is nothing to drop for them.
+        if maybe_left_name is not None:
+            result = restore_column_after_join(
+                result, maybe_left_name, LEFT_ON_COL_NAME
+            )
+        if maybe_right_name is not None:
+            result = restore_column_after_join(
+                result, maybe_right_name, RIGHT_ON_COL_NAME
+            )
 
         return result
 
--- a/src/fenic/_backends/local/semantic_operators/sim_join.py
+++ b/src/fenic/_backends/local/semantic_operators/sim_join.py
@@ -1,4 +1,5 @@
-import uuid
+import os
+import tempfile
 
 import lancedb
 import polars as pl
@@ -21,6 +22,8 @@ RIGHT_ON_COL_NAME = "__right_on__"
 LEFT_ID_COL_NAME = "__left_id__"
 RIGHT_ID_COL_NAME = "__right_id__"
 MATCH_RESULT_COL_NAME = "__match_result__"
+DEFAULT_LEFT_BATCH_SIZE = 1024
+
 
 class SimJoin:
     def __init__(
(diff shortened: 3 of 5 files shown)


bcallender:
Closing this larger memory-bound sim_join PR for now. Current scale testing shows the runtime regression is not worth the relatively small memory improvement in this form. The small temp index cleanup has been split out separately in #320, and we can continue the performance investigation/optimization from a cleaner branch.

PR #322, opened 2026-06-19
feat: agent-enablement tooling — fenic-mechanics skill, `fenic check`, cross-agent install

cpard:
## What & why

fenic post-dates most coding agents' training, so they don't reliably know its API. This PR adds tooling that makes AI coding agents write **correct** fenic — built from an empirical study of where a frontier model actually breaks on fenic, then measured.

## What's included

**1. `fenic-mechanics` skill** (`.claude/skills/fenic-mechanics/`) — a hybrid Agent Skill:
- *Generated* API reference (`reference/*.md`) introspected from the installed fenic by `scripts/generate_reference.py` — version-stamped, can't drift.
- *Hand-authored* judgment (`SKILL.md`, `gotchas.md`): the import/namespace law, semantic-operator calling conventions, and the handful of **silent** traps a type-checker can't catch (`json.jq` returns an array; `{{ }}` vs `{ }` templates; `datediff` arg order; `to_timestamp` takes Spark/Java patterns).
- `update-fenic-skill` maintenance skill: regenerate the reference + re-review judgment on a fenic version bump.

**2. `fenic check`** (`src/fenic/scripts/fenic_check.py`, new `fenic` console script) — validate a fenic script **without executing it**: static symbol/namespace lint + dry-run logical-plan construction. Returns the result schema or the precise error as JSON. No tokens, no data; an API key only if the script configures semantic models.

**3. `fenic skill install`** — copy the skill into the skill directories your agents read. Detects installed agents (Claude Code, Codex, Cursor, Gemini, Copilot), asks which to install for, and whether globally (`~/.claude/skills` + `~/.agents/skills`) or for the current project. Ships in the wheel via maturin `include`.

**4. Cross-agent + plugin extras** — `AGENTS.md` + an `@AGENTS.md` bridge in `CLAUDE.md`, `.cursor/rules/fenic.mdc`, and a `.claude-plugin/` (validated with `claude plugin validate --strict`) for Claude Code plugin install.

**5. README** — a "Writing fenic with an AI coding agent" section.

## How it was validated

An A/B eval re-ran the tasks a naive agent failed on, this time with the skill as the only knowledge source: **10/12 blocked→pass** (the 2 misses were a skill bug, since fixed → effectively 12/12). The eval also caught two real bugs that are fixed here: a `fenic check` `__name__` false-pass, and an inverted `to_timestamp` format rule in the skill.

## Tests

`tests/test_fenic_check.py` + `tests/test_fenic_assets.py` — **14 hermetic tests** (lint diagnostics, dry-run validation incl. a regression guard for the `__name__` bug, agent detection, skill-dir mapping + dedup, install). All pass locally; ruff clean. CI will run the full suite.

## Notes for reviewers

- **Purely additive — no existing behavior changes.** New subcommands under a new `fenic` console entry; `fenic-serve` untouched.
- **The `fenic` / `fenic skill install` commands go live with the next PyPI release** (they ride this branch); the README documents them accordingly.
- **Build step:** `just bundle-agent-assets` mirrors the committed skill into `src/fenic/_agent_assets/` (a gitignored build artifact) before `maturin build`. Verified with a real wheel build that it packs under `fenic/_agent_assets/`, and a clean-venv `pip install <wheel>` runs `fenic skill install` / `fenic check` end-to-end.
- The skill installs **user-global by default and never writes into your project's own files** (no `AGENTS.md`/`CLAUDE.md` ownership).


.claude/skills/update-fenic-skill/SKILL.md
```diff
@@ -0,0 +1,75 @@
+---
+name: update-fenic-skill
```
bcallender, line 2: Can you add a reference to this skill in the fenic-feature-developer subagent we configure in .claude? Just to ensure that agents working in the repo respect that these skills need to be updated if the feature changes/adds api surface

bcallender (changes requested):
I’d recommend limiting `fenic check` to static linting in this PR and deferring exec-based validation until we add first-class dry-run support in fenic.

Right now `fenic check` monkey-patches selected DataFrame actions, but it still executes arbitrary user Python in-process. That means imports, top-level code, session/catalog operations, readers, and any non-fenic side effects can still run. There are also fenic-specific gaps: `save_as_view` is not stubbed and can mutate the catalog, while writer methods like `df.write.csv(...)` are stubbed before their normal validation/sink-plan construction runs, so invalid writer calls can incorrectly pass.

The safer path seems pretty achievable: add a real dry-run mode at the execution/session boundary, letting public APIs still validate arguments and build logical/sink plans, while `collect/show/count/save_to_file/save_as_table/save_as_view` return schema/plan metadata without materializing or mutating. Physical-plan dry run can be the implementation detail for query actions, but catalog writes need explicit handling above that layer.

So for this PR I’d either ship `fenic check` as lint-only, or mark/remove exec validation until the real dry-run path exists. That keeps the tool’s safety contract honest and avoids normalizing a check that can still execute or mutate state.

cpard:
Thanks @bcallender — agreed on both, and that's a fair call on the safety contract. Addressed in c5c5d92.

**`fenic check` is now lint-only (no execution).** I removed the exec-based validate pass entirely. Since nothing runs anymore, the specific issues you flagged are all moot:
- no more arbitrary user code (imports / top-level / session + catalog ops / readers) executing in-process;
- `save_as_view` catalog mutation — gone (nothing executes);
- writer methods stubbed before their validation — gone (no stubbing at all).

It's now a pure static AST lint: it resolves `fc.*` symbols against the installed fenic and flags the dominant mistakes (`fenic.functions`, `fc.array` vs `fc.arr`, `fc.explode`, internal imports, unknown symbols). I also added a test that asserts a script with a top-level side effect is **not** executed, to lock the contract in.

The richer validation you described — a real dry-run at the execute/session boundary (public APIs validate args + build logical/sink plans; `collect/show/count/save_*` return schema/plan metadata without materializing or mutating, with explicit handling for catalog writes) — is the right design, but it's a fenic feature, so I've deferred it out of this PR. Happy to open a tracking issue for it if you'd like.

**Subagent:** the `fenic-feature-developer` agent now has an explicit note that any change to public API surface must update the `fenic-mechanics` skill in the same PR — via the `update-fenic-skill` skill, which regenerates the reference from the installed fenic and flags which hand-authored judgment to re-review.

Docs (`SKILL.md` / `gotchas.md` / `AGENTS.md` / `.cursor` / `README`) updated to describe `fenic check` as a static lint rather than "builds the plan." Re-requesting review when you have a moment.

bcallender (approved):
Approved. The updated direction is much safer now that `fenic check` is static lint only. I’m leaving the existing comments open for Kostas or his agent to address, especially the remaining import-lint cleanup, but I don’t consider them blocking for approval.

src/fenic/scripts/fenic_check.py
```diff
@@ -0,0 +1,138 @@
...
+    import fenic as fc
+
+    try:
+        tree = ast.parse(src, filename=path)
+    except SyntaxError as e:
+        return [_finding("error", "SyntaxError", str(e), line=e.lineno)]
+
+    findings: list[dict] = []
+    aliases: set[str] = set()
+    for node in ast.walk(tree):
```
bcallender, line 66: `fenic check` still misses one of the advertised import traps: `import fenic.functions as F` currently returns `ok: true`, even though Python raises `ModuleNotFoundError: No module named 'fenic.functions'`. Since the tool is now intentionally static lint, this import form should probably be a lint error alongside `from fenic import functions ...`.

src/fenic/scripts/fenic_cli.py
```diff
@@ -0,0 +1,109 @@
+"""`fenic` umbrella CLI.
+
+Subcommands:
+  check <file>     Validate a fenic script without executing it (lint + dry-run plan).
```
bcallender: Stale docstring: this still describes `check` as "lint + dry-run plan", but the implementation and CLI help now correctly make it static lint only. Worth updating so future readers do not reintroduce the old safety contract.

.claude/skills/update-fenic-skill/SKILL.md
cpard: Added in c5c5d92 — the `fenic-feature-developer` subagent now instructs updating the `fenic-mechanics` skill via the `update-fenic-skill` skill whenever public API surface changes.

src/fenic/scripts/fenic_check.py
cpard: Fixed in 0eb9776 — `import fenic.functions` (with or without `as`) now lints as a BadImport, same as `from fenic import functions`. Added a test.

src/fenic/scripts/fenic_cli.py
cpard: Fixed in 0eb9776 — the module docstring no longer says 'lint + dry-run plan'; it now reads 'Statically lint a fenic script's symbol/namespace usage (no execution)'.

PR #335, opened 2026-07-24
docs: improve crawler and LLM discoverability

cpard:
## Summary

- fix sitemap generation and canonical URLs by configuring the docs site URL and Mike's `/latest/` canonical path
- add unique page descriptions, Open Graph metadata, JSON-LD, crawler policy, and noindex handling for internal planning pages
- generate `llms.txt`, `llms-full.txt`, per-page Markdown alternatives, agent instructions, and root discovery assets
- publish root assets during tagged docs releases and canonicalize legacy version pages to the latest equivalent
- validate sitemap entries, canonicals, descriptions, structured data, Markdown alternatives, LLM assets, and robots configuration during docs builds

## Testing

- `GITHUB_ACTIONS=true DISABLE_MKDOCS_2_WARNING=true just docs-check`
- `uvx ruff check scripts/docs_metadata.py scripts/docs_llms.py scripts/validate_docs_site.py scripts/backfill_docs_version_metadata.py`
- `uvx ruff format --check scripts/docs_metadata.py scripts/docs_llms.py scripts/validate_docs_site.py scripts/backfill_docs_version_metadata.py`
- `python3 -m py_compile scripts/docs_metadata.py scripts/docs_llms.py scripts/validate_docs_site.py scripts/backfill_docs_version_metadata.py`
- `bash -n scripts/publish_docs_root_assets.sh`
- `git diff --check`

The generated docs validation covers 63 sitemap URLs. The legacy metadata backfill and root-asset publisher were also tested against temporary Git worktrees/remotes for idempotency without modifying the live `gh-pages` branch.

### DIFF
--- a/.github/workflows/docs_check.yaml
+++ b/.github/workflows/docs_check.yaml
@@ -1,0 +1,40 @@
+---
+name: docs check
+on:
+  pull_request:
+    types: [opened, reopened, synchronize]
+    paths:
+      - .github/workflows/docs_check.yaml
+      - docs/**
+      - mkdocs.yml
+      - pyproject.toml
+      - scripts/docs_llms.py
+      - scripts/docs_metadata.py
+      - scripts/validate_docs_site.py
+      - src/fenic/**
+      - uv.lock
+  workflow_dispatch:
+concurrency:
+  # yamllint disable-line rule:line-length
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+permissions:
+  contents: read
+jobs:
+  docs:
+    name: docs build and discovery assets
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+      - name: setup just
+        # yamllint disable-line rule:line-length
+        uses: extractions/setup-just@e33e0265a09d6d736e2ee1e0eb685ef1de4669ff # v3
+      - name: setup uv
+        uses: astral-sh/setup-uv@f0ec1fc3b38f5e7cd731bb6ce540c5af426746bb # v6
+        with:
+          python-version: "3.11"
+          activate-environment: true
+          enable-cache: true
+      - name: build and validate docs
+        run: just docs-check
--- a/.github/workflows/release_docs.yaml
+++ b/.github/workflows/release_docs.yaml
@@ -29,6 +29,8 @@ jobs:
       - name: deploy docs
         shell: bash
         env:
+          GIT_AUTHOR_NAME: typedef-gha
+          GIT_AUTHOR_EMAIL: noreply@typedef.ai
           GIT_COMMITTER_NAME: typedef-gha
           GIT_COMMITTER_EMAIL: noreply@typedef.ai
         run: |
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -38,6 +38,15 @@ PySpark/pandas intuition. _(Developing fenic itself? See `CLAUDE.md`.)_
 - `fc.dt.to_timestamp(col, fmt)` takes **Spark/Java** patterns (`yyyy-MM-dd HH:mm:ss`),
   **not** Python `%`-tokens.
 
+## Documentation maintenance
+
+Authored public documentation pages must be listed under `extra.llms.sections`
+in `mkdocs.yml`; generated API-reference pages are included automatically.
+When adding, renaming, or removing a docs page, update that list or explicitly
+mark the page `llms: false` and `noindex` if it is internal. Run
+`just docs-check` to validate the sitemap, metadata, Markdown alternatives, and
+LLM-facing assets before committing.
+
 ## Full detail
 
 `.claude/skills/fenic-mechanics/` — `SKILL.md` (rules), `gotchas.md` (the
--- a/docs/agents.md
+++ b/docs/agents.md
@@ -1,0 +1,86 @@
+---
+title: Instructions for AI agents
+description: Authoritative guidance for AI assistants using, describing, or generating code with fenic.
+---
+
+# Instructions for AI agents
+
+This page tells AI assistants and coding agents how to use and describe fenic
+accurately. It is maintained with the fenic documentation.
+
+## How to consume these docs
+
+- Start with <https://docs.fenic.ai/llms.txt> for a concise, structured index.
+- Use <https://docs.fenic.ai/llms-full.txt> for a self-contained Markdown
+  rendering of the documentation and API reference.
+- Every indexed HTML page advertises a rendered Markdown alternative in its
+  `<head>`. The Markdown URLs are also linked from `llms.txt`.
+- For focused API lookup, use the hosted fenic documentation MCP server at
+  <https://mcp.fenic.ai>.
+- Prefer URLs under <https://docs.fenic.ai/latest/> when citing current
+  behavior. Versioned URLs describe historical releases.
+
+## What fenic is
+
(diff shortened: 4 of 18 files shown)


bcallender:
Read through this — the SEO plumbing and the agent-facing assets are both genuinely valuable, and `docs/agents.md` in particular reads like a real product artifact rather than boilerplate (the anti-hallucination guardrails are the useful part). Two suggestions, both about keeping it correct over time rather than about what it does today.

## 1. The `llms.txt` curation list is a maintenance trap — could it ride the mechanics-reference tooling instead?

`extra.llms.sections` in `mkdocs.yml` is a hand-maintained allowlist of what lands in `llms.txt` / `llms-full.txt` / per-page `.md`. Anyone adding a guide or example has to remember to add it there, and if they forget the page is silently absent from the LLM-facing assets — no build failure, no warning. That is the kind of drift that is invisible until someone notices the agent has stale knowledge.

This repo already solved exactly this problem once, for the `fenic-mechanics` skill reference: `generate_reference.py` introspects the installed fenic, and `.github/workflows/fenic_mechanics_reference.yaml` runs it in **CHECK** mode on every PR (fails if the committed reference drifted) and **FIX** mode on the release-please PR (regenerates and pushes so the version stamp rides the release commit).

Could `llms.txt` be generated from — or by — that same machinery, so the existing drift gate maintains it for free? Even partially: if the API-reference portion of `llms.txt` came from the mechanics reference rather than a globbed allowlist, the largest and most drift-prone section would become self-maintaining, and the manual list would shrink to genuinely curated prose pages.

If a full merge is not practical, a smaller version of the same idea: add an `llms.txt`-drift check to CI so a new page under a curated section fails the build until it is either listed or explicitly excluded — turning a silent omission into a loud one.

## 2. Add a pointer in the skills so agents know to update the LLM assets

Related: since agents (including Claude Code sessions working in this repo) are the audience for `agents.md` and `llms.txt`, it would help to state the obligation where an agent will actually see it — a line in the `fenic-mechanics` skill (and/or `AGENTS.md`) along the lines of *"when you add or rename a docs page, update the `extra.llms.sections` list / regenerate the LLM assets."*

Right now an agent editing docs has no signal that a second artifact needs updating. That is the same class of problem the `MCP exposure is not desktop-agent exposure` note solves in the data-intelligence repo — the tooling silently succeeds while the thing you actually care about is stale.

## 3. Heads-up: #336 removes the two directories you are noindexing

I opened #336, which moves `docs/td-flow/` and `docs/plans/` into `specs/` (alongside the existing internal design docs) so internal planning artifacts are not built into the public site at all. It overlaps the `docs/plans/.meta.yml` and `docs/td-flow/.meta.yml` you add here — whichever lands second needs a trivial rebase, and if #336 lands first those two files can be dropped. Flagging so it is not a surprise; happy to sequence whichever way you prefer.

## Minor

CI is currently red on `trunk check`: markdownlint **MD025** (multiple top-level headings) at `docs/agents.md:6`, plus two Bandit findings (**B405/B314**) on `xml.etree.ElementTree` in `scripts/validate_docs_site.py`. The Bandit ones look like false positives — that script parses its own build output, not untrusted input — but they are failing the gate as configured.

Two things worth a sanity check before this ships, since `agents.md` will be treated as authoritative by agents: that `mcp.fenic.ai` actually resolves and serves something, and that the AI-crawler allowlist in `robots.txt` (`ai-train=yes` for OpenAI/Anthropic/Perplexity/Meta/Bytespider et al.) is a deliberate policy call rather than an inherited default.

bcallender:
Correction to my earlier comment: the CI note in the **Minor** section was already stale when I posted it — Trunk Check and `allcheckspassed` are both green on this PR now. Disregard that paragraph; the MD025 and Bandit findings are resolved.

The two substantive suggestions (generating `llms.txt` off the `fenic-mechanics` reference tooling rather than a hand-maintained `extra.llms.sections` list, and a skills pointer so agents know the LLM assets need updating alongside docs) still stand, as does the #336 overlap heads-up.

cpard:
@bcallender Thanks for the detailed pass — I addressed the maintenance concerns in `b009cf1`.

1. The generated API-reference section was already picked up automatically via `reference/**/*.md`; the authored prose pages remain intentionally curated. I added a dedicated PR workflow that runs `just docs-check` whenever the docs, generators, API source, or relevant dependencies change. The validator requires every sitemap page to have a generated Markdown alternative, so adding a public page without listing it now fails the PR check instead of silently omitting it.
2. I added a documentation-maintenance section to `AGENTS.md` telling agents to update `extra.llms.sections` (or explicitly exclude internal pages) and run `just docs-check`.
3. I made the excluded-page validation dynamic rather than hardcoding the `plans/` and `td-flow/` paths, and tested it with both directories absent. That avoids a validator failure when #336 lands; the two `.meta.yml` files and corresponding `not_in_nav` entries can be dropped as part of whichever PR lands second.

I also verified `mcp.fenic.ai` with an MCP initialization request; it returned HTTP 200 and a valid `fenic-docs` server response.

And yes, `ai-train=yes` is intentional.

bcallender:
This all looks right, and thank you for the correction on point 1 — I overstated the maintenance trap. I saw `extra.llms.sections` and read it as *the* curation surface without checking that `reference/**/*.md` was already globbing the generated API reference in automatically. The largest and most drift-prone section was never manual, so the exposure was only ever the authored prose pages, which you are curating deliberately.

And the way you closed it is better than what I proposed. Wiring `llms.txt` into the mechanics-reference generator would have coupled two things that do not need to be coupled; making the validator require a Markdown alternative for every sitemap page gets the same guarantee — a forgotten page fails the PR instead of silently vanishing — with no coupling at all. The dedicated `docs_check.yaml` triggering on docs/generators/API-source/deps is the right trigger set.

The `AGENTS.md` maintenance section is also the better placement over a skill, since it is always loaded.

Particularly appreciate you making the excluded-page validation dynamic rather than hardcoding `plans/` and `td-flow/`, and testing it with both directories absent — that unblocks #336 from your side before I even asked. Agreed on the cleanup: whichever lands second drops the two `.meta.yml` files and their `not_in_nav` entries. #336 is green and small (a file move plus a `specs/README.md`), so I am happy to land it first and clean up in your rebase, or hold and let you land first and I will do the cleanup in mine — your call, no preference.

Thanks also for actually exercising `mcp.fenic.ai` with an MCP initialization request rather than just confirming DNS, and for the explicit `ai-train=yes` confirmation. Both were the answers I was hoping for rather than assuming.

bcallender:
#336 is merged (`5d9756651`), so `docs/plans/` and `docs/td-flow/` are gone from `main`. This PR looks good to me otherwise — the cleanup below is the only change I'd ask for before it lands.

**Needs removing** — these two now *recreate* the directories they were meant to suppress, since each would be the only file left in an otherwise-deleted tree:

- `docs/plans/.meta.yml`
- `docs/td-flow/.meta.yml`

**Stale, worth dropping** — `mkdocs.yml:67-70`, the `not_in_nav` entries for `/plans/` and `/td-flow/`. Dead config now that neither path exists.

**Your call** — `scripts/validate_docs_site.py:99`:

```python
if any("/plans/" in url or "/td-flow/" in url for url in urls):
```

Now that the directories are gone this never fires, so it is dead in practice. But it is also a cheap regression guard: if someone later recreates `docs/plans/` or `docs/td-flow/`, this fails the build rather than silently publishing internal material. I would keep it for that reason, but I do not feel strongly — and if you keep it, a comment noting it is a guard against reintroduction rather than a live exclusion would save the next reader the same confusion I just had.

Worth noting the `specs/README.md` that landed with #336 states the convention explicitly — everything under `docs/` is public, internal material goes in `specs/` — so the structural rule is now written down rather than living in per-directory exclusions.

bcallender (approved):

PR #346, opened 2026-07-27
ci(mcp): deploy documentation service on release

bcallender:
## Summary

Each Fenic release can now update the hosted documentation MCP service automatically after its exact wheel is available on PyPI. The protected production job publishes and verifies versioned Hugging Face artifacts, prepares an exact-version Modal image, deploys it, and exercises the live MCP contract before reporting success.

Reruns are safe: complete datasets are reused and verified, while partial exports are regenerated. Generated data records the Fenic version and source commit, and pull requests affecting the service run the native contract and Fenic pipeline checks without production credentials.

## Deployment controls

- Runs automatically only after the tagged PyPI publication job succeeds—not on ordinary merges to `main`.
- Supports manual reruns by selecting the release tag in the workflow UI.
- Uses the protected GitHub `production` environment for deployment credentials and approval rules.
- Keeps the Modal-side `llm_api_keys` secret separate from GitHub Actions credentials.

## Validation

- `trunk check --fix`
- `uv run pytest` — 7 passed
- `uv run fenic check src/fenic_mcp/setup/populate_tables.py`
- `uv run fenic check src/fenic_mcp/server/native.py`

This PR is stacked on the MCP service migration PR.


.github/workflows/check_docs_mcp.yaml
```diff
@@ -0,0 +1,45 @@
+---
```
scottames, line 1: Nit: `docs_mcp_check.yaml` should be `check_mcp_docs` (or deploy renamed to the checks file pattern.

bcallender: Renamed the workflow to `check_mcp_docs.yaml` and updated its path filter. Thanks!

scottames (approved):

PR #357, opened 2026-08-05
feat: support Python 3.13 and 3.14

bcallender:
## Summary

Unblocks the monorepo Python 3.13/3.14 move by making fenic's package metadata, dependency floors, CI, and platform wheels support both interpreters. Publishing the next ordinary fenic release remains the final step before TD-4631 can consume the new wheel support.

The dependency audit uses simple global floor bumps for pandas, tiktoken, zstandard, scikit-learn, and gRPC. It retains interpreter markers only for NumPy and DuckDB, whose Python 3.14-compatible wheel floors differ from earlier interpreters.

PR and manual runs execute synced local tests at the Python 3.10 and 3.14 compatibility boundaries, avoiding the cost of running the live-LLM suite on every supported minor interpreter for each PR. Daily scheduled jobs cover Python 3.10 through 3.14 across synced, minimum, and maximum dependency sets; those jobs preserve their selected resolution and fail when synchronization fails.

The release workflow explicitly lists Python 3.10 through 3.14 for Linux, macOS, and Windows builds. fenic remains a native platform package: the resulting artifacts are `cp39-abi3` wheels containing `fenic/_polars_plugins.abi3.so`, rather than separate minor-version wheels or `py3-none-any` wheels.

Fixes TD-4632.
Related to TD-4631.

## Validation

- Python 3.13 and 3.14 all-extras environments resolved and installed successfully.
- Lowest-direct all-extras installs imported the audited compiled dependencies successfully on Python 3.10, 3.13, and 3.14.
- Focused optional-extra and backend suites passed on both interpreters: 34 tests each.
- Full Python 3.14 non-cloud suite passed: 1,635 passed, 3 skipped, and 28 deselected.
- `just docs-check`, `uv lock --check`, workflow YAML parsing, and matrix JSON validation passed.
- A release-mode macOS arm64 `cp39-abi3` wheel built successfully with the native extension.
- Actionlint reports only four unchanged shellcheck notices already present on `main`.

## Post-Deploy Monitoring & Validation

- Owner/window: release owner from tag creation through 24 hours after PyPI publication.
- Watch the Linux, macOS, Windows, and PyPI publish jobs in the release workflow.
- Healthy signal: each platform publishes a `cp39-abi3` wheel, and clean Python 3.13/3.14 environments can install and import fenic with the supported extras.
- Downstream signal: TD-4631 can resolve the published fenic version on both target interpreters.
- Failure signal: a missing platform artifact, source-build fallback, dependency-resolution failure, or import error on either interpreter.
- Mitigation: stop or yank the affected release, restore the previous Python cap or tighten the failing dependency marker, then publish a corrected release.

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)


.github/workflows/test.yaml
```diff
@@ -34,10 +34,10 @@ jobs:
         run: |
           if [ "${{ github.event_name }}" = "schedule" ]; then
             # Run all combinations on scheduled runs
-            matrix='{"dependency": ["true", "min", "max"], "test": ["local", "cloud"], "python-version": ["3.10", "3.11", "3.12"]}'
+            matrix='{"dependency": ["true", "min", "max"], "test": ["local", "cloud"], "python-version": ["3.10", "3.11", "3.12", "3.13", "3.14"]}'
           else
-            # Run only true/local/3.11 on PR/manual runs
-            matrix='{"include": [{"dependency": "true", "test": "local", "python-version": "3.11"}]}'
+            # Run synced local tests on the floor, default, and newly supported interpreters.
+            matrix='{"include": [{"dependency": "true", "test": "local", "python-version": "3.10"}, {"dependency": "true", "test": "local", "python-version": "3.11"}, {"dependency": "true", "test": "local", "python-version": "3.13"}, {"dependency": "true", "test": "local", "python-version": "3.14"}]}'
```
YoungVor: Do we want to build 3.10, 3.11, 3.13, AND 3.13 on every manual run?  (And why not 3.12?)

YoungVor:
Probably worth updating the interpreter list in build_and_release.yaml as well (even though the published wheel is the same either way).  I assume we just don't want to enforce new requirements on the build nodes

.github/workflows/test.yaml
bcallender: > Do we want to build 3.10, 3.11, 3.13, AND 3.13 on every manual run? (And why not 3.12?)

Agreed—the live-LLM suite is not worth multiplying across every supported minor interpreter for each PR. Adjusted in 9a45f31: PR and manual runs test Python 3.10 and 3.14 because they are the compatibility boundaries and exercise both dependency-marker paths. The daily scheduled matrix continues to cover Python 3.10 through 3.14 across synced, minimum, and maximum dependency sets.

bcallender:
> Probably worth updating the interpreter list in build_and_release.yaml as well (even though the published wheel is the same either way).

Addressed in aeb9591: the Linux, macOS, and Windows wheel-build interpreter lists now all cover Python 3.10 through 3.14. The published native wheel remains ABI3, while the workflow metadata now matches the advertised support range.

YoungVor (approved):

PR #361, opened 2026-08-12
perf(local): avoid redundant DataFrame work and bound sim-join memory

bcallender:
## What problem does this solve?

Local execution was allocating replacement Polars DataFrames for operations
whose inputs already satisfied the requested result, and `semantic.sim_join`
expanded its entire left-side search through one transient transform.

This change removes only those two internal costs. It does not add a public API
or change result shapes.

### Bound transient work in semantic similarity joins

`semantic.sim_join` now searches the left input in fixed 1,024-row slices and
concatenates the existing narrow match rows. The final N×k result contract is
unchanged, while the transient search/map/explode/unnest work is bounded.

The regression test forces one-row slices, observes both search calls, and
asserts the complete ordered match set. A second test rejects a non-positive
slice size.

### Skip provable local DataFrame no-ops

- Ingestion returns the original DataFrame only when every target dtype already
  matches; any required coercion follows the existing select-and-cast path.
- Projection returns its child only for raw input columns in the same order.
  Aliases, casts, selectors, computed expressions, and reordered columns still
  use `select`.
- Union skips right-side alignment only when the right column list exactly
  equals the left list; reordered columns still align with `select`.

Focused tests cover row and empty-frame identity behavior, datetime/Struct/
EmbeddingType normalization, required coercions, union reordering, and every
projection guard boundary.

### Document the no-copy boundary

The public `Session.create_dataframe`, `DataFrame.collect`, and
`DataFrame.to_polars` docstrings now state that a schema-free, no-op local plan
may alias an already canonical Polars input rather than copying it. The schema
guide distinguishes this pass-through from explicit-schema ordering/casts and
coercion-required inputs. The timestamp guide retains its UTC value guarantees
while clarifying that already canonical microsecond-UTC Polars columns can pass
through and inputs requiring timezone or precision normalization still
materialize a new frame.

## Verification

```bash
uv run --no-sync pytest tests/_backends/local/physical_plan/test_eager_noop_fast_paths.py
uv run --no-sync pytest tests/_backends/local/dataframe/test_semantic_sim_join.py -k 'batches_left_searches_without_changing_matches or rejects_non_positive_left_batch_size'
uvx ruff check src/fenic/_backends/local/semantic_operators/sim_join.py src/fenic/_backends/local/physical_plan/transform.py src/fenic/_backends/local/physical_plan/utils.py tests/_backends/local/dataframe/test_semantic_sim_join.py tests/_backends/local/physical_plan/test_eager_noop_fast_paths.py
just docs-check
```

Observed locally: 8 eager-no-op tests passed, 2 direct sim-join tests passed,
Ruff passed, `just docs-check` validated 63 sitemap URLs and LLM assets, and
`git diff --check origin/main..HEAD` passed.

## Walkthrough

The accompanying walkthrough comment maps the two changes and their focused
verification coverage.

### DIFF
--- a/docs/topics/create_dataframe_schema.md
+++ b/docs/topics/create_dataframe_schema.md
@@ -166,11 +166,18 @@ provider — the `embedding_model` on the type is metadata.)
 
 ## Relationship to Schema-Free Ingestion
 
-The no-schema path is unchanged and fully behavior-compatible:
+The no-schema path retains its existing schema semantics:
 
 - Empty lists are still rejected.
 - Types are still inferred from the data.
 - Fixed-size arrays are still normalized to `pl.List`.
+
+Its object-identity behavior differs for schema-free Polars inputs: when the
+input already has Fenic-compatible physical dtypes, local execution can retain
+and return the same Polars frame instead of copying it. If ingestion must
+normalize an array, timestamp, or nested dtype, it materializes a new frame.
+Supplying an explicit schema also applies its authoritative ordering and casts,
+so it does not use this schema-free pass-through path.
 
 Only supplying a `schema` opts you into authoritative names, ordering, coercion,
 and logical-type/embedding preservation. For related ingestion-normalization
--- a/docs/topics/timestamps_and_dates.md
+++ b/docs/topics/timestamps_and_dates.md
@@ -14,9 +14,12 @@ Both types and related functionality follow similar patterns to PySpark, except 
 ## Key Features
 
 - **Microsecond precision**: Timestamps support up to 1/1,000,000 second precision
-- **UTC-first design**: All timestamps are normalized to UTC during ingestion
+- **UTC-first design**: Timestamps that need conversion are normalized to UTC
+  during ingestion; already canonical UTC Polars columns can pass through
+  unchanged
 - **Automatic timezone conversion**: Timezone-aware inputs are converted to UTC automatically
-- **Consistent behavior**: Same behavior across all data sources (Parquet, CSV, in-memory DataFrames, tables)
+- **Consistent values**: All data sources expose the same UTC timestamp
+  semantics, although their materialization and copy behavior can differ
 - **Rich date/time functions**: Comprehensive set of functions for temporal operations
 - **Timezone conversion utilities**: Functions to work with local timezones while maintaining UTC storage
 
@@ -192,7 +195,10 @@ The same behavior applies to:
 
 - **Parquet files**: Timestamps stored with timezone metadata, automatically converted to UTC on read
 - **Fenic tables** (DuckDB): DuckDB session timezone is always UTC, ensuring consistency
-- **In-memory DataFrames**: Polars DataFrames with timezone-aware or naive timestamps are normalized to UTC
+- **In-memory DataFrames**: Polars DataFrames with timezone-aware or naive
+  timestamps are normalized to UTC when needed. An already canonical
+  `Datetime("us", "UTC")` column can pass through without copying the input
+  frame.
 
 ### SQL Queries
 
@@ -241,6 +247,12 @@ This behavior ensures:
 ### Timezone Conversion During Ingestion
 
 Fenic automatically handles different timezone inputs:
+
+For a schema-free in-memory Polars DataFrame that already uses Fenic's canonical
+microsecond UTC dtype, ingestion does not perform a conversion and local
+collection may alias the original frame. Naive timestamps, non-UTC timestamps,
+or non-microsecond precision still require normalization and produce a new
+frame.
 
 ```python
 import datetime
--- a/src/fenic/_backends/local/physical_plan/transform.py
+++ b/src/fenic/_backends/local/physical_plan/transform.py
@@ -23,6 +23,27 @@ from fenic._backends.local.physical_plan.base import (
 
 logger = logging.getLogger(__name__)
 
+
+def _is_identity_projection(
+    projections: List[pl.Expr], columns: List[str]
+) -> bool:
+    """Return whether projections preserve every input column in its current order."""
+    return len(projections) == len(columns) and all(
+        projection.meta.is_column()
+        and projection.meta.output_name() == column
+        for projection, column in zip(projections, columns, strict=True)
+    )
+
+
+def _align_union_right_dataframe(
+    left_df: pl.DataFrame, right_df: pl.DataFrame
+) -> pl.DataFrame:
+    """Align only a differently ordered right side to the left union schema."""
+    if right_df.columns == left_df.columns:
+        return right_df
+    return right_df.select(left_df.columns)
+
+
 @dataclass(frozen=True)
 class SeriesLiteralCheck:
     column_name: str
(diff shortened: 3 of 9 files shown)


bcallender:
## PR #361 walkthrough — avoid redundant DataFrame work and bound sim-join memory

<strong>Two independently safe local-execution improvements land together:</strong> bounded temporary work while semantic similarity joins search the left input, and skipped Polars DataFrame materialization only when a transformation is provably a no-op. Every other execution-engine layer remains deliberately outside this PR.

**+356 / −54** diff (9 files) · **3** commits · **7dbef64** head · **10 focused tests** local pytest · **63 docs URLs** docs-check validated · **No API behavior change** result contracts retained

<details><summary>❓ <b>Problem: local execution performs avoidable eager work</b></summary>


Similarity join previously expanded all left-side searches at once, while identity projections, normalized ingestion, and already-aligned unions unconditionally created new DataFrames.


</details>

<details><summary>📐 <b>Operating principle: bound or skip work without widening semantics</b></summary>


The sim-join output is still fully materialized; only each search/explode slice is bounded. Fast paths require exact identity conditions and fall back to the incumbent Polars select path otherwise.


</details>

<details><summary>🔹 <b>1. Bound each semantic sim-join left search slice</b></summary>


Process the left input in fixed-size slices before explode/unnest, preserving the final N×k result while limiting the transient left-side transform.


- ✏️ [`src/fenic/_backends/local/semantic_operators/sim_join.py`](https://github.com/typedef-ai/fenic/pull/361/files#diff-1d499226879e4a3adac824d5894de7f707729652ac3896fb1203e8c8bfceb0ea) `+56/−39` — Adds the internal 1,024-row default batch size, validates positive overrides, and concatenates per-slice narrow match rows in the existing result shape.
- ✅ [`tests/_backends/local/dataframe/test_semantic_sim_join.py`](https://github.com/typedef-ai/fenic/pull/361/files#diff-f1ad1947f5c046b368a8033faf746bc4691233053ff51dc340eec1f158d2915d) `+53/−0` — Uses direct local vectors with a one-row batch to observe both slice calls and the unchanged ordered match values; also covers zero batch-size rejection.

</details>

<details><summary>🔹 <b>2. Skip eager DataFrame work only for proven no-ops</b></summary>


Avoid allocation for three local physical-plan cases while preserving every non-identity or coercing behavior.


- ✏️ [`src/fenic/_backends/local/physical_plan/transform.py`](https://github.com/typedef-ai/fenic/pull/361/files#diff-bb2d8079adcf812cca4d19dbf1684e2d2d531233ef7c9621e183c7987d885bc0) `+25/−2` — Returns the child frame only for every raw input column in unchanged order, and skips union alignment only when both column lists already match exactly.
- ✏️ [`src/fenic/_backends/local/physical_plan/utils.py`](https://github.com/typedef-ai/fenic/pull/361/files#diff-22332582b8aaf70ab3b673b127bdfd2e3e875e7bd939dc825866bb1744441739) `+15/−6` — Collects only actual coercions; if there are none it returns the original frame, otherwise it retains the incumbent select-and-cast behavior.
- ✅ [`tests/_backends/local/physical_plan/test_eager_noop_fast_paths.py`](https://github.com/typedef-ai/fenic/pull/361/files#diff-ba19b1536f75f9b52ed90aca82688828ec16cc9138ff3f41c6a3e4a3b2f8ae1e) `+175/−0` — Covers row and empty frames, datetime/struct/embedding identity, coercion-required arrays, union order changes, and every projection condition that must still select.

</details>

<details><summary>🔹 <b>3. Disclose when local Polars results can alias their input</b></summary>


Public API docstrings and topic guides distinguish schema-free pass-through from paths that still coerce, reorder, or copy.


- ✏️ [`src/fenic/api/session/session.py`](https://github.com/typedef-ai/fenic/pull/361/files#diff-7e542c87b8e0836505ce4ed18f62c8b375ca4859633a3ce518ddf85b20d2bdb0) `+3/−1` — Session.create_dataframe explains that schema-free Polars input can pass through when no ingestion coercion is required.
- ✏️ [`src/fenic/api/dataframe/dataframe.py`](https://github.com/typedef-ai/fenic/pull/361/files#diff-e9ed789db86282c7c2b7b4305f0270caa4846f869f8a0b6752c1b3ff008027d1) `+6/−2` — collect(polars) and to_polars explain that a no-op local plan may return a frame aliasing the original in-memory Polars input.
- ✏️ [`docs/topics/create_dataframe_schema.md`](https://github.com/typedef-ai/fenic/pull/361/files#diff-642718cc632ab2537e30b7735d34437074755e6d552c8348d8797b2f86231947) `+8/−1` — Replaces the stale whole-path compatibility claim with the exact pass-through, coercion, and explicit-schema boundaries.
- ✏️ [`docs/topics/timestamps_and_dates.md`](https://github.com/typedef-ai/fenic/pull/361/files#diff-a27ca0337ebac428e7e230cf4888326c6f5daa8b1862a48a6a7dfd5921de28f4) `+15/−3` — Keeps the valid UTC guarantees while distinguishing canonical UTC pass-through from timezone or precision normalization that materializes a new frame.

</details>

<details><summary>🚫 <b>Deliberately NOT changed</b></summary>


This change does not include streaming execution, semantic joins, operator fusion, validation harnesses, or unrelated execution-engine work.


</details>

<details><summary>✅ <b>Verification and landing status</b></summary>


Focused pytest, Ruff, docs-check, and diff checks pass.


</details>


<sub>🤖 Walkthrough generated with `td-walkthrough` (Claude Code).</sub>

YoungVor (changes requested):
Looks good!  Only concern I have is more documentation for the change in dataframe behavior to pass-through/alias when there are no coercions.

probably worth a line in the Session.create_dataframe and DataFrame to_polars/collect docstrings.

Also create_dataframe_schema.md has stale content, for e.g. saying no-schema path has the same behavior as other dataframe paths.  timestamps_and_dates.md i think also makes some untrue assertions too.

bcallender:
> more documentation for the change in dataframe behavior to pass-through/alias when there are no coercions

Addressed in `7dbef64`:

1. `Session.create_dataframe`, `DataFrame.collect`, and `DataFrame.to_polars`
   now disclose that schema-free, no-op local execution may return a Polars
   frame aliasing the original input rather than copying it.
2. `create_dataframe_schema.md` no longer says the no-schema path is wholly
   unchanged. It now distinguishes canonical Polars pass-through from required
   ingestion coercions and explicit-schema ordering/casts, both of which still
   materialize a frame.
3. I audited the timestamp guide's UTC-first, timezone-conversion,
   cross-source-consistency, in-memory-ingestion, and conversion-table claims.
   The UTC value and conversion claims remain true. I corrected the overbroad
   cross-source “same behavior” wording and documented that canonical
   microsecond-UTC Polars input can pass through, while naive, non-UTC, or
   non-microsecond timestamps still require normalization and produce a new
   frame.

`just docs-check` passes and validates all 63 sitemap URLs, canonical metadata,
Markdown variants, and LLM assets. Ruff and `git diff --check` also pass.

YoungVor (approved):
Thank you!  Lgtm!

PR #364, opened 2026-08-13
feat(inference): add opt-in sliding-window request streaming

bcallender:
## What problem does this solve?

Row-local semantic operations currently materialize their full request list
before execution. Streaming request production bounds live memory, but a naive
window that admits only `batch_size` requests turns the memory bound into a
concurrency ceiling: standard execution submits the provider's available burst
up front, so any window smaller than that burst loses throughput on unique-row,
non-rate-limited workloads.

## What changed

This PR adds internal opt-in ordered request streaming for row-local semantic operations
while retaining the established queue, cache, token accounting, rate-limit, and
error-normalization paths.

Streaming admission has one explicit contract:

```text
live window = max(batch_size, rate-limit RPM burst)
```

`batch_size` is a minimum look-ahead, not a hard live-object memory cap when
the RPM burst is larger. The iterator fills that window, admits a successor when
the next ordered request settles, and yields responses in input order. The
window bounds not-yet-yielded futures, completed out-of-order results, and live
deduplication entries. The rate limiter remains the single authority for the
admission burst.

Streaming is disabled by default. The internal docstrings and changelog state
the effective-window contract so internal callers who need a strict B-sized memory bound
can make that tradeoff knowingly.

## Safety and behavior preservation

- Default list-based execution is unchanged.
- Ordered output, configured response-cache behavior, normalized errors,
  lifecycle events, and rate-limit behavior remain on the existing client path.
- Live-window deduplication remains bounded: a duplicate after its original
  request settles uses the configured response cache when available, otherwise
  it is a separate request.
- A focused deterministic test covers the B < RPM case: admission can exceed B
  but never exceeds the declared effective window, and output remains ordered.
- Exact-ceiling tests hold the ordered head and prove that input admission,
  in-flight futures, out-of-order results, and the live deduplication map cannot
  escape W. A separate adaptive-RPM proof shows W is captured once rather than
  growing during an active iterator.
- A later in-window failure can occur after a successor was dispatched; the
  normalized error still aborts the operation before final materialization.

## Real-provider throughput check

The retained 72-cell benchmark compares standard execution and opt-in streaming
with the same `gpt-4.1-nano` inputs, fresh local cache/database per run, and
three repetitions. The unique-row cells are the throughput gate; cache-heavy
controls (eight physical requests per run) are reported only as noisy controls.

| Unique workload | B | standard median | streaming median | streaming delta |
| --- | ---: | ---: | ---: | ---: |
| map, 200 rows | 32 | 2.791s | 2.541s | -9.0%* |
| map, 200 rows | 100 | 2.113s | 1.760s | -16.7% |
| map, 1,000 rows | 32 | 6.349s | 6.282s | -1.1% |
| map, 1,000 rows | 100 | 6.127s | 6.127s | +0.0% |
| predicate, 200 rows | 32 | 1.867s | 1.894s | +1.5% |
| predicate, 200 rows | 100 | 1.959s | 1.959s | +0.0% |
| predicate, 1,000 rows | 32 | 6.160s | 6.708s | +8.9% |
| predicate, 1,000 rows | 100 | 6.153s | 5.688s | -7.6% |

\* The initially noisy map/200/B=32 cell was rerun at seven repetitions per arm
with alternating arm order. Standard median and spread were 2.791s
(1.975–3.398s); streaming was 2.541s (1.532–3.941s), 9.0% faster.

All 72 runs completed with matching result-row and physical-request counts
between arms: 28,992 requests queued, dispatched, and settled; zero failures
and zero rate-limit events. The full matrix cost $0.431984 including its
preflight; the expanded disputed-cell sample cost $0.032200. This is a one-host,
real-provider measurement; short cache-heavy controls remain too variable to
support a throughput conclusion.

## Production-shaped pipeline check

A separate A/B used the same frozen implementation for both arms and exercised
four real production semantic-pipeline stages with real prompts, schemas,
data shaping, and model profiles. Across 24 accepted cells (1,884 physical
requests), request counts matched, provider caching was excluded, and no
failed, retried, or rate-limited events occurred.

| Pipeline stage | requests / cell | standard median | streaming median | streaming delta |
| --- | ---: | ---: | ---: | ---: |
| column classification | 290 | 13.291s | 12.430s | -6.5% |
| query metadata | 10 | 1.625s | 1.956s | +20.3% |
| filter intent | 4 | 3.475s | 3.926s | +13.0% |
| analysis summary | 10 | 5.302s | 4.609s | -13.1% |

The high-fanout classification stage was faster with 25.9 MiB lower median peak
RSS, and the large-prompt summary stage was 13.1% faster. The two very small
stages were slower and visibly noisy, so the honest conclusion is a
stage-dependent win rather than a universal overlap gain. Peak-RSS ranges
overlapped and include a roughly 450–500 MiB process baseline; this supports no
memory regression more strongly than it proves a memory saving.

## Adversarial ceiling verification

The bounds are mutation-tested, not only green-path tested. An implementation
that deliberately raises admission to `RPM * 2` makes all four exact-ceiling
proofs fail: input admission, live dedup-map size, the default-RPM whole-list
case, and the post-settlement cache boundary. A second mutant that rereads an
increased adaptive RPM inside both fill loops fails the captured-once proof at
the W+1 overflow witness. The production implementation is unchanged by these
test-only checks.

## How to verify

- `uv run pytest tests/_inference/test_model_client_cache_behavior.py tests/_inference/test_request_lifecycle.py tests/_inference/test_streaming_parity_column_classification_shape.py -q` — 38 passed
- `uvx ruff check src/fenic/_inference/model_client.py src/fenic/_inference/language_model.py tests/_inference/test_model_client_cache_behavior.py`
- `trunk check --ci`
- `git diff --check`

## Deliberately not changed

There is no default-on behavior change, no change to final result
materialization, and no fusion or semantic-join work in this PR.

## Validation campaign

This change was held to an exhaustive measurement bar before merge; the full write-ups are committed in this PR and summarized here.

**Design grounding** — `specs/td-flow/admission-watermark/research.md` and `design.md`: the instrumented investigation that motivated the indexed-slot admission design (slot-wait attribution showed the prior design's wait dominated wall time), the chosen retained-window contract, and the ordering guarantees.

**Synthetic matrix, real providers** — 8-cell matrix over operator × row-uniqueness × window size against live provider endpoints: 7 of 8 unique-row cells measured streaming at parity or faster than list-based execution (e.g. map at 1,000 unique rows, W=32: −1.2%; predicate: −4.6%). A provider-free grading pass separately proved the admission path adds no measurable serialization, with a negative control demonstrating the harness detects one when it exists.

**Real-corpus A/B** — a four-stage classification/metadata/filter/summary workload (300/10/3/10 requests per cell) run as interleaved A/B cells with fresh caches, per-request nonces, and cache-contamination acceptance bars. An initial 3-repetition pass showed two adverse stage deltas; a dedicated investigation (`specs/td-flow/td-4729-colclass-regression/evidence.md`) attributed both to provider service-time drift, not this design: provider-free reproduction at the exact shape measures the arms within ±0.33% across two environments with every streaming-only cost accounted in the tens of milliseconds, and three full-scale interleaved real-corpus pairs with complete per-request lifecycle streams measured streaming at −5.2% median — including one cell whose entire adverse wall decomposes into two provider-tail requests (15.1 s and 18.4 s service time) against normal client dispatch receipts.

**Permanent guard** — `tests/_inference/test_streaming_parity_column_classification_shape.py`, added in this PR: a provider-free parity benchmark at the 290-request structured-output shape, with a committed negative control (serializing the caps must fail the test, proven at 24× wall inflation). A future change that introduces a real client-side regression at this shape fails loudly.


bcallender:
## Opt-in sliding-window semantic request streaming

Combines lifecycle visibility, bounded row-local request streaming, and sliding-window admission in one reviewable implementation while preserving default execution behavior.

**+1023 / −53** diff (13 files) · **6** commits · **a91d8af** head · **≤ B** Live request cap · **38 passed** Focused provider-free tests · **10/10** No-barrier proof

<details><summary>❓ <b>Stream requests without recreating a chunk barrier</b></summary>


Row-local semantic work can admit a successor as the next ordered request settles.


</details>

<details><summary>🔹 <b>Instrument lifecycle and expose bounded streaming</b></summary>


Private lifecycle metrics and iterator plumbing remain behind explicit opt-in execution.


- ✏️ [`src/fenic/_backends/local/semantic_operators/base.py`](https://github.com/typedef-ai/fenic/pull/364/files#diff-fa97408decd9dd6665ed1c577f96faf749fd0cd82615f0ea2c4a3e783b0a72b4) `+64/−8` — Adds the internal execution hooks used by row-local semantic operators while leaving final materialization and the default path unchanged.
- ✏️ [`src/fenic/_inference/language_model.py`](https://github.com/typedef-ai/fenic/pull/364/files#diff-b91296a57b1da546e70d20bd05d2f53ae13dd20e52286fd8e57c543e7b143cdf) `+49/−1` — Makes row-local language-model completion streaming available to operators that explicitly select it.
- ✏️ [`src/fenic/_inference/model_client.py`](https://github.com/typedef-ai/fenic/pull/364/files#diff-19d194579c91fb3369bc5285976fc0677453dcd2d2cc54a04ac3839b716c38b0) `+177/−6` — Replaces fixed chunk waits with a deque-backed window: settle the ordered head, evict its live dedup entry, admit a successor, then yield the response.
- 🆕 [`src/fenic/_inference/request_lifecycle.py`](https://github.com/typedef-ai/fenic/pull/364/files#diff-38f9705a25b7774a589045214429d74ee643e41c9126f0870b1120d2755a6538) `+157/−0` — Collects request lifecycle events and derives idle-gap measurements while excluding rate-limit time.

</details>

<details><summary>🔹 <b>Prove streaming semantics with deterministic clients</b></summary>


Tests cover bounded concurrency, output order, cache/dedup scope, error normalization, and direct operator use.


- ✏️ [`tests/_inference/test_model_client_cache_behavior.py`](https://github.com/typedef-ai/fenic/pull/364/files#diff-c0782b2d93f6b549a7da3169c09226590fe5228a80c231ffc9a3c4e7ad9a5174) `+265/−0` — Demonstrates successor admission without an inter-chunk barrier and keeps the active request count bounded.
- 🆕 [`tests/_inference/test_request_lifecycle.py`](https://github.com/typedef-ai/fenic/pull/364/files#diff-afee04a215d61d04a9c582b5247465997118d878a67f25ab76d6ef5ab17ea353) `+181/−0` — Covers lifecycle events and rate-limit-aware idle-gap accounting.
- ✏️ [`tests/_backends/local/functions/test_semantic_predicate.py`](https://github.com/typedef-ai/fenic/pull/364/files#diff-b024081d100d81e4e836220b094a3d28efee79f79aacc2af01611da06171b5e9) `+64/−0` — Ensures the transpiler-built predicate path reaches the streaming iterator when explicitly enabled.

</details>

<details><summary>🚫 <b>Deliberately NOT changed</b></summary>


Streaming remains opt-in; no fusion, semantic-join, or incremental final-result behavior enters this PR.


</details>

<details><summary>✅ <b>Verification</b></summary>


Provider-free focused tests, repeated no-barrier proof, lint, Trunk, and whitespace checks pass.


</details>


<sub>🤖 Walkthrough generated with `td-walkthrough` (Claude Code).</sub>

bcallender:
## Opt-in throughput-preserving semantic request streaming

Combines lifecycle visibility, opt-in row-local request streaming, and an admission window that preserves the client’s existing rate-limit burst throughput.

**+1319 / −57** diff (13 files) · **10** commits · **03e29df** head · **max(B, RPM burst)** Effective window · **17 passed** Focused provider-free tests · **72/72 complete** Real-provider matrix

<details><summary>❓ <b>Stream requests without an artificial concurrency ceiling</b></summary>


A B-sized look-ahead must not prevent the client from using its established RPM burst capacity.


</details>

<details><summary>🔹 <b>Keep one admission authority</b></summary>


The existing rate limiter supplies the burst watermark; no second concurrency policy is introduced.


- ✏️ [`src/fenic/_inference/model_client.py`](https://github.com/typedef-ai/fenic/pull/364/files#diff-19d194579c91fb3369bc5285976fc0677453dcd2d2cc54a04ac3839b716c38b0) `+181/−6` — The batch client fills and refills the RPM-aware live window, keeping live futures, completed out-of-order results, and dedup entries bounded by that window.
- ✏️ [`src/fenic/_inference/language_model.py`](https://github.com/typedef-ai/fenic/pull/364/files#diff-b91296a57b1da546e70d20bd05d2f53ae13dd20e52286fd8e57c543e7b143cdf) `+52/−1` — Language-model and client docstrings explain that batch_size is minimum look-ahead rather than a hard memory cap when the RPM burst is larger.

</details>

<details><summary>🔹 <b>Prove the new B < RPM boundary</b></summary>


Deterministic tests pin exact W ceilings, captured-once behavior under adaptive RPM, ordering, and slot-release behavior.


- ✏️ [`tests/_inference/test_model_client_cache_behavior.py`](https://github.com/typedef-ai/fenic/pull/364/files#diff-c0782b2d93f6b549a7da3169c09226590fe5228a80c231ffc9a3c4e7ad9a5174) `+554/−4` — Covers exact input and dedup-map ceilings, captured-once adaptive-RPM behavior, no inter-chunk barrier, output ordering, cache scope, and failure hygiene without provider calls.

</details>

<details><summary>✅ <b>Measure throughput under a real provider</b></summary>


The retained 72-cell matrix eliminates the prior systematic unique-row regression; a seven-repetition adjudication resolves its one noisy cell.


</details>

<details><summary>✅ <b>Check the production-shaped pipeline</b></summary>


A 24-cell Mattermost pipeline A/B found a 6.5% classification gain and a 13.1% large-summary gain, while two very small stages remained noisy and slower.


</details>

<details><summary>✅ <b>Make ceiling regressions turn red</b></summary>


Over-admission and live adaptive-RPM reread mutants fail the exact-ceiling tests at their W+1 witnesses.


</details>

<details><summary>🚫 <b>Deliberately NOT changed</b></summary>


Streaming stays opt-in; default execution, final materialization, fusion, and semantic joins stay outside this change.


</details>


<sub>🤖 Walkthrough generated with `td-walkthrough` (Claude Code).</sub>

src/fenic/_inference/request_lifecycle.py
```diff
@@ -0,0 +1,157 @@
...
+            continue
+
+        if event.event == "rate_limited":
+            state.rate_limited_at_ns.setdefault(request, event.timestamp_ns)
+            continue
+
+        if event.event == "dispatched":
+            queued_at_ns = state.queued_at_ns.get(request)
+            if queued_at_ns is not None:
+                queue_delays_ns.append(event.timestamp_ns - queued_at_ns)
```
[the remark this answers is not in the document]
bcallender: > `queued_at_ns` entries are never removed after dispatch.

Fixed in `90a013c`. First dispatch now pops the queued timestamp, which bounds retained metric state and prevents a retry dispatch from counting the original queue delay again. The retry lifecycle test now pins the queue-delay total.

src/fenic/_inference/model_client.py
```diff
@@ -473,6 +531,88 @@ def make_batch_requests(
...
+        ``batch_size`` remains the caller's minimum look-ahead; the rate-limit
+        burst is also admitted so streaming does not impose a lower concurrency
+        ceiling than the existing list-based path. Once the next ordered response
+        settles, its successor is admitted before that response is yielded, so a
+        later slow request does not recreate the old whole-batch barrier.
+
+        Request fingerprint deduplication is intentionally scoped to a single
+        bounded window. Repeated requests after their earlier entry has settled
+        are served without a second provider call when the configured response
+        cache contains the first result; without a cache they are independent
```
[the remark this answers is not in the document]
bcallender: > Please align the docstring with the actual live-window semantics.

Fixed in `90a013c`. The docstring now states that the dedup entry expires after its ordered result is consumed and removed from the live window, matching the implementation boundary.

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 * reduce the default install footprint by moving PDF parsing, clustering, and similarity join dependencies behind opt-in `pdf`, `cluster`, and `sim-join` extras
 
+### Behavior Changes
+
+* add opt-in ordered request streaming that preserves the rate limiter's burst concurrency; its ``batch_size`` is a minimum look-ahead and the live window is ``max(batch_size, rpm-burst)``, so callers must not treat ``batch_size`` alone as a hard memory cap. When enabled, duplicate requests after an earlier live-window entry settles are independent unless the response cache serves the later request, and a successor may be dispatched before a later in-window request fails
```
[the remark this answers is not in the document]
bcallender: > Updating this wording will better match observed behavior.

Fixed in `90a013c`. The changelog now uses the same ordered-consumption/live-window-removal boundary as the implementation and public docstring.

rust-toolchain.toml
```diff
@@ -0,0 +1,2 @@
+[toolchain]
+channel = "1.94.1"
```
cpard, line 2: This pins Rust below the MSRV of locked sysinfo 0.39.6 (Rust 1.95), so clean uv/maturin and cargo --locked builds fail. Please raise or remove the pin.

src/fenic/_inference/model_client.py
```diff
@@ -473,6 +571,240 @@ def make_batch_requests(
...
+
+            record_stage(
+                "window_admission",
+                admission_started_ns,
+                stage_request_index=slot_index,
+            )
+            request_key = self.get_request_key(request) if request is not None else None
+            dispatch_started_ns = stage_started_ns()
+            request_futures, _, _ = self._submit_batch_requests(
+                [request],
```
cpard, line 670: This singleton submission causes one cache get_batch call per row. In a 20-row duplicate probe, streaming made 20 lookups versus 1 on the batch path (both made one provider call). Please batch cache reads within the bounded window.

src/fenic/_inference/model_client.py
```diff
@@ -473,6 +571,240 @@ def make_batch_requests(
...
+                settled_pending_indices.put(settled_index)
+        except Exception as e:
+            # Preserve the public batch API's error boundary for Polars callbacks.
+            raise ExecutionError(str(e)) from e
+
+    @staticmethod
+    def _streaming_slot_caps(look_ahead_basis: int) -> tuple[int, int]:
+        """Return the bounded pending and completed slot capacities for streaming."""
+        return (
+            min(1_000, 3 * look_ahead_basis),
```
cpard, line 804: These caps do not match the documented L=max(batch_size, rpm) contract: retained state can reach 3L+10L, while pending admission is capped below L when L>1000. Please align the implementation and public contract.

CHANGELOG.md
```diff
@@ -25,6 +25,9 @@
 ### Performance Improvements
 
 * **local:** avoid redundant DataFrame work and bound sim-join memory ([#361](https://github.com/typedef-ai/fenic/issues/361)) ([2c3adc5](https://github.com/typedef-ai/fenic/commit/2c3adc5c1d4ea6ce0060eca89904a41bf502139f))
+### Behavior Changes
```
cpard, line 28: This new behavior is listed under the already released 0.13.0 section. Please move it under Unreleased.

rust-toolchain.toml
bcallender: Fixed in `84d7578`. The repository now pins Rust 1.95.0, which matches the locked `sysinfo` 0.39.6 MSRV. Clean builds with the pinned toolchain now pass through both `cargo check --locked` and `maturin build --locked`.

src/fenic/_inference/model_client.py
bcallender: Fixed in `84d7578`. Streaming admission now collects each bounded group and submits that group once through the existing batch cache path. The provider-free 20-row duplicate regression now asserts one `get_batch` call and one provider call.

src/fenic/_inference/model_client.py
bcallender: Fixed in `84d7578`. Pending and completed slots now share one retained budget `L = max(batch_size, rpm)`, with exact-bound tests below and above 1,000 plus adaptive-RPM and dedup coverage. The provider-free top-stack matrix changed from a bound failure at 115 outstanding admissions for `L=100` to PASS at the exact bound. Streaming median was 300.392 ms before and 298.746 ms after, so the contract fix did not reduce throughput in this benchmark.

CHANGELOG.md
bcallender: Fixed in `84d7578`. The ordered-streaming behavior entry now sits under `Unreleased`, before the 0.13.0 release heading.

src/fenic/_backends/local/semantic_operators/base.py
```diff
@@ -86,10 +105,32 @@ def send_requests(
...
+        )
+        for response in responses:
+            yield response.completion if response else None
+
 class BaseOperator(Generic[ModelResponseType, OperatorOutputType], ABC):
     """Abstract base class for any operator that sends builds LM requests and processes responses."""
 
     request_sender: RequestSender[ModelResponseType]
+    stream_requests = False
+    request_batch_size = 100
```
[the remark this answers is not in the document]
bcallender: Kept this intentionally internal rather than widening the public API in a review remediation. Commit e2c6721 changes the Unreleased entry to say “internal opt-in” and “internal callers,” and the PR description now uses the same wording. The path remains disabled by default and is only wired by internal operators/tests; a public configuration surface is out of scope for this foundation PR.

src/fenic/_inference/model_client.py
```diff
@@ -473,6 +571,224 @@ def make_batch_requests(
...
+            for slot_index, request_key, req_future in zip(
+                admitted_indices,
+                admitted_keys,
+                request_futures,
+                strict=True,
+            ):
+                record_stage(
+                    "request_dispatch",
+                    dispatch_started_ns,
+                    stage_request_index=slot_index,
```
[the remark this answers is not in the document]
bcallender: Fixed in e2c6721. The batch-wide `request_dispatch` record is now emitted once immediately after `_submit_batch_requests`, outside the per-slot loop, so aggregate duration is no longer multiplied by group size. The 20-request cache-batching test now attaches a lifecycle collector and asserts exactly one dispatch-stage event for that admission group. Focused lifecycle/streaming tests pass (37 passed), and exact-head Claude review returned Ready with no findings.

src/fenic/_inference/model_client.py
```diff
@@ -473,6 +571,224 @@ def make_batch_requests(
...
+
+                record_stage(
+                    "window_admission",
+                    admission_started_ns,
+                    stage_request_index=slot_index,
+                )
+                admitted_requests.append(request)
+                admitted_keys.append(
+                    self.get_request_key(request) if request is not None else None
+                )
```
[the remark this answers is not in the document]
bcallender: > `admitted_keys` uses the overridable `get_request_key()`, but `_submit_batch_requests()` indexes `unique_futures` with `_safe_build_request_key()`.

Fixed in e0b448b. Admission tracking now uses the same canonical `_safe_build_request_key(request, slot_index)` plus opaque object-identity fallback as submission, so slot refcounts release the actual `unique_futures` entry at the ordered boundary. A regression with a deliberately mismatched legacy override and a one-slot window proves two sequential same-content requests make two physical calls and the dedup map empties. The focused suite passes (38 tests), and exact-head Claude Sonnet 5 review through Droid returned Ready with no findings.

bcallender:
> Test the full 290-request column-classification shape.

Fixed in e0b448b. The parity arm now runs all 290 requests documented by the regression evidence, while the deliberately serialized negative control stays at 16 rows so the guard remains fast. The focused suite passes (38 tests), and exact-head Claude Sonnet 5 review through Droid returned Ready with no findings.

PR #376, opened 2026-08-15
ci: prevent manual runs from satisfying PR checks

bcallender:
## Summary

Removes `workflow_dispatch` from the pull-request test workflow. This keeps
`allcheckspassed` scoped to pull-request test runs: a manually triggered run
can no longer supersede or satisfy the pull-request merge gate.

The aggregate continues to ignore obsolete reruns of the pull-request workflow.

## Validation

- `git diff --check`
- `actionlint .github/workflows/test.yaml`; only four existing shellcheck
  advisories in unchanged commands remain.
- Live regression: after canceling the pull-request test run at the same SHA,
  a successful manually dispatched `docs check` run did not satisfy the
  aggregate; `allcheckspassed` remained failed until the pull-request checks
  were rerun.
- Rebased onto current `main` with an equality-only `git range-diff`; the final
  change remains limited to `.github/workflows/test.yaml`.


.github/workflows/rust_cache_measurement.yaml
```diff
@@ -0,0 +1,62 @@
+---
+name: Rust cache measurement
```
scottames: This name is very confusing. We're still running tests here, no? Or what is the point of this workflow?

bcallender: Good catch. The workflow did run the full credentialed test suite, so it was not a narrowly scoped measurement and its name obscured that cost. It existed only to collect the one-time cold/warm Rust-cache evidence; that work has now landed, so I removed the workflow and its aggregate exclusion. The PR now contains only the gate fix: the PR test workflow no longer accepts manual dispatches, while the aggregate still ignores obsolete PR reruns.

scottames (approved):

PR #301, opened 2026-04-01
feat: add base_url support for OpenAI and Anthropic model configs

bcallender:
Allow users to point OpenAI and Anthropic models at custom endpoints
(proxies, gateways) via an optional base_url field on model configs.

- Thread base_url through config resolution, providers, and clients
- Include base_url in cache key fingerprint to prevent cross-endpoint
  cache poisoning
- Skip API key validation for providers with custom base URLs since
  proxies may not expose the /models endpoint

PR #238, opened 2025-09-29
feat: add Ollama as chat completion and embedding provider

iceysteel:
## TL;DR

  Adds native Ollama provider support to fenic, enabling local LLM inference with automatic model discovery, dynamic embedding dimensions,
   and optimized batch processing.

  ## What changed?

  This PR introduces comprehensive Ollama integration as a first-class model provider in fenic, allowing users to run semantic operations
  with locally hosted models. The implementation includes:

  ### 1. Complete Ollama Provider Implementation (~856 new lines):
  - `src/fenic/_inference/ollama/ollama_provider.py` (45 lines): Core provider class with connection management
  - `src/fenic/_inference/ollama/ollama_batch_chat_completions_client.py` (365 lines): Batch completion client with structured output
  support and completion time tracking
  - `src/fenic/_inference/ollama/ollama_batch_embeddings_client.py` (268 lines): Batch embedding client with dynamic dimension detection
  - `src/fenic/_inference/ollama/ollama_model_manager.py` (178 lines): Model metadata management using Ollama's `/api/show` endpoint for
  capabilities detection

  ### 2. Enhanced Rate Limiting Strategy:
  - Extended `src/fenic/_inference/rate_limit_strategy.py` with `OllamaQueueAwareRateLimitStrategy` that respects local model constraints
  (OLLAMA_NUM_PARALLEL, queue limits)

  ### 3. Dynamic Model Management:
  - Auto-pull functionality for missing models during validation in `src/fenic/core/_inference/model_catalog.py`
  - Dynamic embedding dimension detection from model metadata instead of hardcoded values
  - Capabilities-based model type detection (embedding vs completion models)

  ### 4. API Integration:
  - Added `OllamaLanguageModel` and `OllamaEmbeddingModel` configuration classes in `src/fenic/api/session/config.py`
  - Exported Ollama models in main API (`src/fenic/api/__init__.py`)
  - Updated session config to support Ollama provider

  ### 5. Example Updates:
  - All examples now support `--language-model-provider ollama` and `--embedding-model-provider ollama` CLI arguments
  - Updated `examples/hello_world/hello_world.py`, `examples/feedback_clustering/feedback_clustering.py`,
  `examples/news_analysis/news_analysis.py`, `examples/semantic_joins/semantic_joins.py`, and `examples/enrichment/enrichment.py`

  ### 6. Bug Fixes:
  - Fixed JSON structured output for local models by simplifying prompts to avoid schema confusion
[...]

PR #381, opened 2026-08-18
feat: target OpenAI-compatible endpoints with arbitrary model names

zenprocess:
## Use case

`base_url` (added in #301) lets `OpenAILanguageModel` / `OpenAIEmbeddingModel` point at a custom endpoint, but `model_name` is still a `Literal` of OpenAI catalog names. That covers a proxy or gateway that forwards to OpenAI, and stops short of the other common case: an OpenAI-compatible server hosting a **different** model — vLLM, LiteLLM, Ollama, llama.cpp, TGI, SGLang.

On a self-hosted deployment where external LLM calls are not permitted, there is no supported way to reach local inference today:

```python
fc.OpenAILanguageModel(
    model_name="qwen3-coder",                 # pydantic literal_error
    rpm=100,
    tpm=100_000,
    base_url="http://vllm.internal:8000/v1",
)
```

The transport is already fine — with a catalog model name, requests do reach a custom `base_url` and structured output parses correctly. The only things in the way are the name check and the fact that fenic has no context window, output limit, or price for a model it has never heard of.

(#237 / #238 proposed a full Ollama provider and were closed as too large to land. This is the small version: it builds on the `base_url` machinery already in `main` instead of adding a provider.)

## The change

`model_name` on both OpenAI config classes becomes `Union[<catalog literal>, str]`, and each class gains an optional nested `ModelParameters` describing an out-of-catalog model:

```python
fc.OpenAILanguageModel(
    model_name="qwen3-coder",
    rpm=100,
    tpm=100_000,
    base_url="http://vllm.internal:8000/v1",
    model_parameters=fc.OpenAILanguageModel.ModelParameters(
        context_window_length=32_768,
        max_output_tokens=4_096,
    ),
)

fc.OpenAIEmbeddingModel(
    model_name="bge-m3",
    rpm=100,
    tpm=100_000,
    base_url="http://vllm.internal:8000/v1",
    model_parameters=fc.OpenAIEmbeddingModel.ModelParameters(
        output_dimensions=1024,
        max_input_size=8_192,
    ),
)
```

A `model_validator` registers the declared parameters into `model_catalog` under `ModelProvider.OPENAI` via the existing `model_catalog.add_model()` — the same mechanism `OpenRouterModelProvider` already uses for models it discovers at runtime. Every downstream lookup (`SemanticConfig.validate_models`, `LanguageModel`, `OpenAIChatCompletionsCore`, cost accounting) then resolves normally, so nothing else needed to change.

Two rules keep it honest:

- **`model_parameters` requires `base_url`.** The parameters describe a model on *your* endpoint, so declaring them against `api.openai.com` is a mistake rather than a configuration.
- **An out-of-catalog `model_name` requires `model_parameters`.** fenic needs the context window and output limit to batch requests; silently guessing them would be worse than asking.

Token costs default to `0.0`, which is what self-hosted inference costs. A session then reports `$0.00` instead of billing local tokens at OpenAI's rates.

## Backwards compatibility

No behaviour change for any existing configuration:

- Catalog model names take the same path as before — the validator returns early as soon as the name resolves in the catalog.
- `base_url` on its own behaves exactly as it does today.
- An unrecognized name with no `model_parameters` is still rejected, and typos are still caught (`model_name="gpt-4.1-nanoo"` fails).
- No new dependencies, no new provider, no changes to the inference clients or the resolved-config shape.

The one visible difference: for a bad model name the exception is now fenic's `ConfigurationError` carrying the catalog's own `Supported Models: ...` message, rather than pydantic's `literal_error`, because the check moved from the field annotation into a validator. `ConfigurationError` is already what `SemanticConfig.validate_models` raises for the same mistake one layer up.

## Tests

Nine tests added to `tests/api/test_session.py`, alongside the existing `base_url` block:

- language and embedding models accepted with `base_url` + `model_parameters`, asserting both the resolved config and the resulting catalog entry
- declared token costs are the ones `calculate_completion_model_cost` uses
- **the declared `max_output_tokens` is enforced at plan time** — a `semantic.map` requesting more raises `[openai:<model>] max_output_tokens must be a positive integer less than or equal to 512`, i.e. `validate_effective_output_token_limit` reads the declared cap rather than a missing catalog entry
- out-of-catalog name without `model_parameters` raises `ConfigurationError`
- `model_parameters` without `base_url` raises `ConfigurationError`
- an unrecognized name with no endpoint configured is still rejected
- non-positive `output_dimensions` is rejected

That plan-time test needs no reachable server: because `base_url` is set, `SessionModelRegistry` skips provider API-key validation, so the session is built without touching the endpoint.

`uv run pytest tests/api/test_session.py` → **65 passed**. Three tests in that file (`test_local_session_with_*`) fail on my machine because they need a live `OPENAI_API_KEY`; they fail identically on unmodified `main` here. Across `tests/api tests/core tests/_backends` the failure set is byte-identical with and without this change (74 API-key-dependent failures on both), so there are no regressions.

`ruff check` is clean on both changed files.

Verified end-to-end against a local OpenAI-compatible server: `semantic.classify` over three rows with an arbitrary model name, structured output parsed, usage accounted, reported cost `$0.00`.

`.claude/skills/fenic-mechanics/reference/config-and-types.md` is regenerated (`just generate-reference`) for the two changed field lists, so the reference-drift check stays green.

## Notes

- `TiktokenTokenCounter` falls back to `o200k_base` for an unknown model name, so token *estimates* used for rate-limit pacing are approximate for non-OpenAI tokenizers. That is pre-existing fallback behaviour and affects pacing only, not correctness — happy to add a tokenizer hook in a follow-up if you want one.
- Registration is process-wide and idempotent: the first config to declare a name wins and a later config with the same name reuses it. Happy to make a conflicting redeclaration an error instead if you'd prefer.
- I kept this to the OpenAI classes, since that is where the OpenAI-compatible wire format lives. `AnthropicLanguageModel` also has `base_url`, but its catalog names are Anthropic-specific, so I left it alone.
- Unrelated to this change, but worth noting for anyone using it: the OpenAI SDK still requires `OPENAI_API_KEY` to be set even when the endpoint needs no key, so a keyless self-hosted server needs a placeholder value in the environment.


src/fenic/api/session/config.py
```diff
@@ -537,22 +557,116 @@ class OpenAILanguageModel(BaseModel):
         ```
     """
 
-    model_name: OpenAILanguageModelName = Field(
-        ..., description="The name of the OpenAI model to use"
+    model_name: Union[OpenAILanguageModelName, str] = Field(
```
bcallender, line 560: This widening is where the separate class would start: `OpenAICompatibleLanguageModel` with `model_name: str`, `base_url: str`, and `model_parameters` all required, leaving `OpenAILanguageModel.model_name` as the `Literal` so a typo here still gets the catalog error.

src/fenic/api/session/config.py
```diff
@@ -614,17 +734,138 @@ class OpenAIEmbeddingModel(BaseModel):
...
+            model_parameters=OpenAIEmbeddingModel.ModelParameters(
+                output_dimensions=768, max_input_size=512
+            ),
+        )
+        ```
     """
 
-    model_name: OpenAIEmbeddingModelName = Field(
-        ..., description="The name of the OpenAI embedding model to use"
+    model_name: Union[OpenAIEmbeddingModelName, str] = Field(
```
bcallender, line 753: Same shape for embeddings: an `OpenAICompatibleEmbeddingModel` with the embedding `ModelParameters` required, rather than the optional pair on `OpenAIEmbeddingModel`.

src/fenic/api/session/config.py
```diff
@@ -1904,6 +2145,60 @@ def _validate_embedding_profile(
...
+        model_parameters: The user-declared parameters, if any.
+        catalog_lookup: Catalog accessor for the relevant model kind.
+        unsupported_model_error: Builds the error message listing supported models.
+        config_class_name: The config class name, used in error messages.
+
+    Raises:
+        ConfigurationError: If `model_parameters` is set without `base_url`, or if the model
+            name is outside the OpenAI catalog and `model_parameters` is missing.
+    """
+    if model_parameters is not None and base_url is None:
```
bcallender, line 2179: This validator disappears on a separate class, since both fields become plain required fields there.

src/fenic/api/session/config.py
```diff
@@ -1904,6 +2145,60 @@ def _validate_embedding_profile(
...
+        raise ConfigurationError(
+            f"{config_class_name} 'model_parameters' requires 'base_url'. Set 'base_url' to "
+            f"the OpenAI-compatible endpoint serving '{model_name}', or remove "
+            f"'model_parameters' to use a model from the OpenAI catalog."
+        )
+
+    if catalog_lookup(ModelProvider.OPENAI, model_name) is not None:
```
bcallender, line 2186: The collision lives here: a user-declared model whose name matches a real OpenAI entry returns early and inherits OpenAI's context window and pricing. Registering under a distinct provider identity for user-declared models avoids that structurally, and also stops two configs declaring the same name from silently keeping only the first.

bcallender (changes requested):
Thank you for this — it's a well-built PR. The validation is thorough, the
errors point at the fix rather than just naming the problem, and the docstrings
carry real examples. This is a capability we want, and you've thought about the
parts that are easy to skip.

I'd like to ask for one structural change before we merge, and I want to lay out
the reasoning rather than just the request.

**The ask: put this on its own config class rather than widening
`OpenAILanguageModel` and `OpenAIEmbeddingModel`.**

Three things pushed us here.

*It matches what the module already does.* `GoogleDeveloperLanguageModel` and
`GoogleVertexLanguageModel` are separate classes for the same Gemini models,
split purely because the deployment surface differs. `OpenRouterLanguageModel`
is separate even though OpenRouter speaks the OpenAI wire format. Each time the
question has come up, the answer has been a class per surface rather than a mode
flag on an existing one.

*The PR already concedes the shape varies.* There are two `ModelParameters`
classes here with different fields — context window and output cap for
completions, output dimensions and max input size for embeddings. That's correct!
But it means the shape is varying by nesting rather than by class, and
`OpenAILanguageModel` now needs a validator to enforce that `base_url` and
`model_parameters` are required together and invalid apart. When a config class
needs a cross-field rule to keep two modes from mixing, that's usually two config
types sharing one name.

*Widening `model_name` costs the existing path something real.* With
`Union[OpenAILanguageModelName, str]`, a typo in a genuine OpenAI model name no
longer produces the catalog error listing valid models — it falls through to the
`str` branch and reports that `base_url` and `model_parameters` are required.
That sends someone chasing a self-hosted-endpoint story when they've actually
just misspelled a model name. Every current user pays that cost for a feature
they aren't using; on a separate class, the `Literal` stays intact.

**One behavioral thing worth fixing while you're in here**, independent of where
the class lives. `_register_openai_compatible_model` registers user-declared
models under `ModelProvider.OPENAI` in the module-level catalog singleton, keyed
on name alone, and returns early when the name already resolves. So a user who
names their self-hosted model something that collides with a real OpenAI entry
gets their declared parameters silently discarded, and fenic sizes and prices
their local endpoint using OpenAI's real context window and per-token costs. The
same first-write-wins behavior applies to two configs declaring the same name
with different parameters. A distinct provider identity for user-declared models
removes both collisions structurally.

**What we are not asking for.** We deliberately don't want three classes
(`VLLMLanguageModel`, `LiteLLMLanguageModel`, and so on) at this point — today
the config shape is genuinely identical across those backends, so it'd be the
same class written three times. One `OpenAICompatibleLanguageModel` states the
actual contract: you're describing a model fenic has never heard of, at an
endpoint fenic doesn't control. If we later add capability discovery, that has to
be per-backend anyway (vLLM's `/v1/models` and Ollama's `/api/show` are different
shapes), and per-backend subclasses become cheap once the base class exists.

The move should be mechanical rather than a redesign — the validation, the
parameter classes, the catalog registration, and the tests all carry over. Happy
to talk through any of it, and thanks again for picking this up.

bcallender:
Hi @zenprocess, checking in on this one. We still want this capability in fenic, and the PR is most of the way there.

Two things are outstanding:

1. **The structural change from the review above**: move the feature onto its own config classes (`OpenAICompatibleLanguageModel` and `OpenAICompatibleEmbeddingModel`, with `model_name`, `base_url` and the model parameters required) instead of widening `OpenAILanguageModel` / `OpenAIEmbeddingModel`, and register user-declared models under a distinct provider identity so a name that matches a real OpenAI catalog entry cannot silently inherit its context window and pricing. The inline comments mark where each piece lands.
2. **A signed commit.** `main` requires verified signatures and `f4b2c0ed` is unsigned.

<details>

```bash
```

```bash
git push --force-with-lease
```

</details>

If you have the time and interest to finish it, wonderful. If not, no hard feelings at all: maintainer edits are enabled, so if we have not heard back by **Monday 5 October** we will pick it up from here, either by adding commits on top of yours on this branch or, if the restructure turns out large, in a follow-up PR that builds on your commit. Either way you stay the author of your work and get the credit in the release notes.

PR #384, opened 2026-08-19
fix(inference): stop a failed batch from failing later batches

zenprocess:
## Symptom

One failed batch permanently breaks the model client. Every later query on the same thread fails with a **stale exception from the earlier, unrelated operation** — even after the provider has recovered and is returning perfectly good responses.

Reproduced against a local OpenAI-compatible server rigged to return exactly one non-conforming structured-output response and strictly conforming JSON from then on. Three sequential `semantic.classify` queries:

```
query 1 (server returns 1 bad response):     ExecutionError: 1 validation error for EnumModel
query 2 (server now STRICTLY CONFORMING):    ExecutionError: 1 validation error for EnumModel   <- stale
query 3 (server now STRICTLY CONFORMING):    ExecutionError: 1 validation error for EnumModel   <- stale
```

Queries 2 and 3 never had a problem. Their responses were valid. They failed because query 1 did.

This is badly misleading in practice: the reported error describes a schema and an input value from an operation that already finished, while `This error occurred in the following expression` names the *current* expression. Debugging leads straight to the wrong operator. It also means a single transient provider hiccup poisons the rest of the session rather than failing one batch.

## Root cause

`_handle_exception` records the failure against the submitting thread so a long batch can fail fast instead of waiting on every future:

```python
with self.thread_exceptions_lock:
    self.thread_exceptions[queue_item.thread_id] = exception
```

`_maybe_raise_thread_exception` is called repeatedly while submitting a batch and re-raises it:

```python
with self.thread_exceptions_lock:
    if current_thread_id in self.thread_exceptions:
        raise self.thread_exceptions[current_thread_id]
```

Nothing ever removes the entry. `self.thread_exceptions` is created once per client in `__init__`, and there is no `pop`, `del` or `clear` anywhere in the file. Because clients are cached per model for the life of a session, the record survives every subsequent batch on that thread.

Any unhandled exception reaches this path, so this is not specific to one failure mode — a structured-output validation error is simply the easiest way to trigger it.

## The change

Two lines of intent:

- clear the calling thread's record when a batch starts, so a batch reports only its own failures;
- remove the exception as it is raised, so it is delivered exactly once.

```python
def _clear_thread_exception(self):
    """Discard any exception recorded for the calling thread."""
    with self.thread_exceptions_lock:
        self.thread_exceptions.pop(threading.get_ident(), None)
```

Fail-fast behaviour within a batch is unchanged: the record is still written on failure and still surfaces on the next `_maybe_raise_thread_exception()` call inside that batch. Only cross-batch leakage is removed. Nothing about success paths, retry classification, rate limiting or metrics is touched.

Clearing at the start of the batch is what fixes the user-visible bug; popping on raise is what keeps the invariant true if a batch calls the check more than once after a failure. Both are needed — with only the pop, the first stale delivery still lands on the following batch.

## Tests

`tests/_inference/test_model_client_thread_exception.py`, built on the `DummyProvider` / `DummyRateLimitStrategy` / `DummyTokenCounter` harness already used by `test_model_client_cache_behavior.py`, with a client that raises once and then succeeds:

- `test_failed_batch_does_not_fail_the_next_batch` — the first batch raises `ExecutionError`, and the next two batches return their own results. This is the regression test for the symptom above.
- `test_thread_exception_is_delivered_only_once` — a recorded exception raises once, then the check is a no-op and the map is empty.

Both fail on current `main` and pass with the change.

```
$ uv run ruff check src/fenic/_inference/model_client.py tests/_inference/test_model_client_thread_exception.py
All checks passed!

$ uv run pytest tests/_inference/test_model_client_thread_exception.py -q
2 passed

1 failed, 194 passed, 4 skipped, 2 errors
```

The failure and errors are pre-existing on unmodified `main` in the same tree (a home-directory path assertion, and two tests needing live API keys). No public API change, so the generated reference is unchanged. No version bump, no dependency changes.

### DIFF
--- a/src/fenic/_inference/model_client.py
+++ b/src/fenic/_inference/model_client.py
@@ -463,6 +463,10 @@ class ModelClient(Generic[RequestT, ResponseT], ABC):
         logger.info(
             f"Creating batch {batch_id} with {len(requests)} requests for {operation_name} using (model: {self.model})"
         )
+        # A previous batch on this thread may have left a recorded exception behind
+        # after surfacing it to its own caller. Clear it so this batch reports only
+        # its own failures.
+        self._clear_thread_exception()
         try:
             return self._make_batch_requests(requests, operation_name, batch_id, request_timeout=request_timeout)
         except Exception as e:
@@ -530,12 +534,22 @@ class ModelClient(Generic[RequestT, ResponseT], ABC):
         token_estimate = self.estimate_tokens_for_request(request)
         return new_future, token_estimate
 
+    def _clear_thread_exception(self):
+        """Discard any exception recorded for the calling thread."""
+        with self.thread_exceptions_lock:
+            self.thread_exceptions.pop(threading.get_ident(), None)
+
     def _maybe_raise_thread_exception(self):
-        """Surface exceptions from event loop to calling thread immediately."""
+        """Surface exceptions from event loop to calling thread immediately.
+
+        The exception is removed as it is raised. It has been delivered to its
+        caller, so a later batch on the same thread must not observe it again.
+        """
         current_thread_id = threading.get_ident()
         with self.thread_exceptions_lock:
-            if current_thread_id in self.thread_exceptions:
-                raise self.thread_exceptions[current_thread_id]
+            exception = self.thread_exceptions.pop(current_thread_id, None)
+        if exception is not None:
+            raise exception
 
     def _calculate_backoff_time(self, backoff_iteration: int) -> float:
         """Calculates the backoff duration using exponential backoff with a maximum limit.
--- a/tests/_inference/test_model_client_thread_exception.py
+++ b/tests/_inference/test_model_client_thread_exception.py
@@ -1,0 +1,159 @@
+"""Batch isolation for exceptions surfaced from the event loop.
+
+`_handle_exception` records a failure against the submitting thread so a long
+batch fails fast instead of waiting on every future. That record must not outlive
+the batch that produced it: a later batch on the same thread would otherwise
+re-raise a stale exception belonging to an earlier, unrelated operation.
+"""
+
+import threading
+from typing import List, Optional, Union
+
+import pytest
+
+from fenic._inference.model_client import (
+    FatalException,
+    ModelClient,
+    TransientException,
+)
+from fenic._inference.rate_limit_strategy import RateLimitStrategy, TokenEstimate
+from fenic._inference.types import (
+    FenicCompletionsRequest,
+    FenicCompletionsResponse,
+    LMRequestMessages,
+)
+from fenic.core._inference.model_catalog import ModelProvider
+from fenic.core._inference.model_provider import ModelProviderClass
+from fenic.core.error import ExecutionError
+from fenic.core.metrics import LMMetrics
+
+
+class DummyProvider(ModelProviderClass):
+    @property
+    def name(self) -> str:
+        return "dummy"
+
+    def create_client(self):
+        return object()
+
+    def create_aio_client(self):
+        return object()
+
+    async def validate_api_key(self) -> None:
+        return
+
+
+class DummyRateLimitStrategy(RateLimitStrategy):
+    def __init__(self):
+        super().__init__(rpm=100)
+
+    def backoff(self, curr_time: float) -> int:
+        return 0
+
+    def check_and_consume_rate_limit(self, token_estimate: TokenEstimate) -> bool:
+        return True
+
+    def context_tokens_per_minute(self) -> int:
+        return 60_000
+
+
+class DummyTokenCounter:
+    def count_tokens(self, messages, ignore_file: bool = False) -> int:
+        return 0
+
+    def count_file_input_tokens(self, messages) -> int:
+        return 0
+
+    def count_file_output_tokens(self, messages) -> int:
+        return 0
+
+
+class FlakyCompletionClient(ModelClient[FenicCompletionsRequest, FenicCompletionsResponse]):
+    """Raises an unhandled error on the first request, then succeeds."""
+
+    def __init__(self, failures: int = 1):
+        super().__init__(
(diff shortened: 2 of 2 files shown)


src/fenic/_inference/model_client.py
```diff
@@ -463,6 +463,10 @@ def make_batch_requests(
         logger.info(
             f"Creating batch {batch_id} with {len(requests)} requests for {operation_name} using (model: {self.model})"
         )
+        # A previous batch on this thread may have left a recorded exception behind
+        # after surfacing it to its own caller. Clear it so this batch reports only
+        # its own failures.
+        self._clear_thread_exception()
```
bcallender, line 469: This clear handles the stale-entry case, but it can't close the race: a previous batch that failed fast returns while its delayed sibling requests are still live, and a sibling that fails *after* this clear runs will repopulate the slot — so this batch still inherits a failure it didn't cause. The window is real because `_make_batch_requests` returns on the first failing future while the rest of the batch's tasks keep running.

src/fenic/_inference/model_client.py
```diff
@@ -530,12 +534,22 @@ def _copy_future_result(input_future: Future, output_future: Future):
         token_estimate = self.estimate_tokens_for_request(request)
         return new_future, token_estimate
 
+    def _clear_thread_exception(self):
+        """Discard any exception recorded for the calling thread."""
+        with self.thread_exceptions_lock:
+            self.thread_exceptions.pop(threading.get_ident(), None)
```
bcallender, line 540: The granularity is the underlying issue: the map is keyed by thread id alone, but `QueueItem` already carries both `thread_id` and `batch_id`. If registration and consumption were keyed by `(thread_id, batch_id)`, a late sibling from an old batch could never be mistaken for the current batch's failure — and this clear-at-start call becomes unnecessary rather than best-effort.

src/fenic/_inference/model_client.py
```diff
@@ -530,12 +534,22 @@ def _copy_future_result(input_future: Future, output_future: Future):
...
+        """Surface exceptions from event loop to calling thread immediately.
+
+        The exception is removed as it is raised. It has been delivered to its
+        caller, so a later batch on the same thread must not observe it again.
+        """
         current_thread_id = threading.get_ident()
         with self.thread_exceptions_lock:
-            if current_thread_id in self.thread_exceptions:
-                raise self.thread_exceptions[current_thread_id]
+            exception = self.thread_exceptions.pop(current_thread_id, None)
```
bcallender, line 550: This half is right — deliver-exactly-once via `pop()` under the existing lock is the correct contract, and it should survive unchanged if the key becomes `(thread_id, batch_id)`.

tests/_inference/test_model_client_thread_exception.py
```diff
@@ -0,0 +1,159 @@
...
+        structured_output=None,
+        temperature=None,
+    )
+
+
+def _batch(client: FlakyCompletionClient, user: str) -> List[Optional[FenicCompletionsResponse]]:
+    return client.make_batch_requests([_request(user)], operation_name="test")
+
+
+def test_failed_batch_does_not_fail_the_next_batch():
```
bcallender, line 131: This test uses one request per batch, so the late-sibling case is structurally unreachable here: there's never an old sibling still alive to fail after the new batch starts. Could you add a deterministic case with a two-request old batch — one fast failure, one delayed failure — and a subsequent batch on the same caller thread, asserting the new batch does *not* observe the old batch's late exception? That's the scenario that distinguishes batch-scoped ownership from clear-at-start.

bcallender (changes requested):
Thanks for chasing this down — the root cause is right, and moving from a read to a consume-on-read under the existing lock is the correct fix for the case it targets: an exception that has already been delivered to its caller must not be handed to a second, unrelated caller later.

There’s a narrower race this doesn’t quite close, though. Picture a batch of two requests running on some thread. The first request fails immediately, the batch reports that failure and returns to its caller — but its sibling, the second request, is still in flight. By the time that sibling actually fails, a new batch has already started on the same thread and has already cleared the slot for a clean start. The delayed sibling’s failure then writes into that now-empty slot, and the new batch — which has nothing to do with the old one — picks it up and fails on an exception that belongs to a request it never made.

The underlying issue is that exception ownership is keyed only by thread id, but a thread can run more than one batch over its life, and a batch’s sibling requests can still be resolving after the next batch has already started. I think this needs to scope both writing and consuming that exception to the batch that owns it — for example, keying by thread id and batch id together — rather than clearing a single thread-wide slot at the start of each batch. It would also be worth accounting for or cancelling a sibling’s in-flight work once its batch has already failed, so an abandoned request doesn’t surface as an unrelated, noisy background error later.

Could you add a regression test for this shape specifically: a batch with at least two requests, one that fails fast and one that fails only after a next batch has already started on the same thread? The current end-to-end test uses a single request per batch, so it can’t exercise a sibling arriving late. This still feels like a bounded fix, not a redesign — batch id should own a batch’s failure, not thread reuse.

bcallender:
Hi @zenprocess, checking in on this one. The root cause you found is right and the consume-on-read half of the fix is exactly what we want to keep.

Two things are outstanding:

1. **The late-sibling race from the review above.** Key the exception registration and consumption by `(thread_id, batch_id)` rather than thread id alone (`QueueItem` already carries both), which makes the clear-at-start call unnecessary, and add a deterministic test with a two-request old batch (one fast failure, one delayed failure) followed by a new batch on the same caller thread, asserting the new batch does not observe the old batch's late exception.
2. **A signed commit.** `main` requires verified signatures and `2857fb6e` is unsigned.

<details>

```bash
```

```bash
git push --force-with-lease
```

</details>

If you would like to finish it, great. If not, that is fine too: maintainer edits are enabled, so if we have not heard back by **Monday 5 October** we will add the remaining change on top of your commit on this branch. You stay the author of your work and get the credit in the release notes.

PR #385, opened 2026-08-19
feat(inference): make structured-output validation failures diagnosable

zenprocess:
## Symptom

When a model ignores `response_format` and returns non-conforming content, fenic fails the whole query with an opaque pydantic error attributed to a DataFrame expression:

```
ExecutionError: Failed to execute query: 1 validation error for EnumModel
  Invalid JSON: expected ident at line 1 column 2 [type=json_invalid, input_value='toolchain']
  ...
This error occurred in the following expression:
  col("t").python_udf()
```

Nothing tells the user which operator failed, what schema it expected, or what the model actually returned. With this change the same failure reads:

```
Structured-output validation failed for semantic.classify: expected EnumModel,
but model 'gpt-4.1-nano' returned content that does not match it: 'toolchain'.
This model or provider does not honour response_format.
Original error: 1 validation error for EnumModel
  Invalid JSON: expected ident at line 1 column 2 [type=json_invalid, input_value='toolchain']
```

## Root cause

`OpenAIChatCompletionsCore.make_single_request` delegates parsing to `client.beta.chat.completions.parse(...)`. That SDK method parses the model's content against the requested schema and raises `pydantic.ValidationError` when it does not conform. The core's exception clauses only catch the SDK's own exception types and fenic's `ValidationError`, so pydantic's escaped unclassified and bubbled up as an `ExecutionError` with no context.

## The change

Catch the pydantic `ValidationError` around the `parse` call and re-raise it as a **fatal** error that names:

- the operator (`semantic.classify`, `semantic.extract`, ...), via the `operation_name` that `get_completions` already receives and now carries on `FenicCompletionsRequest`;
- the expected schema class (`EnumModel`, ...) from `structured_output.pydantic_model`;
- a short preview of the raw content the model returned, extracted from the validation error's `input` and capped at 200 characters.

No repair and no retry. The happy path is untouched: strictly conforming responses never enter this branch.

## Why fatal, not transient

The OpenRouter path catches this same class and returns a `TransientException`, but that is rationalised by provider re-routing — it can retry against a *different* provider that honours the schema. A fixed OpenAI-compatible endpoint has no alternate route, so retrying deterministic schema non-compliance burns quota and lands in the same failure. This path therefore treats it as fatal, consistent with the core's existing comment on deterministic request-construction failures ("retrying cannot help").

Worth considering separately: the two paths now classify the same condition differently, for good reason but without a shared type. Would a shared `StructuredOutputValidationException` (or similar) carrying the operator/schema/preview be useful as a cross-provider primitive? Raising it here as a question, not insisting.

## Tests

`tests/_inference/test_openai_structured_output_diagnostics.py`, following the fake-core pattern in `test_output_token_limits.py` (a `.parse()` that raises, no network):

- a bare enum value produces a fatal error carrying `semantic.classify`, `EnumModel`, and the raw `'toolchain'` preview;
- fenced JSON produces a fatal error carrying `semantic.extract` and the object schema name;
- a missing `operation_name` falls back to `unknown operator` instead of crashing;
- a strictly conforming response still succeeds and returns its content unchanged.

```
$ uv run ruff check <4 changed files>
All checks passed!

$ uv run pytest tests/_inference/test_openai_structured_output_diagnostics.py -q
4 passed

1 failed, 196 passed, 4 skipped, 2 errors
```

The failure and errors are pre-existing on unmodified `main` in the same tree (a home-directory path assertion; two tests needing live API keys). No public API change, so the generated reference is unchanged. No version bump, no dependency changes.

Verified end to end against a local OpenAI-compatible server whose model returns a bare enum value: the error now names the operator, schema, and raw value, as shown at the top.

### DIFF
--- a/src/fenic/_inference/common_openai/openai_chat_completions_core.py
+++ b/src/fenic/_inference/common_openai/openai_chat_completions_core.py
@@ -12,6 +12,7 @@ from openai import (
     RateLimitError,
 )
 from openai.types import CompletionUsage
+from pydantic import ValidationError as PydanticValidationError
 
 from fenic._inference.common_openai.openai_profile_manager import (
     OpenAICompletionProfileConfiguration,
@@ -39,7 +40,7 @@ from fenic.core._inference.model_catalog import (
 from fenic.core._inference.output_token_limits import (
     validate_effective_output_token_limit,
 )
-from fenic.core.error import ValidationError
+from fenic.core.error import ExecutionError, ValidationError
 from fenic.core.metrics import LMMetrics
 
 logger = logging.getLogger(__name__)
@@ -141,7 +142,13 @@ class OpenAIChatCompletionsCore:
             # Choose between parse and create based on structured_output
             if request.structured_output:
                 common_params["response_format"] = request.structured_output.pydantic_model
-                response = await self._client.beta.chat.completions.parse(**common_params)
+                try:
+                    response = await self._client.beta.chat.completions.parse(**common_params)
+                except PydanticValidationError as e:
+                    # The SDK parsed the model's content against response_format and it
+                    # did not conform. Surface what actually came back, rather than an
+                    # opaque pydantic error attributed to a DataFrame expression.
+                    return FatalException(_structured_output_validation_error(self._model, request, e))
             else:
                 response = await self._client.chat.completions.create(**common_params)
 
@@ -272,3 +279,38 @@ class OpenAIChatCompletionsCore:
                 else 0
             ),
         )
+_MAX_RAW_PREVIEW_CHARS = 200
+
+
+def _structured_output_validation_error(model, request, error):
+    """Build a diagnosable error for a response that failed structured-output parsing.
+
+    The OpenAI SDK parses the model's content against response_format and raises a
+    pydantic ValidationError when it does not conform. That error is opaque by itself,
+    so this reconstructs the operator, the expected schema and a short preview of what
+    the model actually returned.
+    """
+    schema_name = (
+        request.structured_output.pydantic_model.__name__
+        if request.structured_output is not None
+        else "unknown"
+    )
+    raw_input = next(
+        (item.get("input") for item in error.errors() if "input" in item), None
+    )
+    preview = _truncate_preview(raw_input, _MAX_RAW_PREVIEW_CHARS)
+    operator = request.operation_name or "unknown operator"
+    return ExecutionError(
+        f"Structured-output validation failed for {operator}: expected {schema_name}, "
+        f"but model '{model}' returned content that does not match it: {preview}. "
+        f"This model or provider does not honour response_format. "
+        f"Original error: {error}"
+    )
+
+
+def _truncate_preview(value, max_chars):
+    """Return a capped, single-line string preview of value."""
+    text = repr(value)
+    if len(text) > max_chars:
+        text = text[: max_chars - 3] + "..."
+    return text.replace("\\n", " ")
--- a/tests/_inference/test_openai_structured_output_diagnostics.py
+++ b/tests/_inference/test_openai_structured_output_diagnostics.py
@@ -1,0 +1,157 @@
+"""Diagnosable failures for structured-output validation in the OpenAI core.
+
+When the OpenAI SDK's .parse() rejects a model's response against the requested
+schema, the resulting pydantic error is opaque by itself. The core should re-raise it
+as a fatal error that names the operator, the expected schema, and a preview of what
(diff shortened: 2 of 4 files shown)


src/fenic/_inference/common_openai/openai_chat_completions_core.py
```diff
@@ -272,3 +279,38 @@ def get_max_output_token_request_limit(
...
+    raw_input = next(
+        (item.get("input") for item in error.errors() if "input" in item), None
+    )
+    preview = _truncate_preview(raw_input, _MAX_RAW_PREVIEW_CHARS)
+    operator = request.operation_name or "unknown operator"
+    return ExecutionError(
+        f"Structured-output validation failed for {operator}: expected {schema_name}, "
+        f"but model '{model}' returned content that does not match it: {preview}. "
+        f"This model or provider does not honour response_format. "
+        f"Original error: {error}"
```
bcallender, line 307: This append defeats the cap established two lines up: the stringified pydantic error embeds the rejected input again, in full. With a long payload the assembled message came out at 676 characters with the raw content appearing twice — once in the bounded preview, once here. Suggest replacing the full pydantic string with a sanitized structural summary (error count plus locations/types/messages, excluding `input` and raw values), or attaching the original as the exception cause so it isn't formatted into routine logs.

src/fenic/_inference/common_openai/openai_chat_completions_core.py
```diff
@@ -272,3 +279,38 @@ def get_max_output_token_request_limit(
...
+    )
+
+
+def _truncate_preview(value, max_chars):
```
bcallender, line 311: The helper itself is right — but it only bounds the preview, not the assembled message. Worth asserting a total bound on the final `ExecutionError` string so the guarantee is on what actually reaches logs, not on one component of it.

tests/_inference/test_openai_structured_output_diagnostics.py
```diff
@@ -0,0 +1,157 @@
...
+        temperature=None,
+        operation_name=operation_name,
+    )
+
+
+def _run(core, request):
+    return asyncio.run(core.make_single_request(request, None))
+
+
+def test_bare_enum_value_produces_diagnosable_fatal():
```
bcallender, line 112: The coverage here is good — especially `test_conforming_response_still_succeeds`, which proves the new except clause doesn't swallow the success path. One addition would close the size/safety claim: a test with a long, multiline, sentinel-bearing payload asserting the *entire* surfaced message is bounded and contains no second copy of the raw input. Right now the 200-char cap is tested on the preview but the appended original error reintroduces the content unbounded.

bcallender (changes requested):
Thanks — this is a real improvement on diagnosability, and I like that you added a test proving a conforming response still succeeds under the new code path. That’s exactly the guard a change like this needs, since it’s easy to widen an error-handling block a little too far without noticing it now catches something it shouldn’t.

One thing needs tightening before this lands: the preview you build is capped at 200 characters, which is the right instinct, but the final message also appends the original error’s own text afterward, and that text still contains the full rejected input. So the cap doesn’t actually bound what a caller — or a log line — ends up seeing. With a representative long input, the complete message runs to 676 characters, with the raw input appearing in it twice.

Could you replace the appended original-error text with a sanitized structural summary instead — something like the error count plus each error’s location and message, without the raw input or any URLs — or otherwise keep the original error available as the exception’s cause, for anyone inspecting it programmatically, without formatting its full text into the message that gets surfaced and logged? And could you add a test with a long, multi-line input that asserts the whole message stays bounded and contains no second copy of the raw input? The diagnostic direction here is right — this is just about making the bound you already added actually hold end to end.

bcallender:
Hi @zenprocess, checking in on this one. The diagnosability improvement is real and we want it in.

Two things are outstanding:

1. **The message bound from the review above.** The 200-character preview is capped, but the assembled message still appends the full pydantic error text, which repeats the rejected input in full. Replace that with a sanitized structural summary (error count plus locations, types and messages, without the input), bound the whole surfaced `ExecutionError` string rather than one component of it, and add a test with a long, multiline, sentinel-bearing payload asserting the entire message is bounded and carries no second copy of the raw input.
2. **A signed commit.** `main` requires verified signatures and `185744f7` is unsigned.

<details>

```bash
```

```bash
git push --force-with-lease
```

</details>

If you would like to finish it, great. If not, that is fine too: maintainer edits are enabled, so if we have not heard back by **Monday 5 October** we will add the remaining change on top of your commit on this branch. You stay the author of your work and get the credit in the release notes.

PR #392, opened 2026-09-21
feat(semantic): add semantic.judge typed judgments with a decision-model provider

bcallender:
## Motivation

Closed decisions should expose their probabilities without requiring callers to
wrap an external SDK in a UDF. `semantic.judge` evaluates several questions over
one string state and returns a typed struct that ordinary expressions can inspect.
Threshold selection stays downstream of inference.

## Design

- Immutable Noul, Choice, and Score questions with validated output field names.
- A native logical expression, local batch lowering, and protobuf serialization.
- TypeSafe System One dispatch through the existing scheduler, cache, rate limits,
  and usage settlement. The SDK does not own a second retry loop.
- Cache identity includes ordered questions and the provider endpoint. Invalid
  answers are not cached; known billed usage is retained.
- Estimated question packing preserves intact state and premise-connected groups.
- An optional `typesafe` dependency, public guide, and runnable example.

No existing text-provider route changes. This PR adds the native operation only;
explicit closed-set lowering for existing operators is a separate dependent change.
Automatic lift and fusion are deferred. No private UDF, prompt heuristic, or
schema-only question recovery is included.

## Limits

The provider tokenizer is not public. Envelope checks are estimates, not a guarantee
of admission. State is never truncated. Null states and failed rows return null
structs; malformed distributions cannot become cache hits.

Response decoding relies on the SDK's required `usage` and `answers` fields.
An SDK response missing required usage fields can still fail the query rather
than produce a null row. Question validation and partition checks repeat at
multiple boundaries; this implementation does not remove that redundant work.

Probabilities do not guarantee calibration or cross-question consistency. The
previous experiment reported macro Brier **0.0508**, macro ECE **0.0598**,
and pooled ECE **0.0567** on **1,959 valid responses out of 2,000 requested
study-held-out rows**. Those rows were held out from this study's question and
threshold selection, not an official test split or proven absent from provider
training. These are prior experiment results, not a new benchmark of this PR.
They do not establish production-classifier replacement quality.

## Validation

- Native fake-transport suite: 27 passed.
- Live-provider integration: **1 passed**, 5 warnings, using `TYPESAFE_API_KEY`.
- Scoped scheduler, cache, catalog, optimizer, equality, serialization, and example
  regressions: **318 passed, 16 deselected**, 9 warnings.
- Public documentation and crawler/LLM assets: **65 URLs validated**.
- Ruff and static `fenic check` passed for both news example scripts.
- GitHub CI passed on `b4928a2` (2026-09-21).

The [parallel news judgments example](https://github.com/typedef-ai/fenic/blob/brandon/semantic-judge/examples/typed_judgments/README.md) reuses
25 synthetic articles. It contrasts extraction plus three classifications with
five independent questions in one judge request per article, then demonstrates
probability thresholds, abstention, source profiles, and gated extraction.

Two live runs are documented, including variable timings and empty extraction
results. The final run measured **100 completion requests versus 25 judge requests**;
the cascade extracted **18/25** articles, avoiding **7 generation calls**. These are
workflow demonstrations, not accuracy or calibration claims. Total fenic-reported
cost for both runs plus the integration test was approximately **$0.010707**.
Existing dependency deprecation warnings remain. No operator or provider source
changed in the CI or example follow-ups.

## Review size

This is a large PR. The operator is one cohesive unit (question types, logical node,
physical lowering, serialization, and the provider path through the shared scheduler),
and splitting it would leave an intermediate state that cannot be exercised end to end.
The closed-set lowering for existing operators is a separate, much smaller follow-up PR.

| Category | Added | Deleted | Net |
| --- | ---: | ---: | ---: |
| Handwritten production (includes the protobuf schema) | 1052 | 0 | 1052 |
| Generated protobuf | 263 | 247 | 16 |
| Generated API reference | 2 | 1 | 1 |
| Tests | 545 | 0 | 545 |
| Documentation and examples | 655 | 35 | 620 |
| Dependencies and developer setup | 146 | 77 | 69 |

## Follow-ups

- Expand the public docstring to name `ExecutionError` at collection time and
  distinguish timeout validation exception classes more precisely.
- Add a direct regression for malformed dictionary entries in the question list
  under strict public argument validation.

https://claude.ai/code/session_01GixBPD2MhK4MFn518kxQu3


src/fenic/_inference/typesafe/typesafe_provider.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+
+    def __init__(self, base_url: str | None = None):
+        self._base_url = base_url
+
+    @property
+    def name(self) -> str:
+        return "typesafe"
+
+    def create_client(self):
```
YoungVor, line 60: The client is constructed without a timeout, so its default (10s) always applies, regardless of the `request_timeout` provided by the user

src/fenic/_inference/typesafe/typesafe_system_one_client.py
```diff
@@ -0,0 +1,190 @@
...
+    ModelClient[FenicCompletionsRequest, FenicCompletionsResponse]
+):
+    """Serve explicit typed questions, never recover questions from text prompts."""
+
+    def __init__(
+        self,
+        rate_limit_strategy: RateLimitStrategy,
+        model: str,
+        queue_size: int = 100,
+        max_backoffs: int = 2,
```
YoungVor, line 44: any reason this is different than the num_backoffs on our other clients?

src/fenic/_inference/typesafe/typesafe_system_one_client.py
```diff
@@ -0,0 +1,190 @@
...
+        ) as error:
+            # SDK error bodies can contain row text. The scheduler owns retry policy.
+            return TransientException(RuntimeError(type(error).__name__))
+        except TypeSafeError as error:
+            # Fatal errors must fail the query without exposing SDK response bodies.
+            return FatalException(RuntimeError(type(error).__name__))
+
+        input_tokens = result.usage.input_tokens
+        output_tokens = result.usage.output_tokens
+        if any(
```
YoungVor, line 122: nullable usage fields discard a billed answer set

Usage.input_tokens / output_tokens are Optional[int] in the SDK schema, but this guard runs before decoding and returns None when either is absent. The request succeeded and was billed — we just drop result.answers and null the row.

Missing usage should degrade metrics, not the result. Suggest treating None as 0 for accounting and only bailing on genuinely invalid values:

src/fenic/api/session/config.py
```diff
@@ -1036,12 +1037,38 @@ class Profile(BaseModel):
...
+
+    Attributes:
+        model_name: Provider model identifier, such as ``jev-1.13.0``.
+        rpm: Requests per minute.
+        tpm: Input tokens per minute.
+        base_url: Optional provider endpoint override.
+    """
+
+    model_config = ConfigDict(extra="forbid")
+    model_name: str = Field(..., description="TypeSafe model identifier.", min_length=1)
```
YoungVor: nit: consider making this a catalog-backed literal instead of str (like OpenAILanguageModel), to be consistent with the other providers and so the unsupported model check becomes a Pydantic error that lists the valid names (if we know the models).  

See LanguageModel union - everything besides OpenRouter restrains the model_name to defined literals

src/fenic/_inference/typesafe/typesafe_system_one_client.py
```diff
@@ -0,0 +1,190 @@
...
+            try:
+                future = asyncio.run_coroutine_threadsafe(
+                    self._client.aclose(), self._event_loop
+                )
+                future.result(timeout=10)
+            except Exception:
+                logger.debug(
+                    "Could not close TypeSafe client before scheduler shutdown"
+                )
+        super().shutdown()
```
YoungVor: nit: shutdown() closes the SDK client before super().shutdown() cancels in-flight tasks, so requests still on the wire fail with transport errors instead of clean cancellation; the correct slot (after cancel, before release_loop) isn’t reachable from a subclass, 

consider a _aclose_provider_client() hook in ModelClient, called after the futures are cancelled but before the loop is released

YoungVor:
My claude pass noticed that Serde guard doesn't include expressions.judge in the module list (test_expression_serde.py)

YoungVor (changes requested):
looks good!  Requesting a few tweaks.  Will be cool to see Typesafe's cost-effective models supported

src/fenic/_inference/typesafe/typesafe_provider.py
```diff
@@ -0,0 +1,59 @@
...
+
+class TypeSafeModelProvider(ModelProviderClass):
+    """TypeSafe implementation of ModelProvider.
+
+    Reads the API key from ``TYPESAFE_API_KEY`` via the SDK's own environment handling;
+    fenic never touches the value.
+    """
+
+    def __init__(self, base_url: str | None = None):
+        self._base_url = base_url
```
cpard: **[P2] Resolve the effective TypeSafe base URL before cache keying**

When `base_url` is omitted, the pinned SDK resolves `TYPESAFE_BASE_URL`, but `_base_url` remains `None`. `_build_request_key()` fingerprints this field, so sessions pointed at different environment-configured endpoints generate the same cache key and can reuse judgments from the wrong endpoint. The registry also treats the endpoint as the default and performs `/models` validation. Please resolve and normalize the effective URL once, then store and pass that value.

src/fenic/core/types/judge.py
```diff
@@ -0,0 +1,303 @@
...
+        else:
+            score = answer.get("score")
+            if (
+                isinstance(score, bool)
+                or not isinstance(score, (int, float))
+                or not math.isfinite(score)
+                or not 0 <= score <= len(keys) - 1
+            ):
+                raise ValueError("Judge score is outside its level range")
+            output[question.name] = float(score)
```
cpard, line 309: **[P2] Enforce consistency between a score and its probabilities**

`score` is checked only for range, not for consistency with the validated distribution. For example, probabilities `{"0": 1.0, "1": 0.0}` with `score=1.0` are accepted and marked cacheable, even though the documented probability-weighted score is `0.0`. This produces a self-contradictory public struct. Please derive the score from the probabilities or reject mismatches with a rounding tolerance.

src/fenic/core/_logical_plan/expressions/judge.py
```diff
@@ -0,0 +1,80 @@
...
+    def children(self) -> list[LogicalExpr]:
+        return [self.state]
+
+    def _validate_completion_parameters(self, session_state: BaseSessionState) -> None:
+        from fenic.core._logical_plan.utils import fetch_model_and_completion_parameters
+
+        _, _, parameters = fetch_model_and_completion_parameters(
+            self.model_alias, session_state.session_config
+        )
+        if not parameters.supports_judge:
```
cpard, line 56: **[P2] Reject profiles on TypeSafe judge aliases**

`ModelAlias(name="judge", profile="does-not-exist")` passes this validation even though the TypeSafe config and catalog declare no profile support. The value enters the request and cache key, but the TypeSafe client never applies it, so the query silently uses default behavior while fragmenting the cache. Please reject a non-null profile during planning.

cpard (changes requested):
I found three additional correctness issues beyond the existing review threads.

src/fenic/_backends/local/model_registry.py
```diff
@@ -377,6 +378,25 @@ def _initialize_language_model(
...
+                    from fenic._inference.typesafe.typesafe_system_one_client import (
+                        TypeSafeSystemOneClient,
+                    )
+                except ImportError as err:
+                    raise ImportError(
+                        "To use TypeSafe models, please install the required dependencies by running: pip install fenic[typesafe]"
+                    ) from err
+                rate_limit_strategy = UnifiedTokenRateLimitStrategy(
+                    rpm=model_config.rpm, tpm=model_config.tpm
+                )
```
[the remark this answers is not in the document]
bcallender: > Use an input-only rate-limit strategy (while retaining output usage in metrics).

Fixed in 77c7e7a5c511a34181a56e69f6971c78cd3c86a3, with stronger controls in b191a3fc38ef9a7728eeda51ac1cdf380debafc7. TypeSafe now reserves and settles input tokens only. Output tokens remain in usage and cost metrics. Tests cover requests whose input fits the limit but whose input-plus-output estimate would exceed it.

docs/topics/typed-judgments.md
```diff
@@ -0,0 +1,111 @@
...
+## Question types
+
+Questions are immutable. Their names must be identifiers, and all generated output
+fields must be unique.
+
+| Factory                | Criteria                                   | Output                                                                             |
+| ---------------------- | ------------------------------------------ | ---------------------------------------------------------------------------------- |
+| `JudgeQuestion.noul`   | Optional `true` and `false` descriptions   | `<name>_p`, the affirmative probability                                            |
+| `JudgeQuestion.choice` | Between 2 and 255 described string options | `<name>`, `<name>_confidence`, and `<name>_p_<option>`                             |
+| `JudgeQuestion.score`  | At least two ordered text levels           | Numeric `<name>` on the zero-based level scale, confidence, and `<name>_p_<index>` |
```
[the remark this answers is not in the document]
bcallender: > This documents Score as having only a lower bound, but the provider accepts 2–10 levels.

Fixed in 77c7e7a5c511a34181a56e69f6971c78cd3c86a3. The guide now states 2–10 levels, matching local validation. Boundary tests accept 2 and 10 and reject 1 and 11.

src/fenic/_inference/typesafe/typesafe_provider.py
bcallender: > The client is constructed without a timeout, so its default (10s) always applies

Fixed in aff3bce07158fde29fd7270d60a9ffb5953e6ad7, with constructor tests and docs in 5daf836d13b60599f370b982c6971d0f7f6d242c. Inference SDK clients explicitly use 600 seconds, so the SDK's 10-second default no longer truncates an allowed Fenic request timeout. Fenic enforces the requested deadline, capped at 600 seconds. The separate key-validation client keeps a 10-second timeout and is closed afterward. SDK retries remain disabled.

src/fenic/_inference/typesafe/typesafe_system_one_client.py
bcallender: > any reason this is different than the num_backoffs on our other clients?

There isn't an SDK requirement behind the difference. This preserves the new adapter's 2-retry default and the existing clients' 10-retry defaults rather than retuning other providers in this PR. fd2dc3e751383fcc07d950031126d507faf62838 makes `max_backoffs` configurable and applies the limit per request. SDK retries are disabled, so there is no second retry loop underneath it.

src/fenic/_inference/typesafe/typesafe_system_one_client.py
bcallender: > Missing usage should degrade metrics, not the result.

Fixed in aff3bce07158fde29fd7270d60a9ffb5953e6ad7, with accounting controls in 5daf836d13b60599f370b982c6971d0f7f6d242c. A valid answer with incomplete usage is returned, counted and cacheable, with `usage=None`. Rather than turn unknown values into zero, the entire incomplete request is excluded from token/cost totals, settlement and estimator training. The guide calls out that those totals are incomplete. Genuinely invalid usage values still fail validation.

src/fenic/api/session/config.py
bcallender: > consider making this a catalog-backed literal instead of str

Fixed in aff3bce07158fde29fd7270d60a9ffb5953e6ad7. `TypeSafeLanguageModel.model_name` now uses the catalog-backed literals `jev-1.13.0`, `jev-latest` and `jev-preview`. Unsupported names produce a Pydantic validation error listing the accepted values.

src/fenic/_inference/typesafe/typesafe_system_one_client.py
bcallender: > the correct slot (after cancel, before release_loop) isn’t reachable from a subclass

Fixed in aff3bce07158fde29fd7270d60a9ffb5953e6ad7 through an async provider-close hook after task cancellation and before loop release. abc1f33a7bff3028c38d195c4990f0404a09922b bounds that close wait to 10 seconds and cancels a timed-out close. The CI-racy assertion is now event-synchronized in 5c65868ebba64ea9538b661dcc52b6520771e12f. Tests retain one or two peer clients and verify actual cancellation, loop-reference release and continued peer-loop use; negative controls catch suppressed cancellation or release.

src/fenic/_inference/typesafe/typesafe_provider.py
bcallender: > Please resolve and normalize the effective URL once, then store and pass that value.

Fixed in f316b75c9c855d5c02461b8e17992f926b473b33, with constructor/cache controls in c03f792bdeed4e8688ac95a34adc5c6ec13a7a69. The provider resolves the explicit URL, environment or SDK default once and normalizes its trailing slash. That same value reaches SDK construction and cache identity. Registry validation now distinguishes the effective default endpoint from a custom endpoint instead of relying on whether the stored URL is truthy.

src/fenic/core/types/judge.py
bcallender: > Please derive the score from the probabilities or reject mismatches with a rounding tolerance.

Fixed in f316b75c9c855d5c02461b8e17992f926b473b33 and 16087b9c7959794bdd6529b9c264e638616d76be. We reject inconsistent scores rather than rewrite them. For `k` levels, the absolute tolerance is `0.005 * (1 + k * (k - 1) / 2) + 1e-6`, with zero relative tolerance.

This is our empirical compatibility rule derived from two-decimal rounding, not a TypeSafe precision guarantee. Offline replay accepted all 2,357 archived four-level scores under this rule; a strict `1e-6` tolerance rejected 1,383. That evidence does not establish a service-wide guarantee. Tests cover archived examples, inside/outside boundaries for 2 and 10 levels, and rejection of the all-mass-at-level-0 / score-1 contradiction.

src/fenic/core/_logical_plan/expressions/judge.py
bcallender: > Please reject a non-null profile during planning.

Fixed in f316b75c9c855d5c02461b8e17992f926b473b33. Planning rejects a non-null profile when the selected judge model does not support profiles, including TypeSafe aliases. It no longer silently applies defaults while creating a separate cache identity.

bcallender:
> My claude pass noticed that Serde guard doesn't include expressions.judge in the module list (test_expression_serde.py)

Fixed in aff3bce07158fde29fd7270d60a9ffb5953e6ad7. The serde inventory includes `expressions.judge`, with a judge round-trip example. A negative control verifies that removing its example makes the completeness guard fail.

Original comment: https://github.com/typedef-ai/fenic/pull/392#issuecomment-5769788234

src/fenic/_inference/openai/openai_provider.py
```diff
@@ -33,6 +33,7 @@ def create_aio_client(self):
         return AsyncOpenAI(
             base_url=self._base_url,
             http_client=httpx.AsyncClient(timeout=MAX_MODEL_CLIENT_TIMEOUT),
+            max_retries=0,
```
YoungVor, line 36: Worth documenting the behavior change in PR description: this turns off SDK retries for every OpenAI client, not just TypeSafe. The scheduler only retries timeouts, connection errors and 429s, so a 5xx now hits except OpenAIError → FatalException (openai_chat_completions_core.py:234) and fails the query. The SDK used to retry these twice.

src/fenic/_inference/typesafe/typesafe_system_one_client.py
YoungVor: ah, makes sense

src/fenic/_inference/language_model.py
```diff
@@ -81,6 +83,60 @@ def get_completions(
...
+        requests = []
+        owners = []
+        failed = set()
+        for index, state in enumerate(states):
+            if state is None:
+                failed.add(index)
+                continue
+            try:
+                groups = self.client.judge_partitions(state, questions)
+            except ValueError:
```
YoungVor, line 106: an oversized state or question group comes back null with no log, no request and no cost, so an all-null column gives no hint why. A logger.warning with the row count (no state text) would be enough.

src/fenic/_inference/typesafe/typesafe_system_one_client.py
```diff
@@ -0,0 +1,193 @@
...
+        except (
+            TypeSafeRateLimitError,
+            TypeSafeInternalServerError,
+            TypeSafeAPIConnectionError,
+        ) as error:
+            # SDK error bodies can contain row text. The scheduler owns retry policy.
+            return TransientException(RuntimeError(type(error).__name__))
+        except TypeSafeError as error:
+            # Fatal errors must fail the query without exposing SDK response bodies.
+            return FatalException(RuntimeError(type(error).__name__))
```
YoungVor, line 113: The guide says failed responses produce null structs and only auth or config errors fail the query, but every non-transient TypeSafeError ends up here. The token estimates “do not guarantee provider admission”, so a row just over the provider’s real limit gets a 400 and fails the whole DataFrame. Could request-level errors return None and keep FatalException for auth and permission errors?

src/fenic/_inference/model_client.py
```diff
@@ -817,7 +862,7 @@ async def _process_single_request(self, queue_item: QueueItem[RequestT]):
                 logger.warning(
                     f"Request for model {self.model} in batch {queue_item.batch_id} timed out after {timeout} seconds. Retrying."
                 )
-                await self.retry_queue.put(queue_item)
+                await self._retry_or_fail(queue_item, asyncio.TimeoutError())
```
YoungVor, line 865: Nice find - I think requests that would always timeout would just hang forever before this fix.

Worth documenting in the PR description that timeouts count toward the client's max_backoffs.

YoungVor (approved):
A few more minor comments - mainly to make sure we include notes in the pr description since this changes behavior in all the client sdks not just the new typesafe one

src/fenic/_inference/openai/openai_provider.py
bcallender: > The scheduler only retries timeouts, connection errors and 429s, so a 5xx now hits except OpenAIError → FatalException

Fixed, rather than only documented, in 30ddb180d4c596614cce080bbee890f3e452a015, 7e55dafb844f48e43f28dddf4a755d6e3f72c9a1 and 83480771acb16fea84ff1e99244bc475f364cdfe. Both chat and embeddings now classify 408, 409, non-quota 429 and 5xx responses as scheduler-transient, subject to the SDK retry header. Auth, permission, request/config errors and exhausted quota remain fatal. SDK retries stay disabled.

Real OpenAI SDK requests through a fake HTTP transport exercise the actual scheduler: 503 then success, repeated 503 exhaustion, and a zero-backoff one-send control for both clients. Additional controls cover ordinary 429, retry headers and fatal quota. The PR description now states the retry ownership change.

src/fenic/_inference/language_model.py
bcallender: > A logger.warning with the row count (no state text) would be enough.

Added in 30ddb180d4c596614cce080bbee890f3e452a015. `get_judgments` emits one warning per batch with the number of input rows rejected by size/packing. It excludes pre-existing null inputs and logs no state, question or example text. Mixed valid/oversized/null inputs and oversized question groups are covered. The dependent PR also exercises this same gate through filter and join.

src/fenic/_inference/typesafe/typesafe_system_one_client.py
bcallender: > Could request-level errors return None and keep FatalException for auth and permission errors?

Fixed in 30ddb180d4c596614cce080bbee890f3e452a015, with additional coverage in 83480771acb16fea84ff1e99244bc475f364cdfe. TypeSafe 400, 422 and SDK response-validation failures return `None` for the affected request. Valid neighboring rows still succeed; rejected results are neither cached nor assigned fabricated usage. Auth, permission and generic/configuration errors remain fatal, with sanitized exception-type-only messages. Connection, rate-limit and server errors remain retryable.

src/fenic/_inference/model_client.py
bcallender: > Worth documenting in the PR description that timeouts count toward the client's max_backoffs.

The updated description now states that timeouts and transient provider failures consume the same request-local `max_backoffs` budget. The maximum is one initial send plus `max_backoffs` retries. The existing timeout-budget implementation is preserved in rebased commit ab48285, with its retry and shutdown regressions passing at 83480771acb16fea84ff1e99244bc475f364cdfe.

src/fenic/_inference/typesafe/typesafe_provider.py
bcallender: > Please resolve and normalize the effective URL once, then store and pass that value.

The fix is preserved in the reviewed, rebased head 83480771acb16fea84ff1e99244bc475f364cdfe. Its fixing commit is now ec19adc, with endpoint-construction/cache controls in f22c036. The effective endpoint still drives SDK construction, cache identity and default-versus-custom validation. The final parent controls passed 345 tests. I am re-requesting your review of this updated head.

src/fenic/core/types/judge.py
bcallender: > Please derive the score from the probabilities or reject mismatches with a rounding tolerance.

The rejection check and empirical tolerance are preserved in rebased commits ec19adc and 0733cea, now included in reviewed head 83480771acb16fea84ff1e99244bc475f364cdfe. Scores are not rewritten. For `k` levels, the absolute tolerance remains `0.005 * (1 + k * (k - 1) / 2) + 1e-6`, with zero relative tolerance.

This is our empirical compatibility rule derived from observed two-decimal answers, not a TypeSafe precision guarantee. The current PR description makes that distinction explicit. I am re-requesting your review.

src/fenic/core/_logical_plan/expressions/judge.py
bcallender: > Please reject a non-null profile during planning.

The native `semantic.judge` alias check remains in rebased commit ec19adc and reviewed head 83480771acb16fea84ff1e99244bc475f364cdfe. Non-null TypeSafe profiles are rejected during native-judge planning. The separate request to move the closed-set compatibility operators' runtime constraints to planning is tracked in TD-5624; it does not remove this existing native check. I am re-requesting your review of the rebased head.

PR #393, opened 2026-09-21
feat(semantic): route closed-set operators through the decision provider

bcallender:
## Motivation

Applications can use a decision model for familiar predicate, classification,
join, and sentiment operations without rewriting them as explicit judgments.
This change builds on the native `semantic.judge` implementation.

## Design

- Predicate and join inputs become explicit Noul questions. Filtering uses the
  same predicate route. A probability greater than 0.5 means true; a tie is false.
- Classification and sentiment become explicit Choice questions. Stable option
  identifiers preserve original labels, descriptions, and their order.
- Complete rendered inputs and labeled examples remain intact. No prompt
  heuristic or response-schema recovery is involved.
- Only TypeSafe selects the typed request sender. Other providers retain their
  completion requests, parameters, examples, and postprocessing.
- Native scheduling, retry ownership, cache validation, and usage accounting
  remain shared. Decision-model joins forward their request timeout.
- Unsupported open-ended operations fail before a provider request.

This PR is stacked on the native `semantic.judge` PR and should merge after it.

## Limits

The integration is local-backend-only. Classification accepts 2 through 255
labels. Custom temperatures and model profiles are unsupported. Null handling
follows each existing operator; classification and sentiment also skip empty
strings. There is no operator fusion or reduction in join candidate pairs.

The predicate cutoff is not workload-calibrated. Use `semantic.judge` for explicit
probabilities and downstream threshold selection. The earlier experiment reported
macro Brier 0.0508, macro ECE 0.0598, and pooled ECE 0.0567 on 1,959 valid
responses from 2,000 study-held-out rows. Those rows were held out from that
study's question and threshold selection, not an official test split or proven
absent from provider training. These prior results are not a new measurement of
the compatibility routes and do not establish production-replacement quality.

## Validation

- Closed-set fake-transport tests: 20 passed.
- Scoped operator, scheduler, cache, optimizer, and serialization regressions:
  357 passed, 21 deselected.
- Keyless live-provider integration: one test skipped.
- Scoped Ruff checks passed; the new files pass the formatter check.
- `just docs-check` validated 65 sitemap URLs and their documentation assets.
- `fenic check` found no issues in the new runnable example.

Tests exercise real SDK response models with a fake transport and refuse network
connections. Other-provider tests verify the existing completion path, including
its unchanged join timeout behavior. No live inference ran.

## Review size

Counts cover this dependent change only. No generated code or dependency metadata changes.

| Category | Added | Deleted | Net |
| --- | ---: | ---: | ---: |
| Handwritten production | 155 | 1 | 154 |
| Tests | 429 | 0 | 429 |
| Documentation and examples | 115 | 0 | 115 |

https://claude.ai/code/session_01GixBPD2MhK4MFn518kxQu3


src/fenic/_backends/local/semantic_operators/decision.py
```diff
@@ -0,0 +1,107 @@
...
+            )
+
+    def send_requests(
+        self, messages_batch: list[Optional[LMRequestMessages]]
+    ) -> list[Optional[str]]:
+        """Preserve input and examples, then convert a typed answer for the operator."""
+        states = [
+            None
+            if message is None
+            else json.dumps(
```
cpard: **P2 — Keep decision state structured on the wire**

json.dumps(...) turns the intended {input, examples} object into a string before get_judgments. TypeSafeSystemOneClient passes request.messages.user directly to system_one, and the SDK serializes that value without parsing it, so the request body contains a string-valued state rather than a JSON object. This contradicts the appended instruction that the state is a JSON object and forces the provider to infer and parse a textual blob, losing the structured state boundary for examples. The fake transport currently calls json.loads(state), which masks this wire-format difference.

Please thread a JSON object through the judgment request path (canonicalizing it as needed for token counting and cache fingerprints), then add a transport-level assertion that the prepared SDK request contains an object-valued state. TypeSafe's state documentation recommends passing dictionaries or lists directly for related context and examples: https://docs.typesafe.ai/concepts/state

src/fenic/_backends/local/transpiler/plan_converter.py
```diff
@@ -274,7 +274,8 @@ def _convert_to_physical_plan(self, logical: LogicalPlan, cache_keys: set[str])
                 session_state=self.session_state,
                 examples=logical.examples(),
                 temperature=logical.temperature,
-                model_alias=logical.model_alias
+                model_alias=logical.model_alias,
+                request_timeout=logical.request_timeout,
```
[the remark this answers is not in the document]
bcallender: > Add the optional proto field and serde handling, plus a round-trip assertion for a non-default timeout.

Fixed in 1eaca0e52db34c6fb95a28007fabde61693ebf3e. `SemanticJoinProto` now has optional timeout field 10, with generated bindings and serialization/deserialization support. Absent values remain `None`. The regression round-trips a 9-second timeout through the logical plan and verifies that it reaches the decision request.

src/fenic/_backends/local/semantic_operators/decision.py
bcallender: > Please thread a JSON object through the judgment request path

Fixed in dfd64b78a0d76a4e6530acd2e856d68b1187098a. The decision sender now passes `{input, examples}` as an object through a dedicated judgment-state field to the SDK. Canonical text is used for counting, partitioning and cache fingerprints, not as the wire value. Ordinary completion messages and native string-valued judge states stay strings.

The fake transport now accepts the dict directly. A real `AsyncTypeSafeClient` with `MockTransport` asserts the prepared HTTP body's object-valued state, including nested values and Unicode. It also verifies that literal JSON-looking text stays a string with a distinct cache identity. Restoring the old string conversion makes that transport test fail.

src/fenic/_inference/language_model.py
```diff
@@ -54,6 +57,13 @@ def get_completions(
         operation_name: Optional[str] = None,
         request_timeout: Optional[float] = None,
     ) -> list[Optional[FenicCompletionsResponse]]:
+        if self.provider == ModelProvider.TYPESAFE:
```
YoungVor, line 60: is it also possible to also check whether TypeSafe is used with map, extract, summarize, and reduce in the planning phase?

src/fenic/_backends/local/semantic_operators/decision.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def __init__(
+        self,
+        sender: CompletionOnlyRequestSender,
+        instructions: str,
+        classes: Optional[Sequence[ResolvedClassDefinition]] = None,
+    ):
+        self.model = sender.model
+        self.config = sender.inference_config
+        if self.config.temperature not in (None, 0):
```
YoungVor, line 27: Also here, can we check the temperature and profile config restraints with TypeSafe in the planning phase?

src/fenic/_backends/local/semantic_operators/decision.py
```diff
@@ -0,0 +1,104 @@
...
+    def send_requests(
+        self, messages_batch: list[Optional[LMRequestMessages]]
+    ) -> list[Optional[str]]:
+        """Preserve input and examples, then convert a typed answer for the operator."""
+        states = [
+            None
+            if message is None
+            else {
+                "input": message.user,
+                "examples": [
```
YoungVor, line 79: Every row’s state includes all the examples, and sentiment adds its 9 built-in examples to every row. So ordinary-length rows can go over the estimated 32k limit, and when they do they come back null with no log (the same silent path as my language_model.py:106 comment on #392). Here the effect is worse than a null value: filter and join treat null as false, so rows and matches disappear without any signal. The null behavior itself is documented. A warning with the count of rows nulled for size, added at the #392 spot, would cover both PRs.

YoungVor (approved):
Looks good!  I have some suggestions, but I don't think they are blocking.

src/fenic/_inference/language_model.py
bcallender: > is it also possible to also check whether TypeSafe is used with map, extract, summarize, and reduce in the planning phase?

Tracked in [TD-5624](https://linear.app/typedef-ai/issue/TD-5624). Per the agreed scope, moving these checks to planning is a follow-up rather than another change in this PR. The existing execution-time rejection remains in place.

src/fenic/_backends/local/semantic_operators/decision.py
bcallender: > can we check the temperature and profile config restraints with TypeSafe in the planning phase?

Tracked with the operator checks in [TD-5624](https://linear.app/typedef-ai/issue/TD-5624). This PR retains the current runtime restrictions. Moving the closed-set operators' temperature/profile checks to planning is explicitly deferred to that follow-up.

src/fenic/_backends/local/semantic_operators/decision.py
bcallender: > A warning with the count of rows nulled for size, added at the #392 spot, would cover both PRs.

The shared warning is in parent commit 30ddb180d4c596614cce080bbee890f3e452a015. Child commit 032598f13f156e3b6c74086e68139f63541e7f49 verifies both real operator paths through the packing gate: one rejected filter input and three rejected join candidate pairs in the fixtures. Null inputs do not inflate the count, valid rows/matches survive, and the warning contains no input text. No separate filter/join warning path was needed.

src/fenic/_backends/local/semantic_operators/decision.py
bcallender: > Please thread a JSON object through the judgment request path

The structured-state fix is preserved in rebased commit 7838c57, included in reviewed head 032598f13f156e3b6c74086e68139f63541e7f49. The SDK still receives the actual `{input, examples}` object; canonical text is only for counting and fingerprints. The new shared size warning also works on this path. The real SDK prepared-body regression and full core suite passed at this head, and I am re-requesting your review.

